PR #1, opened 2021-11-08
feat: subtree get/insert

fominok:
This is an attempt to implement simple insert and get operations for GroveDb (inside subtrees without exposing to public API yet) to store binary data

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /target
+.idea
 Cargo.lock
 subtrees
+test
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 /target
 Cargo.lock
-*.db
+subtrees
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 /target
 .idea
 Cargo.lock
-grove.db
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
 /target
 .idea
 Cargo.lock
-subtrees
-test
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,18 @@
+unstable_features = true
+
+blank_lines_lower_bound = 0
+condense_wildcard_suffixes = true
+error_on_line_overflow = true
+error_on_unformatted = true
+format_code_in_doc_comments = true
+format_macro_matchers = true
+format_strings = true
+license_template_path = ""
+imports_granularity = "Crate"
+normalize_comments = true
+normalize_doc_attributes = true
+reorder_impl_items = true
+group_imports = "StdExternalCrate"
+use_field_init_shorthand = true
+use_try_shorthand = true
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -23,24 +30,30 @@ pub struct GroveDb {
 impl GroveDb {
     pub fn new() -> Result<Self, Error> {
         let subtrees = vec![
-            Merk::open("./test1.db")?,
-            Merk::open("./test2.db")?,
-            Merk::open("./test3.db")?,
-            Merk::open("./test4.db")?,
-            Merk::open("./test5.db")?,
-            Merk::open("./test6.db")?,
+            Merk::open(Path::new(SUBTREES_DIR).join("test1.db"))?,
```
shumkov, line 33: All trees need to be stored in one database

fominok: Yeah, it will be a scope of a next task

src/subtree.rs
```diff
@@ -0,0 +1,144 @@
...
+                .ok_or(Error::InvalidPath("key not found in Merk"))?
+                .as_slice(),
+        )?;
+        element.follow_reference(&merk)
+    }
+
+    pub fn insert(&self, merk: &mut Merk, path: &[&[u8]], key: &[u8]) -> Result<(), Error> {
+        // check if a tree was inserted by the path
+        if let Some((tree_key, tree_path)) = path.split_last() {
+            Element::get(merk, tree_path, tree_key)?.is_tree()?;
```
QuantumExplorer, line 81: Why use is_tree? Shouldn't this fail no matter what if an element exists already?

QuantumExplorer:
Very good initial work guys!

src/subtree.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    /// Recursively follow `Element::Reference`
+    fn follow_reference(self, merk: &Merk) -> Result<Element, Error> {
+        if let Element::Reference(reference_merk_key) = self {
+            let element = Element::decode(
+                merk.get(reference_merk_key.as_slice())?
+                    .ok_or(Error::InvalidPath("key not found in Merk"))?
+                    .as_slice(),
+            )?;
+            element.follow_reference(merk)
```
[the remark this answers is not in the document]
fominok: Good catch, thank you

src/subtree.rs
fominok: It checks whether there is a tree at path, it uses `split_last` to get path part and last path element to reuse our `get` method which expects a key, so in this case the key will be last element of `path`

Tldr: it checks the presence of a tree to insert key into

QuantumExplorer: Sorry I misunderstood the code.

src/lib.rs
fominok: Actually, done in this PR, please take a look

QuantumExplorer (approved):

PR #7, opened 2021-12-06
feat: abstract Merk storage

fominok:
This PR includes abstracting from RocksDB as well as making Merk unaware of prefixes and other details.

GroveDB chooses the storage, but that could be also moved out to the caller site later if needed.


QuantumExplorer (approved):
We can create other PRs for the work needed in comments.

antouhou (approved):

PR #9, opened 2021-12-08
feat: Simple checkpoint

fominok:
This PR adds an ability to make a checkpoint of current GroveDB state, basically by reusing RocksDB checkpointing funcitonality;
Also there is an attempt to deprecate `failure` crate and to do a little cleanup of error handling story

### DIFF
--- a/grovedb/src/tests.rs
+++ b/grovedb/src/tests.rs
@@ -239,3 +239,62 @@ fn test_root_tree_leafs_are_noted() {
     assert_eq!(db.root_leaf_keys, hm);
     assert_eq!(db.root_tree.leaves_len(), 2);
 }
+
+#[test]
+fn test_checkpoint() {
+    let mut db = make_grovedb();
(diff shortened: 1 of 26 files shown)


grovedb/src/tests.rs
```diff
@@ -239,3 +239,62 @@ fn test_root_tree_leafs_are_noted() {
...
+            .get(&[b"key1"], b"key4")
+            .expect("cannot get from checkpoint"),
+        element2,
+    );
+
+    assert_eq!(
+        db.get(&[b"key1"], b"key4")
+            .expect("cannot get from GroveDB"),
+        element3
+    );
```
QuantumExplorer, line 299: Let's just add an assert that verifying that a key/value pair inserted into the checkpoint doesn't appear in the db and vice versa.

QuantumExplorer (changes requested):

QuantumExplorer (approved):

PR #10, opened 2021-12-09
feat: proof verification

iammadab:


grovedb/src/lib.rs
```diff
@@ -10,10 +10,16 @@ use std::{
...
-use rs_merkle::{algorithms::Sha256, MerkleTree};
+use merk::{
+    self, execute_proof,
+    proofs::{query::Map, Query},
+    rocksdb, Merk,
+};
+use rs_merkle::{algorithms::Sha256, MerkleProof, MerkleTree};
 use subtree::Element;
 
+use crate::Error::InvalidProof;
```
fominok, line 21: This import is not needed

grovedb/src/lib.rs
```diff
@@ -36,6 +42,8 @@ pub enum Error {
     CyclicReference,
     #[error("reference hops limit exceeded")]
     ReferenceLimit,
+    #[error("incalid proof")]
```
fominok, line 45: A typo

grovedb/src/lib.rs
```diff
@@ -274,6 +283,49 @@ impl GroveDb {
         Ok(proof_result)
     }
 
+    pub fn verify_proof(
+        path: &[&[u8]],
+        proofs: Vec<Vec<u8>>,
```
fominok, line 288: Looks like we take ownership of proof Vec, but do we really need it?

grovedb/src/lib.rs
```diff
@@ -274,6 +283,49 @@ impl GroveDb {
...
+        }
+
+        if proofs.len() - 1 != path.len() {
+            return Err(Error::InvalidProof("Proof length should be one greater than path"));
+        }
+
+        let (mut last_root_hash, mut last_result_map) = execute_proof(&proofs[0][..]).unwrap();
+
+        for i in 1..proofs.len() {
+            if i == proofs.len() - 1 {
```
fominok, line 303: Instead of iteration by index we can iterate by elements, also to split slice into all elements except last and a last one there is a method `split_last`

grovedb/src/lib.rs
```diff
@@ -274,6 +283,49 @@ impl GroveDb {
...
+                if root_proof.verify(expected_root_hash, &vec![0], &[a], 2) {
+                    break;
+                } else {
+                    return Err(Error::InvalidProof("Root hashes didn't match"));
+                }
+            } else {
+                // Merk proof, validate that the proof is valid and
+                // the result map contains the last hash i.e the previous
+                // merk was a child of this merk
+                let proof_result = execute_proof(&proofs[i][..]).unwrap();
```
fominok, line 316: If these `unwrap`s cannot panic please use `expect(..)` with a description why the won't do so, but I suppose these may panic

grovedb/src/lib.rs
```diff
@@ -36,6 +42,8 @@ pub enum Error {
     CyclicReference,
     #[error("reference hops limit exceeded")]
     ReferenceLimit,
+    #[error("invalid proof")]
```
fominok, line 45:
```suggestion
    #[error("invalid proof {0}")]
```

grovedb/src/lib.rs
```diff
@@ -247,14 +255,16 @@ impl GroveDb {
                     .root_leaf_keys
                     .get(*key)
                     .ok_or(Error::InvalidPath("root key not found"))?;
+                println!("Key index {}", root_key_index);
```
fominok, line 258:
```suggestion
```

grovedb/src/lib.rs
```diff
@@ -278,6 +288,86 @@ impl GroveDb {
...
+            .expect("Constraint checks above enforces leaf proof must exist");
+
+        let (mut last_root_hash, leaf_result_map) = match execute_proof(&leaf_proof[..]) {
+            Ok(result) => Ok(result),
+            Err(e) => Err(Error::InvalidProof("Invalid proof element")),
+        }?;
+
+        let mut proof_path_zip = proof_iterator.zip(reverse_path_iterator).peekable();
+
+        while let Some((proof, key)) = proof_path_zip.next() {
```
fominok, line 323: `while let Some(x) = iterator.next()` looks like `for x in iterator` for me, what do you think?

grovedb/src/lib.rs
```diff
@@ -278,6 +288,86 @@ impl GroveDb {
...
+            } else {
+                // Last proof (root proof)
+                let root_proof = match MerkleProof::<Sha256>::try_from(&proof[..]) {
+                    Ok(root_proof) => Ok(root_proof),
+                    Err(e) => Err(Error::InvalidProof("Invalid proof element")),
+                }?;
+                let a: [u8; 32] = last_root_hash;
+                if root_proof.verify(
+                    expected_root_hash,
+                    &vec![root_leaf_keys[&key.to_vec()]],
```
fominok, line 357:
```suggestion
                    &[root_leaf_keys[*key]],
```

grovedb/src/lib.rs
fominok: oh it's because of `peek` I see, nvm then

iammadab: Yeah

fominok (approved):

grovedb/src/lib.rs
```diff
@@ -306,6 +310,87 @@ impl GroveDb {
         Ok(proof_result)
     }
 
+    pub fn verify_proof(
+        path: &[&[u8]],
+        proofs: &mut Vec<Vec<u8>>, // Generic into_iterator (trait) u8
+        expected_root_hash: [u8; 32],
```
QuantumExplorer, line 316: Instead of this function, we should instead have an execute_proof function that returns the proof root hash. Then we can also have a verify proof function that uses the execute_proof function and compares the root hashes.

QuantumExplorer (changes requested):

QuantumExplorer (approved):

PR #11, opened 2021-12-20
feat: insert if not exists

iammadab:
Insert an element at a path only if that path is empty


grovedb/src/tests.rs
```diff
@@ -315,3 +315,29 @@ fn test_checkpoint() {
...
+
+    // Should propagate errors from insertion
+    let result = db.insert_if_not_exists(&[TEST_LEAF, b"unknown"], b"key1".to_vec(), Element::empty_tree());
+    match result {
+        Ok(_) => {
+            panic!("Should result in an error")
+        },
+        Err(_e) => {},
+    }
+}
```
fominok, line 343:
```suggestion
    assert!(matches!(result, Err(Error::InvalidPath(_))));
```

fominok (approved):

QuantumExplorer (approved):

PR #20, opened 2021-12-30
add `get_query` method

fominok:
This PR adds `get_query` method which adds an ability to get elements from GroveDB using same `PathQuery` as used for proof generation.
Current implementation uses raw iterator.


shumkov (approved):
utACK

QuantumExplorer (approved):

PR #21, opened 2021-12-30
feat: aux data

fominok:
This PR adds `put_aux`, `get_aux` and `delete_aux` operations to GroveDB;
Also reorganizes the code a little bit splitting into modules.

edit: also proxies `flush` method to rocksdb from GroveDB API


grovedb/src/operations/aux.rs
```diff
@@ -0,0 +1,17 @@
...
+    pub fn put_aux(&mut self, key: &[u8], value: &[u8]) -> Result<(), Error> {
+        Ok(self.meta_storage.put_aux(key, value)?)
+    }
+
+    pub fn delete_aux(&mut self, key: &[u8]) -> Result<(), Error> {
+        Ok(self.meta_storage.delete_aux(key)?)
+    }
+
+    pub fn get_aux(&mut self, key: &[u8]) -> Result<Option<Vec<u8>>, Error> {
+        Ok(self.meta_storage.get_aux(key)?)
```
shumkov, line 15: Does it stores with prefixes in the same CF like merk data

fominok: No, this storage has empty prefix and it all be stored as is in aux column family

QuantumExplorer (approved):

grovedb/src/operations/aux.rs
shumkov: Sounds good! Thank you!

PR #25, opened 2022-01-13
feat: path references

QuantumExplorer:


.gitignore
```diff
@@ -3,3 +3,5 @@
 Cargo.lock
 index.node
 node_modules
+.DS_Store
```
fominok, line 6: what is `.DS_Store` ?

fominok (approved):

.gitignore
QuantumExplorer: It's a file that is created by the mac indexer.

PR #27, opened 2022-01-17
feat: range after items

iammadab:
Contains implementation for 
- range_after
- range_after_to
- range_after_to_inclusive

### DIFF
--- a/merk/src/proofs/query/mod.rs
+++ b/merk/src/proofs/query/mod.rs
@@ -317,38 +317,6 @@ impl QueryItem {
                 end: end.to_vec(),
             });
         }
-
-        // if lower_unbounded {
-        //     if upper_unbounded {
-        //         QueryItem::RangeFull(RangeFull)
-        //     } else {
-        //         if end_inclusive {
-        //             QueryItem::RangeToInclusive(RangeToInclusive {
-        //                 end: end.0.to_vec(),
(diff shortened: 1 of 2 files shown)


merk/src/proofs/query/mod.rs
```diff
@@ -169,18 +207,24 @@ pub enum QueryItem {
...
     RangeTo(RangeTo<Vec<u8>>),
     RangeToInclusive(RangeToInclusive<Vec<u8>>),
+    RangeAfter(RangeFrom<Vec<u8>>),
+    RangeAfterTo(Range<Vec<u8>>),
+    RangeAfterToInclusive(RangeInclusive<Vec<u8>>),
 }
 
 impl QueryItem {
-    pub fn lower_bound(&self) -> &[u8] {
+    pub fn lower_bound(&self) -> (&[u8], bool) {
```
QuantumExplorer, line 216: The bool here seems to signify not_present, shouldn't we have it be present? like the upper bound?

QuantumExplorer (changes requested):

merk/src/proofs/query/mod.rs
QuantumExplorer: nevermind, i understand now

QuantumExplorer (approved):

PR #29, opened 2022-01-17
feat: transactions support for aux methods

fominok:
This PR adds optional transactions for *_aux family of GroveDB methods


shumkov (approved):
LGTM! 👍

PR #30, opened 2022-01-17
feat: transaction rollback

shumkov:
- Introduced `rollback_transaction` method to rollback a transaction.
- Cleanup temp tree after transaction commitment

### DIFF
--- a/grovedb/src/lib.rs
+++ b/grovedb/src/lib.rs
@@ -476,9 +476,9 @@ impl GroveDb {
     /// transaction usage, please check [`GroveDb::start_transaction`]
     pub fn rollback_transaction(
         &mut self,
-        db_transaction: OptimisticTransactionDBTransaction,
+        db_transaction: &OptimisticTransactionDBTransaction,
     ) -> Result<(), Error> {
-        // Cloning all the trees to maintain original state before the transaction
+        // Cloning all the trees to maintain to rollback transactional changes
         self.temp_root_tree = self.root_tree.clone();
         self.temp_root_leaf_keys = self.root_leaf_keys.clone();
         self.temp_subtrees = self.subtrees.clone();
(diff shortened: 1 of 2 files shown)


grovedb/src/lib.rs
```diff
@@ -461,11 +460,31 @@ impl GroveDb {
...
         Ok(db_transaction
             .commit()
             .map_err(PrefixedRocksDbStorageError::RocksDbError)?)
     }
+
+    /// Rollbacks previously started db transaction. For more details on the
+    /// transaction usage, please check [`GroveDb::start_transaction`]
+    pub fn rollback_transaction(
+        &mut self,
+        db_transaction: &OptimisticTransactionDBTransaction,
```
fominok, line 479: Like `commit_transaction` this method should take ownership of `db_transaction`, because it is rolled back and should be no longer available

grovedb/src/lib.rs
```diff
@@ -461,11 +460,31 @@ impl GroveDb {
...
+    ) -> Result<(), Error> {
+        // Cloning all the trees to maintain to rollback transactional changes
```
fominok, line 481: These are not required either if transaction would be consumed on rollback

Also, it leaves db in readonly mode and there is no way to start new transaction

fominok (changes requested):

grovedb/src/lib.rs
shumkov: Rollback doesn't drop the transaction. It rolls it back to the initial state so you can continue to use it.

grovedb/src/lib.rs
```diff
@@ -461,11 +460,31 @@ impl GroveDb {
         self.subtrees = self.temp_subtrees.drain().collect();
 
         // TODO: root tree actually does support transactions, so this
-        // code can be reworked to account for that
+        //  code can be reworked to account for that
+
+        // Free transactional data
```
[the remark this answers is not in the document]
shumkov: Agreed to keep it for now

grovedb/src/lib.rs
shumkov: Agreed to keep it for now

fominok (dismissed):

fominok (approved):

PR #34, opened 2022-01-18
feat: add `root_hash` method

fominok:
this PR adds `root_hash` method for GroveDb with respect to transactions


shumkov (approved):
LGTM! 👍

PR #38, opened 2022-01-22
refactor: fix more warnings, avoid duplicate code, fix typos

PastaPastaPasta:


fominok (approved):

shumkov (approved):
👍

grovedb/src/lib.rs
```diff
@@ -25,7 +22,7 @@ use subtrees::Subtrees;
 
 /// A key to store serialized data about subtree prefixes to restore HADS
 /// structure
-const SUBTREES_SERIALIZED_KEY: &[u8] = b"subtreesSerialized";
+// const SUBTREES_SERIALIZED_KEY: &[u8] = b"subtreesSerialized";
```
shumkov, line 25: Do we need to commit commented and unused code?

PastaPastaPasta: imo, normally no, but this is a really new project and the standard here seems to be yes

PR #46, opened 2022-01-27
ci: fix NPM auth problem

shumkov:


PastaPastaPasta (approved):
utACK, seems fine. Probably better to use the native action rather then running a command.

PR #50, opened 2022-01-31
build: optimize develop builds

PastaPastaPasta:
This enables full optimization on develop builds (by default), note that debug information is still included, and stuff like debug_assert is still enabled. Everything should work as you'd expect in develop mode

On a clean build, this results in build going from 55 seconds -> 65 seconds, running tests only went from 14.5 seconds -> 4.5 seconds with this change, an incremental build and test (representative of normal development and test setup) 17.277 seconds -> 8.96 seconds. 

I'm pretty confident, a 10 second degredation once is worth 10 seconds every single time we run tests

I did also test other opt-level's, however, opt-level 3 seemed to be the most optimal.


QuantumExplorer (approved):

iammadab (approved):

shumkov:
But probably for CI tests it's not very good.

shumkov:
Please also check nodejs and release workflows

shumkov:
Yeah, it seems it's different profiles. Maybe just set profile in cargo config for our CI jobs? Or you still need to pass the flag in other workflows

QuantumExplorer:
@PastaPastaPasta @shumkov @iammadab @fominok Can this be merged?

QuantumExplorer:
As I got no response I'm going to close this for now.

PR #53, opened 2022-02-01
refactor: mark a number of functions as const

PastaPastaPasta:
When a function can be const, it probably should be const. This way it can be called at compile time, and it makes it clear to the caller that nothing on the heap is being touched.

### DIFF
--- a/merk/src/tree/link.rs
+++ b/merk/src/tree/link.rs
@@ -53,7 +53,7 @@ pub enum Link {
 impl Link {
     /// Creates a `Link::Modified` from the given `Tree`.
     #[inline]
-    pub fn from_modified_tree(tree: Tree) -> Self {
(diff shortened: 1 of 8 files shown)


merk/src/tree/link.rs
```diff
@@ -134,7 +134,15 @@ impl Link {
     /// if any (note: not the height of the referenced tree itself). Return
     /// value is `(left_child_height, right_child_height)`.
     #[inline]
-    pub fn height(&self) -> u8 {
+    pub const fn height(&self) -> u8 {
+        const fn max(a: u8, b: u8) -> u8 {
```
fominok, line 138: is it due to lack of const version of `cmp::max` ?

fominok (approved):

merk/src/tree/link.rs
PastaPastaPasta: correct

PR #56, opened 2022-02-01
fix: bugfix for a transactional raw iter

fominok:
This PR allows to create raw iterator with transaction and fixes subtrees deletion with transactions


QuantumExplorer (approved):
I must admit I didn't understand the macro, would love to understand it tomorrow.

iammadab (approved):
LGTM

PR #59, opened 2022-02-05
feat: better node hashing

QuantumExplorer:
First this pull request uses var ints instead of fixed size (panicking) keys and values.

Next we change the way the hash is calculated from kv_hash = H(key, value) to kv_hash = H(key, Hash(value))
This allows nodes to be returned without returning their value, hence giving much shorter proofs.


shumkov (approved):
LGTM! 👍

PR #60, opened 2022-02-06
fix: compress subtree can no longer be attacked

QuantumExplorer:
There was a small attack possible as length could be bigger than 255 causing a commutativity between a very big path and a small path to result in same value.


QuantumExplorer:
Since all lengths are same size the previous approach should also be working, this PR is less important. I'll ask research team to take a look to see what is better.

PR #62, opened 2022-02-07
fix: delete_up_if_empty without transaction

iammadab:
when opening merk, if opened merk has no keys, then either the merk
doesn't not exist or it is empty. To resolve this, we check if the
parent merk contains it, if it does then empty, if it doesn't then
doesn't exist.

The problem with deleting the main subtree first before the child
subtress is that, if the child subtree is empty, we need the parent
to confirm the status, but if it's deleted there is no way to
leading to an error.


QuantumExplorer (approved):

fominok (approved):
nice catch

PR #63, opened 2022-02-08
fix(root_tree): non-deterministic root hash

shumkov:


grovedb/src/lib.rs
```diff
@@ -388,7 +388,9 @@ impl GroveDb {
         // TODO: root tree actually does support transactions, so this
         //  code can be reworked to account for that
         self.root_tree = self.temp_root_tree.clone();
-        self.root_leaf_keys = self.temp_root_leaf_keys.drain().collect();
+
+        self.root_leaf_keys = self.temp_root_leaf_keys.clone();
```
QuantumExplorer, line 392: @fominok I have a feeling there is a better way, couldn't figure it out myself though

QuantumExplorer (approved):

PR #68, opened 2022-02-09
fix: persist root leaves data

fominok:
At some moment we stopped to persist root leaves data in a storage that's why on GroveDB reopening we cannot get an appropriate root hash or to print any info deeper than just "root"


shumkov (approved):
LGTM! 👍

PR #69, opened 2022-02-10
fix: do not remove subtree from temp_subtrees

fominok:
This PR simplifies subtrees usage returning a kind of a pointer to Merk when accessing data (temporary or not), so no need to insert it back;

Note that as it uses `RefCell` internally multiple sequential accesses to subtrees without dropping previous borrows will cause a panic, to simplify "oneshot" usage of a subtree there is an `apply` helper

Built prefix if presents could be accessed with `get_prefix`


iammadab (approved):
Amazing!

QuantumExplorer (approved):

QuantumExplorer:
Very nice

PR #72, opened 2022-02-16
feat: better path errors

QuantumExplorer:
This PR decomposes path errors into 4 categories.
InvalidPath : the path is not valid/does not exist, and it represents a logical error by the client.
PathNotFound: the path does not exist, but it doesn't represent a logical error by the client. Meaning that the query was valid, but an intermediate path was just not present.
PathKeyNotFound: the path was found, but the key was not, the query was valid but there was nothing there.
CorruptedPath : GroveDB internal error, not a logical error by the client.

### DIFF
--- a/grovedb/src/lib.rs
+++ b/grovedb/src/lib.rs
@@ -56,7 +56,6 @@ pub enum Error {
     CorruptedPath(&'static str),
 
     // Query errors
-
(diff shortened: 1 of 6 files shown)


shumkov (approved):
LGTM! 👌

PR #76, opened 2022-02-27
refactor: Clippy nursery fixes

PastaPastaPasta:


fominok:
Why to remove `pub(crate)` and make it a fully public API when not needed?

PastaPastaPasta:
> Why to remove `pub(crate)` and make it a fully public API when not needed?

@fominok 
image

fominok (approved):

PR #80, opened 2022-03-07
fix: IN query issue in RS-Drive

QuantumExplorer:
When an item was not found, instead of not returning that item an error would be thrown, this would cause in queries to fail if one item was missing. This is fixed with this PR.


shumkov (approved):
Good catch! 👍

PR #81, opened 2022-03-08
refactor: new storage approach

fominok:
### Why
This PR addresses a number of usability and robustness problems in GroveDb:

1. We cared about transactional data (temp root tree, temp subtrees, deleted subtrees) instead of letting a storage do the work
2. Only one transaction was supported, some kind of "read-only" mode was introduced
3. Subtrees were aware of transaction and should handle it accordingly
4. If transaction doesn't handled manually GroveDb will be in a broken state
5. used nightly features

Also this PR acts as a preparation to support batching between many subtrees.

### How does it work
Instead of having one `Storage` trait now it will be split into two: `Storage` and `StorageContext`.

**Storage**: highlevel storage abstraction should manage transactions (and multi-tree batches in future) and providing `StorageContext` (more on that later).

**StorageContext**: an interface to a storage that handles such details as transaction and in case of RocksDb also a prefix, meant to be used in subtrees so that Merk will be unaware of any unnecessary details. This allows us to open Merk on different prefixes with different transactions (and batches in future).

### What is left to do
1. GroveDb on its top-levels still depends on RocksDb, consider moving `get_*context*` methods into a `Storage` trait
2. Merk is generic over `StorageContext` statically, that means `Merk<Context>` is not `Merk<TxContext>` and we have to handle it differently, may blowup when we add batches, maybe we should use dynamic dispatch


QuantumExplorer (approved):

iammadab (approved):
👍🏾

PR #82, opened 2022-03-10
feat: addition of more range types

iammadab:
Merk currently supports proofs for:
- key
- range
- range_inclusive

This pr adds:
- range_to
- range_to_inclusive
- range_from
- range_after
- range_after_to
- range_after_to_inclusive
- range_full

### DIFF
--- a/merk/src/proofs/query/mod.rs
+++ b/merk/src/proofs/query/mod.rs
@@ -1993,7 +1993,7 @@ mod test {
         let mut tree = make_6_node_tree();
         let mut walker = RefWalker::new(&mut tree, PanicSource {});
 
-        let queryitems = vec![QueryItem::RangeAfterTo(vec![5]..vec![7])];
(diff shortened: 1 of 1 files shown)


fominok (approved):

merk/src/proofs/query/mod.rs
```diff
@@ -740,15 +740,15 @@ where
 
                 // if range starts before this node's key, include it in left
                 // child's query
-                let left_query = if left_bound < self.tree().key() {
+                let left_query = if left_bound < self.tree().key() || left_bound == b"" {
```
QuantumExplorer, line 743: Let's have left and right bounds become Options, it will be a lot more clean.

QuantumExplorer (changes requested):

fominok (approved):

QuantumExplorer (approved):

PR #83, opened 2022-03-11
feat: ability to convert non-inclusive-range-type to their inclusive version

iammadab:
**Problem with non inclusive range types**

A query is a collection of sorted unique keys

given a node, all elements that stem from it’s right edge are greater than it, all elements that stem from the left edge are smaller than it.

when a query is applied to a node, you want to split the query into two, (elements less than the current node) and (elements greater than the current node) and then pass those keys to the corresponding subtrees.

range queries makes things more interesting as they are a collection of keys treated as a single entity. 

Take a node key 5 and a range 3..=8

the range contains 5, but also has elements greater than and less than 5, the range is passed both to the left subtree and the right subtree as it contains elements of interest for both trees.

left - (3..=8) right - (3..=8)

take another range 3..=5 (it has elements smaller than 5, but no element larger than 5 hence no reason to pass the range to the right subtree)

The previous algorithm used to split the query is given below:

                let item = &query[index];
                
                // get the bounds of the query item
                let left_bound = item.lower_bound();
                let right_bound = item.upper_bound().0;
      
                // check if the range still has elements of interest for the left subtree
                let left_query = if left_bound < self.tree().key() {
                    // if it does include it
                    &query[..=index]
                } else {
                    // doesn't, exclude it
                    &query[..index]
                };

                 // check if the range still has elements of interest for the right subtree
                let right_query = if right_bound > self.tree().key() {
                    &query[index..]
                } else {
                    &query[index + 1..]
                };

                (left_query, right_query)

When the query item is a non inclusive range type e.g. 3..5, the algorithm thinks the bounds are [3-5] instead of [3-4], leading to the same proof for 3..=5. (which isn't correct behaviour)

This PR implements a solution that extracts the exact bound (converting it to an inclusive range), before making bounds critical decision.

To achieve this, I made the assumption that byte arrays increase and decrease by 1 bit.

### DIFF
--- a/merk/src/proofs/query/mod.rs
+++ b/merk/src/proofs/query/mod.rs
@@ -1010,10 +1010,10 @@ mod test {
         *,
     };
     use crate::{
+        proofs::query::QueryItem::RangeAfter,
         test_utils::make_tree_seq,
         tree::{NoopCommit, PanicSource, RefWalker, Tree},
     };
-    use crate::proofs::query::QueryItem::RangeAfter;
 
     fn make_3_node_tree() -> Tree {
         let mut tree = Tree::new(vec![5], vec![5])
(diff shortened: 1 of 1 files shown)


merk/src/proofs/query/mod.rs
```diff
@@ -301,6 +301,49 @@ impl QueryItem {
         }
     }
 
+    pub fn to_inclusive(&self) -> Self {
+        fn increment_byte_array(input: Vec<u8>) -> Vec<u8> {
+            let mut input = input.clone();
```
fominok, line 306: is it actually needed to be cloned?

fominok (approved):

PR #89, opened 2022-03-16
feat: new node type Node::KVDigest

iammadab:
Node::KVDigest is of the form (key, Hash(value))

This will be used to represent nodes not requested in the 
query but are still needed to prove the absence of another key.


merk/src/proofs/encoding.rs
```diff
@@ -36,6 +43,7 @@ impl Encode for Op {
         Ok(match self {
             Op::Push(Node::Hash(_)) => 1 + HASH_LENGTH,
             Op::Push(Node::KVHash(_)) => 1 + HASH_LENGTH,
+            Op::Push(Node::KVDigest(key, _)) => 2 + key.len() + HASH_LENGTH,
```
QuantumExplorer, line 46: Just curious, why 2?

QuantumExplorer (approved):

merk/src/proofs/encoding.rs
iammadab: 0x04 and key.len() as u8 take 1 byte each (making two bytes)
rest is key and value_hash which  are accounted for with the rest of the equation (key.len() and HASH_LENGTH)

QuantumExplorer: Got it

PR #94, opened 2022-03-22
feat: var int serialization

QuantumExplorer:
This reduces the serialization size of an element.
It also attempts to add methods to figure out the disk space footprint of an element.

### DIFF
--- a/grovedb/src/subtree.rs
+++ b/grovedb/src/subtree.rs
@@ -2,6 +2,10 @@
 //! Subtrees handling is isolated so basically this module is about adapting
 //! Merk API to GroveDB needs.
 
+use bincode::{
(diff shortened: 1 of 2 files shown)


grovedb/src/subtree.rs
```diff
@@ -51,6 +55,54 @@ impl Element {
         Element::Tree(Default::default())
     }
 
+    /// Get the size of an element in bytes
+    pub fn byte_size(&self) -> usize {
+        match self {
+            Element::Item(item) => item.len(),
+            Element::Reference(path_reference) => {
+                path_reference.iter().map(|inner| inner.len()).sum()
```
fominok, line 63: I think this is inaccurate, vector occupies more than just its contents

fominok (approved):

iammadab (approved):

PR #95, opened 2022-03-23
feat: generate right to left merk proofs

iammadab:
Adds capability to set proof direction (ascending, descending) over a range and still works with limits and offsets. 
allows for depth first traversal on the right or left side of the tree depending on the chosen direction
added new opcode PushInverted to signify (right edge depth traversal)


merk/src/proofs/query/mod.rs
```diff
@@ -975,60 +1047,118 @@ pub fn verify_query(
...
-                    // continue to the next push
-                    break;
+                if left_to_right {
+                    // we have not reached next queried part of tree
+                    if *query_item > key.as_slice() {
+                        // continue to next push
+                        break;
+                    } else if start_non_inclusive
+                        && lower_bound.is_some()
+                        && lower_bound.unwrap() == key.as_slice()
```
fominok, line 1059: I see you used `Some(key.as_slice()) == x` somewhere below, will it work here also instead of `unwrap` / `is_some` combo?

merk/src/proofs/query/mod.rs
```diff
@@ -975,60 +1047,118 @@ pub fn verify_query(
...
+                        // continue to the next push
+                        break;
+                    }
+                } else {
+                    if *query_item < key.as_slice() {
+                        // continue to next push
+                        break;
+                    } else if !end_inclusive
+                        && upper_bound.is_some()
+                        && upper_bound.unwrap() == key.as_slice()
```
fominok, line 1071: Same as above

fominok (approved):

PR #96, opened 2022-03-25
feat(storage): multi-context batches

fominok:
This PR adds two storage contexts: one for cross-subtrees batches and another to do it inside transactions.

Note: while this adds a technical ability to apply changes to multiple subtrees in a single pass it's still not integrated with GroveDB


shumkov (approved):
Looks good to me 👍 but I'm not an expert in Rust though

iammadab (approved):

PR #102, opened 2022-04-04
feat: construction and verification of path query proofs with subquery

iammadab:
when you apply a query to a subtree you get a result set
if the query has a subquery, you need to apply that to each element in the result set
the subquery might also have a subquery .....
this pr adds the ability for proofs to take this into account


grovedb/src/operations/proof.rs
```diff
@@ -92,41 +142,50 @@ impl GroveDb {
...
                     None,
                     query.query.query.left_to_right,
-                )
-                .expect("should execute proof");
-                let result_set = proof_result.1.result_set;
+                )?;
 
-                if result_set[0].0 != key.to_vec() {
+                let result_set = proof_result.1.result_set;
+                if result_set.len() == 0 || result_set[0].0 != key.to_vec() {
```
fominok, line 185: will it work if we take slice from `result_set[0].0` rather than making a vector out of `key` ?

grovedb/src/operations/proof.rs
```diff
@@ -135,41 +194,357 @@ impl GroveDb {
...
+        db: &GroveDb,
+        proofs: &mut Vec<u8>,
+        path: Vec<&[u8]>,
+        query: PathQuery,
+        current_limit: &mut Option<u16>,
+        current_offset: &mut Option<u16>,
+    ) -> Result<(), Error> {
+        // there is a chance that the subquery key would lead to something that is not a
+        // tree same thing for the subquery itself
+        merk_optional_tx!(db.db, path.clone(), None, subtree, {
```
fominok, line 248: `merk_optional_tx` macro is made to generate code that will use transactional context in case of `Some(transaction)` and a regular one in `None` case;

As transaction will be always `None`, the branch generated by macro for tx case is useless, so macro is not needed in such case

Howerver, it's a nice shortcut to create a storage context and open a Merk on it, but it still doesn't feel right

grovedb/src/operations/proof.rs
```diff
@@ -135,41 +194,357 @@ impl GroveDb {
...
+                for (key, value_bytes) in verification_result.1.result_set {
+                    let child_element = Element::deserialize(value_bytes.as_slice())?;
+                    match child_element {
+                        Element::Tree(mut expected_root_hash) => {
+                            if expected_root_hash == EMPTY_TREE_HASH {
+                                // child node is empty, move on to next
+                                continue;
+                            }
+
+                            if current_limit.is_some() && current_limit.unwrap() == 0 {
```
fominok, line 441:
```suggestion
                            if current_limit == Some(0) {
```

fominok (approved):

grovedb/src/operations/proof.rs
iammadab: fixed in #103

grovedb/src/operations/proof.rs
iammadab: added a todo to fix this in the latest branch

PR #104, opened 2022-04-28
feat: bidirectional references

iammadab:
referenced items should keep track of all elements that reference them
this allows efficient updates to the references when the base item changes (updated or deleted)

When a base item gets deleted all its references should also get deleted.
When a reference gets deleted, the base item should stop pointing to it.


charlestonXdao (approved):
are we there yet?

fominok:
closing this in favor of #345

PR #108, opened 2022-05-23
fix: subquery key logic + lazy key value iterator

iammadab:


merk/src/merk/mod.rs
```diff
@@ -45,6 +45,84 @@ impl ProofWithoutEncodingResult {
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(query_item) = self.current_query_item {
+            let kv_pair = self.get_kv(&query_item);
+
+            if kv_pair.is_some() {
+                return kv_pair;
+            } else {
+                self.query_position += 1;
+                self.seek();
```
QuantumExplorer, line 117: This seems quite inefficient if you are already at the right position in the iterator.

PR #109, opened 2022-05-26
feat: element flags

iammadab:
API consumers can add an optional one-byte flag (meta-data) to elements on insertion.
Updated the element instantiation interface, and created separate methods for creating elements with and without flags.

### DIFF
--- a/grovedb/src/subtree.rs
+++ b/grovedb/src/subtree.rs
@@ -62,12 +62,24 @@ impl Element {
         Element::Item(item_value, None)
     }
 
+    pub fn new_item_with_flag(item_value: Vec<u8>, flag: ElementFlag) -> Self {
+        Element::Item(item_value, flag)
+    }
+
     pub fn new_reference(reference_path: Vec<Vec<u8>>) -> Self {
         Element::Reference(reference_path, None)
     }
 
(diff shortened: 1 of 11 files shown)


grovedb/src/subtree.rs
```diff
@@ -17,19 +17,22 @@ use crate::{
     Error, Merk, PathQuery, SizedQuery, TransactionArg,
 };
 
+/// Optional single byte meta-data to be stored per element
+type ElementFlag = Option<u8>;
```
QuantumExplorer, line 21: The flag should be an array of bytes, not just 1 byte.

grovedb/src/subtree.rs
```diff
@@ -52,28 +55,58 @@ where
...
+        Element::Tree(tree_hash, flag)
     }
 
     /// Get the size of an element in bytes
     pub fn byte_size(&self) -> usize {
         match self {
-            Element::Item(item) => item.len(),
-            Element::Reference(path_reference) => {
-                path_reference.iter().map(|inner| inner.len()).sum()
+            // +1 for 1 byte flag
```
QuantumExplorer, line 88: Flag needs to be more than 1 byte

QuantumExplorer (changes requested):

QuantumExplorer (approved):

PR #112, opened 2022-06-06
feat: return keys on query

QuantumExplorer:
In some situations returning the keys for a range query are quite important.

Also since we always have them in memory, might as well always return them.


iammadab (approved):

shumkov (approved):
LGTM! 👌

PR #122, opened 2022-06-12
fix: made some fields public of groveDB Op public

QuantumExplorer:


QuantumExplorer:
Just going to merge this one is as I can't see it being controversial in any way.

PR #131, opened 2022-06-24
feat: batches and start of worst case fees

QuantumExplorer:


grovedb/src/batch.rs
```diff
@@ -530,10 +633,59 @@ impl GroveDb {
...
+        // nothing for the commit multi batch?
+        Ok(()).wrap_with_cost(cost)
+    }
 }
 
 #[cfg(test)]
 mod tests {
+    use std::option::Option::None;
+
+    use visualize::visualize_stderr;
```
fominok, line 687: These two imports are not needed

grovedb/src/lib.rs
```diff
@@ -64,6 +64,12 @@ pub enum Error {
     #[error("data corruption error: {0}")]
     CorruptedData(String),
 
+    #[error("corrupted code execution error: {0}")]
+    CorruptedCodeExecution(&'static str),
```
fominok, line 68: What is that error about? Cannot see any usages

storage/src/rocksdb_storage/storage.rs
```diff
@@ -82,6 +82,34 @@ impl RocksDbStorage {
         res = blake3::hash(&res).as_bytes().to_vec();
         res
     }
+
+    // /// A helper method to build a prefix to rocksdb keys or identify a subtree
```
fominok, line 86: I think this could be removed

fominok (approved):

storage/src/rocksdb_storage/storage.rs
QuantumExplorer: Hmmmm, best to keep it for now, just in case.

PR #138, opened 2022-06-28
feat: more worst case

QuantumExplorer:


fominok:
I guess that could be closed

PR #151, opened 2022-07-23
feat: query return different types of results

QuantumExplorer:
Sometimes on a query we are interested in the value, sometimes the key value, and even some other times the path key and the value. This PR allows us to specify the return type of items.

### DIFF
--- a/grovedb/src/lib.rs
+++ b/grovedb/src/lib.rs
@@ -3,6 +3,7 @@ extern crate core;
 pub mod batch;
 mod operations;
 mod query;
+pub mod query_result_type;
 mod subtree;
 #[cfg(test)]
 mod tests;
--- a/grovedb/src/query_result_type.rs
+++ b/grovedb/src/query_result_type.rs
@@ -9,31 +9,53 @@ pub enum QueryResultType {
 
 pub type QueryResultItems = Vec<QueryResultItem>;
 
-pub fn query_result_items_to_elements(query_result_items: QueryResultItems) -> Vec<Element> {
+pub trait GetItemResults {
+    fn to_elements(self) -> Vec<Element>;
+    fn to_key_elements(self) -> Vec<KeyElementPair>;
+    fn to_path_key_elements(self) -> Vec<PathKeyElementTrio>;
+}
+
+impl GetItemResults for QueryResultItems {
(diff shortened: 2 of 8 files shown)


grovedb/src/query_result_type.rs
```diff
@@ -0,0 +1,83 @@
...
+#[derive(Copy, Clone)]
+pub enum QueryResultType {
+    QueryElementResultType,
+    QueryKeyElementPairResultType,
+    QueryPathKeyElementTrioResultType,
+}
+
+pub type QueryResultItems = Vec<QueryResultItem>;
+
+pub trait GetItemResults {
```
fominok, line 12: why to use a trait which will be never used again instead of a struct with methods?

grovedb/src/lib.rs
```diff
@@ -3,6 +3,7 @@ extern crate core;
 pub mod batch;
 mod operations;
 mod query;
+pub mod query_result_type;
```
fominok, line 6: nice, can't wait to see when all query stuff will go away from `subtree` module

grovedb/src/query_result_type.rs
```diff
@@ -0,0 +1,83 @@
...
+            QueryResultItem::ElementResultItem(_) => None,
+            QueryResultItem::KeyElementPairResultItem(key_element_pair) => Some(key_element_pair),
+            QueryResultItem::PathKeyElementTrioResultItem(path_key_element_trio) => {
+                Some((path_key_element_trio.1, path_key_element_trio.2))
+            }
+        })
+        .collect()
+}
+
+fn query_result_items_to_path_key_elements(
```
fominok, line 58: Honestly I don't see a reason to have methods' implementations as separate functions called from a trait;

Also this may done as `Into` trait implementation I guess

grovedb/src/query_result_type.rs
QuantumExplorer: Because it's on Vec, so we are extending Vec with this Trait. Or at least that was my logic.

fominok: I believe it's more than vec now, it deserves a distinct type for it (with methods and conversions)

fominok (approved):

PR #153, opened 2022-07-28
optimisation: unnecessary hashing

iammadab:
When creating or updating a kv node, the value hash is computed twice. This PR removes the unnecessary extra hashing step.


QuantumExplorer (approved):

shumkov:
Conventional comit has type "perf" for such commits https://www.conventionalcommits.org/en/v1.0.0/

PR #154, opened 2022-07-28
GroveDB replication

fominok:
This PR adds replication capability for GroveDB.

It is designed a bit different than just a stream of chunks with a main difference that receiver is who actually drives the replication process, that's because GroveDb is not a binary tree and it would involve some more proofs of how it's organized and might be unnecessarily complicated.

1. Chunks producer and Restorer are initialized, the former from existing GroveDb, the latter works on an empty storage;
2. Restorer processes first chunk of root merk tree and returns a request;
3. Restorer asks for a specific chunk of a specific subtree;
4. If no more things to do restorer returns `Ready` variant.

So basically we just replicate each Merk individually going from the top.

p.s. pay no attention to the branch name


shumkov:
We could have millions of small trees. Won't approach 1 chunk = 1 tree be super slow due to a big number of trees/chunks and overhead which each chunk replication has?

fominok:
> We could have millions of small trees. Won't approach 1 chunk = 1 tree be super slow due to a big number of trees/chunks and overhead which each chunk replication has?

It will, but we'll add another layer next

QuantumExplorer (approved):

shumkov:
Oh, I see.

iammadab (approved):

PR #156, opened 2022-08-04
Bigger chunks

fominok:
This PR adds a wrapper to chunk producer to pack multiple chunks into one message.

Depending on a constant setting one "Grove chunk" may include multiple chunks of one Merk or even multiple Merks if they meet one of these conditions:
1. It's a full root tree and root leaf subtree(s) starting from left,
2. It's a requested subtree's chunk(s) with it's siblings on the right.


iammadab (approved):

QuantumExplorer:
Very quick review on my part.

PR #157, opened 2022-08-05
fix: merging equal path queries

shumkov:
Merging equal subqueries is not allowed for some reason. I didn't find any reason for that. Waiting for @iammadab to discuss.


iammadab:
The problem with this is that different-sized path queries can be converted to the same form.
So if we have:
- [a, b, Query1]
- [a, b, c, Query2]

The second path query can be converted to this form:
- [a, b, Query3] where Query3 = [c, Query2]

Looking at [a, b, Query1] and [a, b, Query3] we see them as equal when they are not.

To correctly solve this, we just have to do a further check on equality to confirm that none of the path queries have a subquery.

shumkov:
@QuantumExplorer updated the PR, so I guess it's ready for review now.

iammadab:
Looks good from grovedb's perspective, we just have to make sure it supports all of drive's needs.

iammadab (approved):

PR #159, opened 2022-08-10
Merk Sum trees

QuantumExplorer:
Merk Sum trees sum up children nodes to their parents allowing the sum of the tree to represent the total amount in the tree.

Each node is calculated as Sum at Node = Node value + left child value + right child value.


QuantumExplorer:
#202 replaces this

PR #161, opened 2022-08-16
perf: insert reference optimization for worst case

iammadab:
For non-batch insert reference 
- Single get call and no node hashing

For batch insert reference
- Clients can specify application-specific max reference hop (allowing worst-case cost to be closer to actual cost)

### DIFF
--- a/grovedb/src/batch.rs
+++ b/grovedb/src/batch.rs
@@ -2,16 +2,14 @@
 
 use core::fmt;
 use std::{
-    borrow::Cow,
     cmp::Ordering,
     collections::{btree_map::Entry, BTreeMap, HashMap, HashSet},
-    hash::Hash,
 };
 
 use costs::{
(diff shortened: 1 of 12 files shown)


grovedb/src/batch.rs
```diff
@@ -415,29 +415,34 @@ where
...
+
                         let referenced_element = cost_return_on_error!(
                             &mut cost,
                             self.follow_reference(
                                 path_reference,
                                 ops_by_qualified_paths,
-                                MAX_REFERENCE_HOPS as u8
+                                element_max_reference_hop
+                                    .or(Some(MAX_REFERENCE_HOPS as u8))
+                                    .expect("should have a value as MAX_REFERENCE_HOP has a value")
```
QuantumExplorer, line 433: I think unwrap_or_default is better here.

QuantumExplorer (changes requested):

QuantumExplorer (approved):
Just need a little more comments around the complicated parts of batches, but looks good.

PR #164, opened 2022-08-22
test: batch insert to root should be two seeks

QuantumExplorer:
One seek for getting the merk tree of the root, and one seek for insertion.


QuantumExplorer:
Also changed root key key from "root" to "r"

iammadab (approved):

PR #170, opened 2022-08-24
feat: return result during tree decode

iammadab:
Decoding operation would panic if input bytes are invalid, this handles such cases.


merk/src/merk/mod.rs
```diff
@@ -631,7 +631,7 @@ where
 fn fetch_node<'db>(db: &impl StorageContext<'db>, key: &[u8]) -> Result<Option<Tree>> {
     let bytes = db.get(key).unwrap()?; // TODO: get_pinned ?
     if let Some(bytes) = bytes {
-        Ok(Some(Tree::decode(key.to_vec(), &bytes)))
+        Ok(Some(Tree::decode(key.to_vec(), &bytes)?))
```
QuantumExplorer, line 634: shouldn't this just be a map?

QuantumExplorer (changes requested):

PR #171, opened 2022-08-24
feat: populate key value storage cost

iammadab:
For better cost UX, written storage cost should be split into added_storage, replaced_storage and removed_storage
where replaced_storage is much less than added_storage.
This PR modifies merk to return the information needed to compute the added, replaced and removed storage

### DIFF
--- a/costs/src/lib.rs
+++ b/costs/src/lib.rs
@@ -137,7 +137,7 @@ impl OperationCost {
         let (key_storage_cost, value_storage_costs) = match storage_cost_info {
             None => (None, None),
             Some(s) => {
-                s.key_storage_cost.verify(key_len)?;
+                // s.key_storage_cost.verify(key_len)?;
                 s.value_storage_cost.verify(value_len)?;
                 (Some(s.key_storage_cost), Some(s.value_storage_cost))
             }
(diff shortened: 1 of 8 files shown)


merk/src/merk/mod.rs
```diff
@@ -714,12 +717,48 @@ impl MerkCommitter {
...
+        };
+        let mut value_storage_cost = StorageCost {
+            ..Default::default()
+        };
+
+        // Update the value storage cost
+        if tree.old_size == 0 {
+            // new node, storage has to be created for entire tree
+            value_storage_cost.added_bytes += (current_tree_size) as u32
+        } else if current_tree_size > tree.old_size {
```
QuantumExplorer, line 739: I would add a == even though it works, just to be a little more clear.

costs/src/lib.rs
```diff
@@ -51,13 +53,23 @@ impl StorageCost {
     pub fn verify(&self, len: u32) -> Result<(), Error> {
```
QuantumExplorer, line 53: let's have verify_key and verify_value

QuantumExplorer (changes requested):

QuantumExplorer (approved):

PR #177, opened 2022-09-01
fix: batch operation consistency check

iammadab:
We had the batch operation consistency checker, but it wasn't used. Enabled it and added tests for verify it's correctness.


QuantumExplorer:
Can you change this to only happen if an option is present in the BatchApplyOptions?

QuantumExplorer (approved):

PR #188, opened 2022-10-12
feat: add reference value to state

iammadab:
Before this pr, reference values didn't affect the state root. This means that if a reference points to a value at point A, and then gets updated to point to a duplicate value at point B, the state root doesn't change as the value hash is maintained.


merk/src/proofs/chunk.rs
```diff
@@ -328,6 +328,7 @@ mod tests {
         kv: usize,
         kvvaluehash: usize,
         kvdigest: usize,
+        kvvaluerefhash: usize,
```
fominok, line 331: shouldn't these fields be in snake_case?

PR #189, opened 2022-10-20
feat: return kvvaluehash from merk

iammadab:


iammadab:
Changes already in cost branch.

PR #192, opened 2022-10-25
fix: replication fails after combined root hash

iammadab:


grovedb/src/replication.rs
```diff
@@ -138,7 +138,7 @@ impl<'db> Restorer<'db> {
                 Merk::open_base(grove_db.db.get_storage_context(empty()).unwrap())
                     .unwrap()
                     .map_err(|e| RestorerError(e.to_string()))?,
-                root_key,
+                None,
```
QuantumExplorer, line 141: Why?

merk/src/merk/restore.rs
```diff
@@ -121,14 +122,14 @@ impl<'db, S: StorageContext<'db>> Restorer<S> {
     /// Verifies the trunk then writes its data to the RocksDB.
     fn process_trunk(&mut self, ops: impl IntoIterator<Item = Op>) -> Result<usize, Error> {
         let (trunk, height) = verify_trunk(ops.into_iter().map(Ok)).unwrap()?;
-        if trunk.key() != self.expected_root_key {
```
QuantumExplorer, line 124: why remove this?

merk/src/merk/restore.rs
```diff
@@ -286,8 +287,8 @@ impl<'db, S: StorageContext<'db>> Merk<S> {
     /// Creates a new `Restorer`, which can be used to verify chunk proofs to
     /// replicate an entire Merk tree. A new Merk instance will be initialized
     /// by creating a RocksDB at `path`.
-    pub fn restore(merk: Merk<S>, expected_root_key: Vec<u8>, expected_root_hash: CryptoHash) -> Restorer<S> {
-        Restorer::new(merk, expected_root_key, expected_root_hash)
```
QuantumExplorer, line 290: Why none?

grovedb/src/replication.rs
iammadab: The root tree is not bounded by an Element::Tree node as such it doesn't need a combining value.

merk/src/merk/restore.rs
iammadab: It was originally trunk.hash().unwrap() != self.expected_root_hash, but this was comparing combined hash with underlying merk hash.
I am now checking if it's a non-root tree and generating the combined hash before comparing.

merk/src/merk/restore.rs
iammadab: because chunking starts from the root tree and at that point, there is no combining value since it is not bounded by a tree element.

PR #249, opened 2023-04-27
improve: merk proofs adr

iammadab:
A document describing how Merk proofs work 

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

### DIFF
--- a/adr/merk-proofs.md
+++ b/adr/merk-proofs.md
@@ -599,8 +599,6 @@ Simply put, a Merk proof encodes a subset of the tree state that contains all th
 
 Proof generation algorithm:
 
-TODO: consider changing query set to batches might be easier to understand
-
 - Given a node and a set of keys to include in the proof:
     - if the query set is empty, append `Op::Push(Node::Hash(node_hash))` to the proof and return
         - since the query set is empty, it means the current node is not part of our requested keys and we also don’t care about any of it’s child nodes
--- a/adr/merk-proofs.md
+++ b/adr/merk-proofs.md
@@ -499,6 +499,7 @@ Op::Push(Node::KV(1, v1))
 Op::Push(Node::KvHash(2b)) // kv hash of node 2
 Op::Parent
 Op::Push(Node::Hash(4d3c5e)) // node hash of node 4
+Op::Child
 ```
 
 #### Explanation
--- a/adr/merk-proofs.md
+++ b/adr/merk-proofs.md
@@ -2,7 +2,7 @@
 
 Merk is a Merkle AVL tree ([https://en.wikipedia.org/wiki/AVL_tree](https://en.wikipedia.org/wiki/AVL_tree))
 
-******************Structure******************
+#### Structure
 
 A node in a merk tree represents a key-value pair, each node can optionally have a left and right child node. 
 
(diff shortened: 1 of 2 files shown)


adr/merk-proofs.md
```diff
@@ -0,0 +1,725 @@
+# Merk Proofs
+
+Merk is a Merkle AVL tree ([https://en.wikipedia.org/wiki/AVL_tree](https://en.wikipedia.org/wiki/AVL_tree))
+
+******************Structure******************
```
fominok, line 5: what this kind of markup stands for? It looks like it should be a level 2 subheading (`## Structure`)

adr/merk-proofs.md
```diff
@@ -0,0 +1,725 @@
...
+```
+
+and the encoding will look like
+
+```rust
+Op::Push(Node::KV(1, v1))
+Op::Push(Node::KvHash(2b)) // kv hash of node 2
+Op::Parent
+Op::Push(Node::Hash(4d3c5e)) // node hash of node 4
+```
```
fominok, line 501: Does is miss `Op::Child` ?

adr/merk-proofs.md
```diff
@@ -0,0 +1,725 @@
...
+
+query = [QueryItem::Range(2..4), QueryItem::RangeInclusive(3..=5)]
+
+QueryItem::Range(2..4) selects for [2, 3]
+
+QueryItem::RangeInclusive(3..=5) selects for: [3. 4. 5]
+
+an overlap exists, we merge the sets: [2, 3, 4, 5]
+
+This is equivalent to QueryItem::RangeInclusive(2..=5), hence the query becomes
```
fominok, line 584: > becomes

It's unclear to me if a query optimizer fixes an overlapping query for us or it will return an error in case of overlap and we should fix it ourselves?

adr/merk-proofs.md
```diff
@@ -0,0 +1,725 @@
...
+- Complete (exact set, no element is added or removed)
+- Correct (result set only contains data that was added to the state at some point in the past)
+- Fresh (result set is based on the latest data)
+
+Simply put, a Merk proof encodes a subset of the tree state that contains all the requested keys and the operations which, when performed, reconstruct a tree that produces the same root hash as the original tree.. 
+
+Proof generation algorithm:
+
+- Given a node and a set of keys to include in the proof:
+    - if the query set is empty, append `Op::Push(Node::Hash(node_hash))` to the proof and return
```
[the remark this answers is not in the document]
fominok: Oh wait, what do you mean by "query set"?

adr/merk-proofs.md
iammadab: fixed

adr/merk-proofs.md
iammadab: You are right

adr/merk-proofs.md
iammadab: Yes it does, we check if there is an overlap then we merge. 
See: https://github.com/dashpay/grovedb/blob/master/merk/src/proofs/query/insert.rs#L151

adr/merk-proofs.md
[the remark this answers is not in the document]
iammadab: Yes, the node_hash of the root of the current tree. 
It's a recursive function so might be called at a later time with a subset of the tree, same applies if we have no query for that subset return the node_hash of it's root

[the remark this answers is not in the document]
iammadab: The query set is a vector of query items.
e.g [QueryItem::Key(a), QueryItem::Range(b..d)]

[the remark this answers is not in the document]
iammadab: replaced with query

fominok:
todo: update proofs section

PR #265, opened 2023-07-21
feat: raw iter data wipe

iammadab:
Clear GroveDB data by deleting each key value pair in the rocksdb column families. 
This is not efficient and should be revised, a constant time solution is desired. 

## Checklist:

### DIFF
--- a/grovedb/src/lib.rs
+++ b/grovedb/src/lib.rs
@@ -249,7 +249,7 @@ impl GroveDb {
         Ok(GroveDb { db })
     }
 
-    /// Deletes GroveDB folder from disk and drop GroveDB instance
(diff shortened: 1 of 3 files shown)


grovedb/src/lib.rs
```diff
@@ -250,7 +250,7 @@ impl GroveDb {
     }
 
     /// Deletes GroveDB folder from disk and drop GroveDB instance
```
fominok, line 252: all good but a comment needs to be updated

fominok (approved):

grovedb/src/lib.rs
iammadab: got it thanks.

QuantumExplorer (changes requested):

PR #277, opened 2023-10-29
fix!: limit will now decrease when subquery has no elements

QuantumExplorer:
When Querying Grovedb using subqueries and with a limit the limit would only decrease if the subquery actually had an item. However in most cases the limit is there to protect against very expensive queries.

The queries would be expensive because we could go through many many trees where the sub elements have no matches, hence the limit would not decrease and hence we would continue on the increasingly expensive query.

## What was done?
Created an option to allow to decrease the limit by one per iteration if subqueries have no items on them.

## How Has This Been Tested?
Updated tests and debugged to make sure it was working

## Checklist:

### DIFF
--- a/grovedb/src/element/query.rs
+++ b/grovedb/src/element/query.rs
@@ -60,6 +60,20 @@ use crate::{
 use crate::{Element, SizedQuery};
 
 #[cfg(feature = "full")]
+#[derive(Copy, Clone, Debug)]
+pub struct QueryOptions {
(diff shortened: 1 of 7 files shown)


QuantumExplorer:
This is being used in production for a while now, however it is not being used by proofs for now (as far as I am aware). I am merging this PR in for now, however more attention needs to taken to make sure such attacks can not happen on proofs.

PR #285, opened 2024-03-18
refactor: remove expensive clones from verify_consistency_of_operations; convert API to slice

PastaPastaPasta:
Remove unneeded cloning, and use slice

## What was done?
There was never any need for this function to take in a &Vec; we should never do that. Instead take in a slice.

Additionally, we refactored some of the loop logic to use .take instead of a check on each iteration to see if we were at the end. We also combined two duplicate for loops into 1.

Finally, we significantly refactor the return type to avoid a lot of very expensive cloning. Previously, we were doing something on the magnitude of 9 clones (of expensive underlying objects including vectors) per input GroveDbOp. All of these have been removed. We instead return references to these values instead of clones of them. This does mean that the lifetime of the return value is tied to the underlying data the slice refers to; but especially as I cannot even find an instance where we use these return values, this feels like a very fair trade off.

## How Has This Been Tested?
running `cargo t` locally

## Breaking Changes
Well... It seems in actually usage it is none; however, if someone is using these return values currently; this probably could be breaking. Also it may require converting some stuff to a slice. Please let me know if I should mark this PR title with a `!`

## Checklist:

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -552,14 +552,12 @@ impl GroveDbOp {
     }
 
     /// Verify consistency of operations
-    pub fn verify_consistency_of_operations(ops: &Vec<GroveDbOp>) -> GroveDbOpConsistencyResults {
-        let ops_len = ops.len();
+    pub fn verify_consistency_of_operations(ops: &[GroveDbOp]) -> GroveDbOpConsistencyResults {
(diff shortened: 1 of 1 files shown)


kxcd:
Hahaha, the original code was 🙈.

QuantumExplorer:
I don't see a point in this PR. The verification is not used in production.

PR #286, opened 2024-03-26
pref(rocksdb): enable async io for iterators

shumkov:

Writes are not blocking in RocksDB but reads are blocking. There is an asynchronous IO provided for multi-gets and iterators.  That should speed up query operations in Drive which are utilizing iterators. It should increase CPU consumption though.

More details:
- https://rocksdb.org/blog/2022/10/07/asynchronous-io-in-rocksdb.html
- https://docs.rs/rocksdb/latest/rocksdb/struct.ReadOptions.html#method.set_async_io

## What was done?

- [x] Enabled async io for RocksDB iterators
- [ ] Benchmarks to prove that this option speeds up Drive's queries.

## How Has This Been Tested?

With tests.

## Breaking Changes

None

## Checklist:

### DIFF
--- a/storage/src/rocksdb_storage/storage_context/context_immediate.rs
+++ b/storage/src/rocksdb_storage/storage_context/context_immediate.rs
@@ -33,7 +33,7 @@ use grovedb_costs::{
     storage_cost::key_value_cost::KeyValueStorageCost, ChildrenSizesWithIsSumTree, CostResult,
     CostsExt,
 };
-use rocksdb::{ColumnFamily, DBRawIteratorWithThreadMode, WriteBatchWithTransaction};
+use rocksdb::{ColumnFamily, DBRawIteratorWithThreadMode, ReadOptions, WriteBatchWithTransaction};
 
 use super::{make_prefixed_key, PrefixedRocksDbBatch, PrefixedRocksDbRawIterator};
 use crate::{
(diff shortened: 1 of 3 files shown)


fominok:
As far as I understand RocksDB loads more data than possibly needed to reduce the number of disk reads, and those pre-fetches are blocking. If this change allows it to load subsequent data while we're processing current data then I see no harm in it, though PR title is a bit misleading since we introduce no "async io" to GroveDB, just some background processing out of our control.

@QuantumExplorer what do you think?

fominok (approved):

QuantumExplorer:
PR #286 is NOT safe to merge as-is. Three concerns, in order of severity:                                                                                                                                                            
                                                                                                                                                                                                                                       
  1. Platform asymmetry (Critical): set_async_io(true) only works on Linux with io_uring. On macOS it's a silent no-op — no error, no warning. This means you can't test the actual async behavior on developer machines (you're on    
  macOS), creating a gap where production Linux deployments get different iterator behavior than anything tested locally.                                                                                                              
  2. Possible stale reads (High): RocksDB's own source contains comments about async_io potentially serving "broken, stale keys." While snapshot isolation should protect transactional reads, this is undocumented territory. For a   
  cryptographic Merkle tree database where any inconsistency corrupts proofs, "should work" isn't good enough.                                                                                                                         
  3. Cost tracking impact (Medium): GroveDB tracks operation costs (seek_count, storage_loaded_bytes). Async prefetching changes the actual I/O patterns, so cost metrics may no longer reflect reality.

PR #291, opened 2024-04-25
fix!: empty parent paths should only error if we want them to error

QuantumExplorer:
If a tree on an intermediate path on a subquery path was not present we would receive an error. Now instead we have a field to allow this situation or not. If the option is set to false (don't error), then we simply don't error, any query for any item in that path will simply give back a None value.

Also we fixed an issue with terminal paths.

## How Has This Been Tested?
Added a test to verify this works as intended.

## Breaking Changes
Breaking in the sense that method calls have changed.

## Checklist:


ogabrielides (approved):
LGTM, utACK

PR #309, opened 2024-07-06
feat!: proof system v1 (from v0.1)

QuantumExplorer:
The previous proof system was causing issues for proofs with limits in Platform, these issues would cause proofs to cause failures in the system. As it was built by someone no longer with the project we needed to either understand how it worked or rewrite it. Understanding how it worked proved to be phenomenally hard to achieve, so instead a new GroveDB
proof system was devised.

## What was done?
Introduced Proof System V1. This proof system resolves around recursive GroveDB proof layers. Each proof layer represents a merk and can have sublayers if we want to query subtrees.

Proofs are no longer verbose/non verbose, instead you just simply need to set in your proof verification options that you are querying a subsection of the proof.

Proofs also can no longer have offsets. Offsets never really worked with proofs as they can balloon proofs. It was an architectural mistake to every have offsets and this is now fixed. Instead users should use proper ranges that were offset by the end value of previous requests.

## How Has This Been Tested?
Many many tests were written, also it passes all platform tests, and also solves the issues we were experiencing in platform.

## Breaking Changes
Since this is a new proof system, it represents a breaking change to Platform.

## Checklist:


grovedb/src/operations/proof/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+
+use bincode::{Decode, Encode};
+use derive_more::From;
+use grovedb_merk::proofs::{query::Key, Decoder, Node, Op};
+
+use crate::operations::proof::util::{element_hex_to_ascii, hex_to_ascii};
+
+#[derive(Debug, Clone, Copy, Encode, Decode)]
+pub struct ProveOptions {
+    pub decrease_limit_on_empty_sub_query_result: bool,
```
ogabrielides, line 18: Please add a more descriptive comment on the purpose of this field. Since it it passed on the proof APIs

merk/src/proofs/query/verify.rs
```diff
@@ -37,266 +42,401 @@ pub fn verify(bytes: &[u8], expected_hash: MerkHash) -> CostResult<Map, Error> {
...
-    let mut output = Vec::with_capacity(query.len());
-    let mut last_push = None;
-    let mut query = query.directional_iter(left_to_right).peekable();
-    let mut in_range = false;
-    let mut current_limit = limit;
-    let mut current_offset = offset;
+#[derive(Copy, Clone, Debug)]
+pub struct VerifyOptions {
+    pub absence_proofs_for_non_existing_searched_keys: bool,
+    pub verify_proof_succinctness: bool,
```
ogabrielides, line 48: Same for this `verify_proof_succinctness`. (The other two are explicit)

ogabrielides (commented):
LGTM. 
Some documentation requests (can be done later)

grovedb/src/operations/proof/mod.rs
ogabrielides: (Someone has to play the role of @thephez)

merk/src/proofs/query/verify.rs
QuantumExplorer: Done

ogabrielides (approved):
👍

QuantumExplorer:
@fominok please review, even after I merge this in.

PR #334, opened 2024-08-23
feat!: just in time flags fix and epoch based storage flags

QuantumExplorer:
A few issues existed in GroveDB.
These were mainly due to how storage flags would change with just in time upgrades.
We were originally experiencing a quadratic rise in the reporting of storage costs, even though these did not actually increase.
There was also a problem with how we merged storage costs and how we removed them.

## What was done?
Storage Flags was moved into GroveDB and heavily modified.
The just in time logic was heavily modified.
Fixed issue where references would break for change of size of sum items.
Updated to v1.1.0

## How Has This Been Tested?
Added a lot of tests.

## Breaking Changes
This is a breaking change.

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,4 +11,4 @@ members = [
     "path",
     "grovedbg-types",
     "grovedb-version"
-]
+, "grovedb-epoch-based-storage-flags"]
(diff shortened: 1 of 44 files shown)


shumkov (approved):
Not sure I understand what I'm approving

PR #342, opened 2024-09-27
feat: added a convenience method that will return an existing item du…

QuantumExplorer:
…ring a check for insertion

This change implements a new function insert_if_not_exists_return_existing_element in GroveDB. The function allows for the insertion of an element if it does not already exist at the specified path and key. If the element already exists, the function returns the existing element instead of performing an insertion. This is useful for scenarios where we want to avoid overwriting existing data but still need to confirm its presence.

## What was done?
•	Added a new function insert_if_not_exists_return_existing_element to GroveDB.
	•	This function checks if an element already exists at the given path and key.
	•	If the element exists, it returns the existing element.
	•	If the element does not exist, it inserts the new element and returns None.

## Breaking Changes
This change does not introduce any breaking changes.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method to insert elements into the database only if they do not already exist, returning the existing element if found.
	- Added a new field to support this functionality in the versioning structure.

- **Bug Fixes**
	- Updated method signatures for clarity on return types.

### DIFF
--- a/grovedb-version/src/version/grovedb_versions.rs
+++ b/grovedb-version/src/version/grovedb_versions.rs
@@ -135,6 +135,7 @@ pub struct GroveDBOperationsInsertVersions {
     pub add_element_on_transaction: FeatureVersion,
     pub add_element_without_transaction: FeatureVersion,
     pub insert_if_not_exists: FeatureVersion,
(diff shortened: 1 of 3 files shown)


QuantumExplorer:
@coderabbitai review

PR #359, opened 2025-02-23
feat: chunk packing

ogabrielides:
This PR allows faster state sync by packing several chunks into one.

## What was done?
A two tier packing is made. 
One for packing N global_chunk_ids into one, and the second one is for packing local_chunk_ids belonging to the same subtree.

## How Has This Been Tested?
Tutorial replication

## Breaking Changes
No

## Checklist:
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

### DIFF
--- a/grovedb/src/replication.rs
+++ b/grovedb/src/replication.rs
@@ -7,7 +7,11 @@ use grovedb_path::SubtreePath;
 use grovedb_version::{check_grovedb_v0, version::GroveVersion};
 
 pub use self::state_sync_session::MultiStateSyncSession;
-use crate::{util::TxRef, Error, GroveDb, TransactionArg};
+use crate::{
+    replication::utils::{pack_nested_bytes, unpack_nested_bytes},
+    util::TxRef,
+    Error, GroveDb, TransactionArg,
+};
 
 /// Type alias representing a chunk identifier in the state synchronization
 /// process.
@@ -15,8 +19,14 @@ use crate::{util::TxRef, Error, GroveDb, TransactionArg};
 /// - `SubtreePrefix`: The prefix of the subtree (32 bytes).
 /// - `Option<Vec<u8>>`: The root key, which may be `None` if not present.
 /// - `bool`: Indicates whether the tree is a sum tree.
-/// - `Vec<u8>`: The chunk ID representing traversal instructions.
-pub type ChunkIdentifier = (crate::SubtreePrefix, Option<Vec<u8>>, TreeType, Vec<u8>);
+/// - `Vec<Vec<u8>>`: Vector containing the chunk ID representing traversal
+///   instructions.
+pub type ChunkIdentifier = (
+    crate::SubtreePrefix,
+    Option<Vec<u8>>,
+    TreeType,
+    Vec<Vec<u8>>,
+);
 
 pub const CURRENT_STATE_SYNC_VERSION: u16 = 1;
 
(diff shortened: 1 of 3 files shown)


grovedb/src/replication.rs
```diff
@@ -7,16 +7,26 @@ use grovedb_path::SubtreePath;
...
 /// process.
 ///
 /// - `SubtreePrefix`: The prefix of the subtree (32 bytes).
 /// - `Option<Vec<u8>>`: The root key, which may be `None` if not present.
 /// - `bool`: Indicates whether the tree is a sum tree.
-/// - `Vec<u8>`: The chunk ID representing traversal instructions.
-pub type ChunkIdentifier = (crate::SubtreePrefix, Option<Vec<u8>>, TreeType, Vec<u8>);
+/// - `Vec<Vec<u8>>`: Vector containing the chunk ID representing traversal
+///   instructions.
+pub type ChunkIdentifier = (
```
fominok, line 24: seems like it would benefit from being a new type with fields rather than an alias to tuple

at least comments above can be proper docs to fields

grovedb/src/replication.rs
```diff
@@ -84,57 +105,78 @@ impl GroveDb {
...
-                Error::CorruptedData(format!(
-                    "failed to open merk by prefix tx:{} with:{}",
-                    hex::encode(chunk_prefix),
-                    e
-                ))
-            })?;
-        if merk.is_empty_tree().unwrap() {
-            return Ok(vec![]);
+        let mut global_chunk_ids: Vec<Vec<u8>> = vec![];
+        let root_app_hash = self.root_hash(None, grove_version).value?;
```
fominok, line 109: `root_hash` takes `TransactionArg` and we have an ongoing `TxRef` above, to keep them inside one transaction this `None` shall be changed to `Some(tx.as_ref())`

grovedb/src/replication.rs
```diff
@@ -84,57 +105,78 @@ impl GroveDb {
...
+                        e
+                    ))
+                })?;
+                // Ensure we iterate once with vec![] if it's empty
+                let iter: Box<dyn Iterator<Item = Vec<u8>>> = if nested_chunk_ids.is_empty() {
+                    Box::new(std::iter::once(vec![]))
+                } else {
+                    Box::new(nested_chunk_ids.into_iter())
+                };
+                for chunk_id in iter {
```
fominok, line 156: I'm not sure actually what will be better but `nested_chunk_ids.is_empty().then(|| vec![]).into_iter().chain(nested_chunk_ids.into_iter())` in place of `iter` could work instead of dynamic dispatch above

grovedb/src/replication.rs
```diff
@@ -402,4 +462,124 @@ pub(crate) mod utils {
...
+    /// # Arguments
+    ///
+    /// * `nested_bytes` - A vector of byte vectors (`Vec<Vec<u8>>`) to be
+    ///   packed.
+    ///
+    /// # Returns
+    ///
+    /// A `Vec<u8>` containing the encoded representation of the nested byte
+    /// vectors.
+    pub fn pack_nested_bytes(nested_bytes: Vec<Vec<u8>>) -> Vec<u8> {
```
fominok, line 484: vector of vectors is a pretty strict requirement, however, what actually used here is a sequence of slices and could be lifted to something like `impl IntoIter<Item = &[u8]>`

this isn't a dealbreaker for me of course, but if somewhere in the code it requires `.to_vec()` conversions to match this requirement then consider this change

grovedb/src/replication/state_sync_session.rs
```diff
@@ -342,78 +350,134 @@ impl<'db> MultiStateSyncSession<'db> {
...
-            return Err(Error::InternalError(
-                "Unable to process incoming chunk".to_string(),
-            ));
-        };
-        let Ok(res) = subtree_state_sync.apply_inner_chunk(&chunk_id, chunk, grove_version) else {
-            return Err(Error::InternalError("Invalid incoming prefix".to_string()));
-        };
+        let mut next_global_chunk_ids: Vec<Vec<u8>> = vec![];
+
+        for (index, (iter_global_chunk_id, iter_packed_chunks)) in nested_global_chunk_ids
```
fominok, line 376: clippy notices unused `index`, could be a mistake or just `_` it

grovedb/src/replication/state_sync_session.rs
```diff
@@ -342,78 +350,134 @@ impl<'db> MultiStateSyncSession<'db> {
...
+        {
+            let mut next_chunk_ids = vec![];
+
+            let (chunk_prefix, _, _, nested_local_chunk_ids) =
+                replication::utils::decode_global_chunk_id(
+                    iter_global_chunk_id.as_slice(),
+                    &self.app_hash,
+                )?;
+
+            let iter_chunk_ds: Box<dyn Iterator<Item = Vec<u8>>> =
```
fominok, line 389: looks like an overkill and can just return empty vec from true branch and collected vec from false branch

grovedb/src/replication.rs
ogabrielides: good catch!

PR #439, opened 2026-03-01
test(query): add comprehensive test suite for coverage improvement

thepastaclaw:
Improves test coverage for the grovedb-query crate (72.37% baseline). Adds tests for query construction, serialization, path queries, edge cases, and error paths.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for Query API functionality, including construction, iteration, and serialization round-trip validation.
  * Added extensive tests for terminal key extraction and query merging operations across various configurations.
  * Added tests for utility functions and proof item processing logic.

PR #443, opened 2026-03-02
test(query): add test suite for coverage improvement

QuantumExplorer:
## Summary
- Tests for encode/decode round-trip and error paths (unsupported version, oversized branches)
- Tests for conditional subquery branch first-match-wins logic
- Tests for terminal_keys recursion, deduplication, and error paths
- Tests for merge APIs including range intersection splitting
- Tests for ProofItems process_key partitioning

Cleaned up from #439 — removed trivial constructor/accessor/Display tests that only inflate coverage numbers without catching real bugs.

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for Query API serialization, deserialization, and round-trip verification
  * Added test coverage for terminal keys behavior and query merging logic
  * Added test coverage for query item boundary processing and state validation

### DIFF
--- a/grovedb-query/tests/query_api_and_serialization.rs
+++ b/grovedb-query/tests/query_api_and_serialization.rs
@@ -11,44 +11,6 @@ fn k(v: u8) -> Vec<u8> {
 }
 
 #[test]
-fn query_constructors_and_iteration_behave_as_expected() {
-    let q = Query::new();
-    assert!(q.is_empty());
-    assert_eq!(q.len(), 0);
-    assert!(q.left_to_right);
-
-    let q_full = Query::new_range_full();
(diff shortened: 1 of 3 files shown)


thepastaclaw:
Thanks for the suggestion! I considered this, but in Rust integration tests each file under `tests/` compiles as a separate crate. Sharing helpers requires either a `mod common;` pattern (with `#[allow(dead_code)]` annotations since not every file uses every helper) or a `tests/common/mod.rs` structure.

For single-line helpers like `fn k(v: u8) -> Vec<u8> { vec![v] }`, the extraction adds more boilerplate than the duplication itself. If these helpers grow more complex or numerous, consolidation would make sense — but for now the locality is clearer.

PR #480, opened 2026-03-04
fix(merk): add runtime key length validation (RUST-044)

QuantumExplorer:
## Summary
- Replace `debug_assert!` with runtime errors for key length >= 256 bytes in Link and KV encoding methods
- In release builds, `debug_assert!` is stripped, so `key.len() as u8` silently truncates keys >= 256 bytes, causing data corruption on encode and misaligned fields on decode
- Added attack test demonstrating the corruption: a 300-byte key gets truncated to 44 bytes (300 % 256)

## Changes
- `link.rs`: `encoding_cost()`, `encode_into()`, `encoding_length()` — return `ed::Error::IOError` instead of `debug_assert!`
- `kv.rs`: `encoding_cost()` — upgraded to `assert!`; `encoding_length()` — return `ed::Error::IOError`
- Updated existing `encode_link_long_key` test from `#[should_panic]` to verify error return

## Test plan
- [x] `attack_key_length_truncation_corrupts_decode` — demonstrates the data corruption
- [x] `encode_link_long_key_returns_error` — verifies all three methods return errors
- [x] Full merk test suite passes (323 tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for key size limits. The system now returns proper error messages for oversized keys instead of crashing, providing better reliability and error reporting.

### DIFF
--- a/merk/src/tree/kv.rs
+++ b/merk/src/tree/kv.rs
@@ -451,7 +451,7 @@ impl KV {
 
     #[inline]
     fn encoding_cost(&self) -> usize {
-        debug_assert!(self.key().len() < 256, "Key length must be less than 256");
+        assert!(self.key().len() < 256, "Key length must be less than 256");
         HASH_LENGTH_X2 + self.value.len() + self.feature_type.encoding_cost()
     }
 }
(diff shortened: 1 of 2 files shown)


QuantumExplorer:
This actually is not useful

PR #493, opened 2026-03-05
perf: reduce cloning in batch structure construction

QuantumExplorer:
## Summary
- Change `TreeCache::insert` trait to take `(&KeyInfoPath, &KeyInfo, TreeType)` instead of `&QualifiedGroveDbOp`, eliminating the need to reconstruct a full op just to pass a reference
- Destructure `QualifiedGroveDbOp` early in the loop and move owned values into final destinations instead of cloning
- Use entry API on IntMap and BTreeMap to avoid redundant path clones for map insertion

Reduces per-op clones from ~7 (for tree element ops) to 3. Addresses audit finding P3.

## Test plan
- [x] All 267 batch tests pass
- [x] Clippy clean


QuantumExplorer (commented):
Reviewed

PR #498, opened 2026-03-05
fix: return error instead of panicking on missing old storage flags

QuantumExplorer:
## Summary
- **Audit finding E3**: `update_element_flags()` in `grovedb-epoch-based-storage-flags` called `.unwrap()` on an `Option<StorageFlags>`, which would panic at runtime if `maybe_old_storage_flags` was `None`.
- Replaced the `.unwrap()` with `.ok_or(StorageFlagsError::RemovingFlagsError(...))` so that a `None` value produces a proper error return instead of a panic.
- All 41 existing tests continue to pass.

## Test plan
- [x] Verified `cargo build -p grovedb-epoch-based-storage-flags` compiles successfully
- [x] Verified `cargo test -p grovedb-epoch-based-storage-flags` passes all 41 tests

## Release Notes

* **Bug Fixes**
  * Improved error handling for storage state management by adding explicit validation and error messaging when critical storage information is unavailable, making the system more robust.

### DIFF
--- a/grovedb-epoch-based-storage-flags/src/update_element_flags.rs
+++ b/grovedb-epoch-based-storage-flags/src/update_element_flags.rs
@@ -28,8 +28,13 @@ impl StorageFlags {
             .ok_or(StorageFlagsError::RemovingFlagsError(
                 "removing flags from an item with flags is not allowed".to_string(),
             ))?;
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Reviewed

PR #501, opened 2026-03-05
fix: propagate error instead of silently discarding in Restorer::process_chunk

QuantumExplorer:
## Summary

- **Audit finding E6**: In `Restorer::process_chunk()`, the call to `self.merk.set_base_root_key(...)` returned a `CostResult<(), Error>` that was silently discarded via `let _ =`. If `set_base_root_key` fails (e.g., due to a storage error), the restorer would continue operating on corrupt state without any indication of failure.
- This fix extracts the inner `Result` via `.value` and propagates errors with `?`, matching the error-handling pattern used elsewhere in the same method.

## Test plan

- [x] Verified the crate compiles successfully with `cargo build -p grovedb-merk`
- [ ] Existing restore/chunk tests continue to pass (`cargo test -p grovedb-merk`)


thepastaclaw:
@coderabbitai review

PR #505, opened 2026-03-05
fix: use u32 for element count in pack/unpack_nested_bytes to prevent truncation

QuantumExplorer:
## Summary

- **Security finding S1**: `pack_nested_bytes` encoded the element count as `u16`, silently truncating if there were more than 65,535 elements. Changed to `u32` to support up to ~4 billion elements.
- Updated `unpack_nested_bytes` to read a 4-byte `u32` element count instead of 2-byte `u16`, and improved the initial bounds check from `is_empty()` to `len() < 4` (also fixes related finding L1 about a potential panic on 1-byte input).
- Updated doc comments on both functions to reflect the new encoding format.
- Updated tests that hard-coded the old `u16` header format.

## Breaking Change

This is a **breaking change to the wire format** for replication data. Packed data produced by the new code is not compatible with the old decoder, and vice versa. Any peers communicating via `pack_nested_bytes` / `unpack_nested_bytes` must be upgraded together.

## Test plan

- [x] `cargo test -p grovedb --lib -- replication` — all 39 tests pass

## Release Notes

* **Bug Fixes**
  * Enhanced data validation and error detection for replication data handling with more comprehensive boundary checks and error reporting for corrupted or incomplete data.

### DIFF
--- a/grovedb/src/replication.rs
+++ b/grovedb/src/replication.rs
@@ -476,8 +476,9 @@ pub(crate) mod utils {
     /// vector.
     ///
     /// The encoding format is as follows:
-    /// 1. The first byte stores the number of elements.
-    /// 2. Each element is prefixed with its length as a 2-byte (`u16`) value in
+    /// 1. The first four bytes store the number of elements as a `u32` in
+    ///    big-endian format.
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Approved

PR #506, opened 2026-03-05
fix: validate key length on insert to prevent encoding corruption

QuantumExplorer:
## Summary

- **Security fix (S2):** The Merk link encoding stores key length as a single `u8` byte (`merk/src/tree/link.rs` line 354: `out.write_all(&[key.len() as u8])?`). Keys longer than 255 bytes have their length silently truncated via `as u8` cast, corrupting the link encoding and breaking proof generation/verification. The only existing protection is a `debug_assert!(key.len() < 256)` which is stripped in release builds.
- Adds explicit key length validation (`key.len() > 255`) to all four public insert entry points in `grovedb/src/operations/insert/mod.rs`: `insert`, `insert_if_not_exists`, `insert_if_not_exists_return_existing_element`, and `insert_if_changed_value`. Returns `Error::InvalidInput` before any work is done if the key exceeds 255 bytes.

## Test plan

- [x] All 195 existing insert-related tests pass (`cargo test -p grovedb -- insert`)
- [ ] Verify that inserting a key with >255 bytes now returns `Error::InvalidInput` instead of silently corrupting data

## Summary by CodeRabbit

* **Bug Fixes**
  * Insert operations now enforce a maximum key length of 255 bytes, returning an error if exceeded. This prevents invalid keys from being processed.

### DIFF
--- a/grovedb/src/operations/insert/mod.rs
+++ b/grovedb/src/operations/insert/mod.rs
@@ -72,6 +72,11 @@ impl GroveDb {
             grove_version.grovedb_versions.operations.insert.insert
         );
 
+        if key.len() > 255 {
+            return Err(Error::InvalidInput("key length must be at most 255 bytes"))
+                .wrap_with_cost(OperationCost::default());
+        }
+
         let subtree_path: SubtreePath<B> = path.into();
         let batch = StorageBatch::new();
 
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Reviewed

PR #507, opened 2026-03-05
fix: harden safety invariants in MultiStateSyncSession

QuantumExplorer:
## Summary

Addresses security finding **S3**: unsafe lifetime extensions in `MultiStateSyncSession` have several concrete issues that could cause panics or mask errors in production.

- **Replace `.unwrap()` with proper error propagation** in `commit()` and `set_new_transaction()` — both call `commit_transaction` which can fail, and both methods already return `Result`, so the unwraps were unnecessary and dangerous
- **Add `debug_assert!`** in `set_new_transaction()` to enforce the documented safety invariant that `current_prefixes` must be empty before replacing the transaction
- **Improve SAFETY comments** on the four pin-projection helper methods to more precisely describe why the `unsafe` access is sound (single-field access, no struct movement)

## Test plan

- [x] `cargo check -p grovedb` passes
- [ ] Existing replication/state sync tests continue to pass
- [ ] No behavioral change in the success path — only error handling on failure paths is improved

### DIFF
--- a/grovedb/src/replication/state_sync_session.rs
+++ b/grovedb/src/replication/state_sync_session.rs
@@ -227,16 +227,29 @@ impl<'db> MultiStateSyncSession<'db> {
         // SAFETY: the struct isn't used anymore and no storage contexts would access
         // transaction
         let session = unsafe { Pin::into_inner_unchecked(self) };
-        session.db.commit_transaction(session.transaction).unwrap()
+        session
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Reviewed

thepastaclaw:
@coderabbitai review

PR #509, opened 2026-03-05
fix: use absolute value in sum limit budget consumption

QuantumExplorer:
## Summary

- **Finding L4**: `saturating_sub(value)` on negative SumItems increases the budget instead of decreasing it
- In `basic_aggregate_sum_push`, `sum_limit_left.saturating_sub(value)` adds to the remaining budget when `value` is negative, because subtracting a negative number is addition. This could allow bypassing sum limits by exploiting negative SumItems to artificially inflate the remaining budget.
- Fix: use `value.saturating_abs()` so that negative sum items consume budget proportional to their magnitude, preventing budget bypass.
- Updated the `test_negative_sum_values` test to reflect the correct behavior: negative items should consume budget (not increase it), and added an additional test case verifying that a budget of 12 correctly stops after processing items with absolute values totaling 13 (10 + 3).

## Test plan

- [x] `cargo test -p grovedb aggregate_sum` — all 51 tests pass
- [x] `cargo clippy -p grovedb -- -D warnings` — no warnings

### DIFF
--- a/grovedb/src/element/aggregate_sum_query/mod.rs
+++ b/grovedb/src/element/aggregate_sum_query/mod.rs
@@ -778,7 +778,7 @@ impl ElementAggregateSumQueryExtensions for Element {
             *limit = limit.saturating_sub(1);
         }
 
-        *sum_limit_left = sum_limit_left.saturating_sub(value);
+        *sum_limit_left = sum_limit_left.saturating_sub(value.saturating_abs());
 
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
No this is wrong, we shouldn't be using absolute values, a +7 and a -4 should give +3, not 11

PR #510, opened 2026-03-05
fix: use u16 for root key length in global chunk ID encoding

QuantumExplorer:
## Summary

- **Finding L5**: The `encode_global_chunk_id` function in `grovedb/src/replication.rs` stored the root key length as a single `u8`, which silently truncates keys longer than 255 bytes. This corrupts replication data for any subtree with a root key exceeding 255 bytes, as the encoded length wraps around and the decoder reads the wrong number of bytes.
- Changed both `encode_global_chunk_id` and `decode_global_chunk_id` to use 2 bytes (`u16` big-endian) for the root key length field, supporting keys up to 65,535 bytes.
- Updated corresponding unit tests that hard-coded single-byte root key size values.

## Test plan

- [x] All 30 existing `replication_utils_tests` pass with the updated encoding/decoding
- [ ] Verify round-trip encode/decode works for root keys > 255 bytes (not currently tested in the suite, but the fix is straightforward)

### DIFF
--- a/grovedb/src/replication.rs
+++ b/grovedb/src/replication.rs
@@ -349,19 +349,21 @@ pub(crate) mod utils {
 
         let (chunk_prefix_key, remaining) = global_chunk_id.split_at(chunk_prefix_length);
 
-        let root_key_size_length: usize = 1;
+        let root_key_size_length: usize = 2;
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Reviewed

PR #511, opened 2026-03-05
fix: remove dead code guards in duplicate detection loop

QuantumExplorer:
## Summary

- **Finding L6**: The `verify_consistency_of_operations` function in `grovedb/src/batch/mod.rs` contained two `if i == ops_len { continue; }` guards in its duplicate detection loops. Since `ops.iter().enumerate()` yields indices `0..ops_len-1`, the condition `i == ops_len` is never true, making these guards dead code.
- The loops already work correctly without the guards: when `i` is the last index, `split_at(i+1).1` returns an empty slice, so the inner filter finds nothing.
- Also removes the now-unused `ops_len` variable.

## Test plan

- [x] `cargo build -p grovedb` compiles cleanly
- [x] All 295 batch-related tests pass (`cargo test -p grovedb batch`)

## Summary by CodeRabbit

* **Refactor**
  * Optimized batch operation processing by removing redundant boundary checks. No user-facing changes.

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -806,13 +806,9 @@ impl QualifiedGroveDbOp {
             .map(|op| (op.clone(), 1))
             .collect();
 
-        let ops_len = ops.len();
         // operations should not have any duplicates
         let mut repeated_ops = internal_only_ops;
         for (i, op) in ops.iter().enumerate() {
-            if i == ops_len {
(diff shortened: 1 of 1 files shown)


QuantumExplorer:
Closing: the dead `i == ops_len` guards were already removed as part of PR #516 (P1 — HashMap-based duplicate detection), which replaced the entire O(n²) loop.

PR #512, opened 2026-03-05
fix: correct version check function name in prove_query

QuantumExplorer:
## Summary

- **Finding A1**: The `check_grovedb_v0_with_cost!` macro call in `prove_query` (line 76 of `grovedb/src/operations/proof/generate.rs`) incorrectly used the string `"prove_query_many"` as the function name identifier. This is a copy-paste error from the neighboring `prove_query_many` method. The correct name is `"prove_query"`.

## Test plan

- [x] Verified the fix is limited to a single string literal change
- [x] Confirmed the neighboring `prove_query_many` method retains its correct label
- [x] Pre-commit hooks pass (fmt, clippy, typos)


QuantumExplorer (commented):
Reviewed

PR #514, opened 2026-03-05
fix: change clear_subtree to return CostResult for API consistency

QuantumExplorer:
## Summary

- **Finding A3**: `clear_subtree` was the only mutation method returning `Result<bool, Error>` instead of `CostResult<bool, Error>`, discarding accumulated operation costs by calling `.unwrap()` on the inner `CostResult`.
- Changed `clear_subtree` to return `CostResult<bool, Error>` and delegate directly to `clear_subtree_with_costs`, matching the return type convention used by all other mutation operations in GroveDB.
- Updated all callers (tests in `delete/mod.rs` and `operations_coverage_tests.rs`) to call `.unwrap()` on the `CostResult` before handling the inner `Result`.

## Test plan

- [x] All existing `clear_subtree` tests pass (`test_subtree_clear`, `clear_subtree_with_items`, `clear_subtree_with_subtrees_not_allowed_error`, `clear_subtree_with_subtrees_not_allowed_returns_false`, `clear_subtree_allowing_subtree_deletion`)
- [x] Full crate build succeeds with no warnings

### DIFF
--- a/grovedb/src/operations/delete/mod.rs
+++ b/grovedb/src/operations/delete/mod.rs
@@ -164,13 +164,12 @@ impl GroveDb {
         options: Option<ClearOptions>,
         transaction: TransactionArg,
         grove_version: &GroveVersion,
-    ) -> Result<bool, Error>
+    ) -> CostResult<bool, Error>
     where
         B: AsRef<[u8]> + 'b,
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
No clear_subtree is made to do it without costs.

PR #515, opened 2026-03-05
docs: mark internal-only GroveOp variants to prevent misuse

QuantumExplorer:
## Summary

**Audit finding A4**: The `GroveOp` enum is fully `pub` but includes four internal-only variants (`ReplaceTreeRootKey`, `InsertTreeWithRootHash`, `ReplaceNonMerkTreeRoot`, `InsertNonMerkTree`) that users should never construct directly. These variants are produced by batch propagation/preprocessing and constructing them manually would create invalid operations.

Since Rust does not allow individual enum variants to have restricted visibility (`pub(crate)`), this PR adds clear `**Internal only — do not construct directly.**` doc-comment warnings on each of the four internal variants and updates the enum-level documentation to explicitly list which variants are user-facing.

- Added enum-level doc comment listing user-facing vs internal variants
- Added `**Internal only — do not construct directly.**` prefix to `ReplaceTreeRootKey`, `InsertTreeWithRootHash`, `ReplaceNonMerkTreeRoot`, and `InsertNonMerkTree` doc comments

## Test plan

- [x] Documentation-only change; no behavioral changes
- [x] Verified `cargo build` succeeds (pre-commit hooks pass)
- [x] Confirmed all four internal variants are annotated

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -165,9 +165,15 @@ impl NonMerkTreeMeta {
     }
 }
 
-/// Operations
+/// Operations for batch processing.
+///
+/// User-facing variants: `InsertOnly`, `InsertOrReplace`, `Replace`, `Patch`,
+/// `RefreshReference`, `Delete`, `DeleteTree`, `CommitmentTreeInsert`,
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Reviewed

PR #517, opened 2026-03-05
perf: use range deletion for wipe instead of per-key iteration

QuantumExplorer:
## Summary

- Replaces O(n) per-key delete loop in `wipe_column_family` with a single `delete_range_cf` call
- Creates one range tombstone per column family instead of issuing individual delete operations for every key
- Handles empty CFs gracefully (early return)

The old code had a comment: "very inefficient way of doing this, time complexity is O(n) — we can do O(1)".

Audit finding P6.

## Test plan

- [x] `cargo test -p grovedb-storage` — all 5 tests pass
- [x] `cargo test -p grovedb wipe` — both wipe tests pass
- [x] `cargo build` clean

## Summary by CodeRabbit

* **Refactor**
  * Improved the wipe operation to iterate over storage partitions and perform efficient range deletions, speeding up large dataset clearing.

* **Tests**
  * Added a test that populates many keys across partitions, verifies wipe removes them all, and confirms normal database operations after wipe.

### DIFF
--- a/storage/src/rocksdb_storage/storage.rs
+++ b/storage/src/rocksdb_storage/storage.rs
@@ -468,16 +468,20 @@ impl RocksDbStorage {
             return Ok(()); // CF is already empty
         };
         iter.seek_to_last();
-        let last_key = iter
-            .key()
-            .expect("CF has at least one key since first_key exists");
+        let Some(last_key) = iter.key() else {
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Reviewed

PR #519, opened 2026-03-05
fix: implement proof succinctness and completeness verification

QuantumExplorer:
## Summary

Implements two proof verification checks that were previously missing:

**Completeness (soundness fix):** Both v0 and v1 `verify_layer_proof` now reject proofs that omit lower layers for non-empty trees the query traverses into. Previously, the else branch silently skipped these cases, allowing a malicious prover to strip subtree proofs while the verifier accepted with fewer results and the same root hash.

**Succinctness (wires up existing flag):** When `verify_proof_succinctness` is true, verification now rejects proofs containing lower layers for keys not required by the query. This implements the behavior that `verify_query` vs `verify_subset_query` was always *intended* to distinguish:
- `verify_query` (succinctness=true) → rejects extra proof data
- `verify_subset_query` (succinctness=false) → tolerates extra proof data

### Changes
- **`grovedb-element/src/element/helpers.rs`** — Added `Element::is_non_empty_tree()` helper for the completeness check
- **`grovedb/src/operations/proof/verify.rs`** — Completeness error in both v0 and v1 verify paths when a non-empty tree has a subquery but no lower layer; succinctness check after result set processing rejects `lower_layers` keys not in `verified_keys`
- **`grovedb/src/tests/succinctness_gap_test.rs`** — TDD tests for both behaviors
- **`grovedb/src/tests/sum_tree_tests.rs`** — Changed `test_verify_query_get_parent_tree_info` to use subset verification (its hardcoded proof has extra lower layers)
- **`grovedb/src/operations/proof/mod.rs`** — Updated doc comments on `verify()` and `verify_with_absence_proof()` (no longer "not yet implemented")
- **`merk/src/proofs/query/verify.rs`** — Updated `verify_proof_succinctness` doc comment
- **`grovedb-version/src/version/v3.rs`** — Fixed `protocol_version` to 3 (merge artifact)

Audit finding H3.

## Test plan

- [x] `cargo test -p grovedb --lib --features full` — 1270 passed
- [x] `cargo test -p grovedb-version` — 47 passed
- [x] `cargo clippy -p grovedb -p grovedb-merk --features full -- -D warnings` — clean
- [x] `test_succinctness_rejects_extra_proof_data` — broad proof rejected by `verify_query` with narrow query, accepted by `verify_subset_query`
- [x] `test_missing_lower_layer_for_non_empty_tree_is_rejected` — tampered proof (stripped lower layer) rejected by both verify methods

### DIFF
--- a/grovedb-element/src/element/helpers.rs
+++ b/grovedb-element/src/element/helpers.rs
@@ -212,6 +212,29 @@ impl Element {
         }
     }
 
+    /// Check if the element is a non-empty tree that should have child data.
+    ///
+    /// For merk-backed trees this means the root key is `Some(_)` (at least one
+    /// node). Non-merk tree types (MmrTree, BulkAppendTree, CommitmentTree,
+    /// DenseAppendOnlyFixedSizeTree) always carry their own data and are
+    /// considered non-empty regardless of their count field.
+    pub fn is_non_empty_tree(&self) -> bool {
+        matches!(
+            self,
+            Element::Tree(Some(_), _)
+                | Element::SumTree(Some(_), ..)
+                | Element::BigSumTree(Some(_), ..)
+                | Element::CountTree(Some(_), ..)
+                | Element::CountSumTree(Some(_), ..)
+                | Element::ProvableCountTree(Some(_), ..)
+                | Element::ProvableCountSumTree(Some(_), ..)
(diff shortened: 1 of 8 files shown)


QuantumExplorer (commented):
Self Reviewed

PR #520, opened 2026-03-05
fix: enforce key length limit in proof encoding at runtime

QuantumExplorer:
## Summary
- **Security fix (audit finding H2)**: Proof encoding in `grovedb-query/src/proofs/encoding.rs` stored key lengths as `u8` but only checked `key.len() < 256` via `debug_assert!`, which is stripped in release builds. Keys >= 256 bytes would have their length silently truncated modulo 256, corrupting proofs and potentially enabling proof forgery.
- Added a `validate_key_len()` helper function that performs the check at runtime, returning `ed::Error::IOError` with `InvalidData` kind for keys >= 256 bytes. All 16 encoding paths that cast `key.len() as u8` now call this validation before proceeding.
- Existing `debug_assert!` lines are retained as belt-and-suspenders for debug builds.

## Test plan
- [x] `cargo build -p grovedb-query` compiles successfully
- [ ] Verify existing tests pass: `cargo test -p grovedb-query`
- [ ] Confirm a key of length 256+ returns an error instead of silently truncating

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced proof encoding validation by implementing comprehensive runtime checks across all encoding paths to ensure key lengths do not exceed 255 bytes. Keys exceeding this limit now trigger detailed error messages during the encoding process, improving error visibility and system reliability.

### DIFF
--- a/grovedb-query/src/proofs/encoding.rs
+++ b/grovedb-query/src/proofs/encoding.rs
@@ -15,6 +15,22 @@ use crate::{
 /// allocations.
 const MAX_VALUE_LEN: u32 = 64 * 1024 * 1024;
 
+/// Validates that a key length fits in a single byte (< 256).
+///
+/// Proof encoding stores key lengths as a u8. Keys with 256 or more bytes
+/// would have their length silently truncated via `as u8`, corrupting the
+/// proof. This function enforces the limit at runtime.
+#[inline]
+fn validate_key_len(key: &[u8]) -> ed::Result<()> {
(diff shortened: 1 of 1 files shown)


QuantumExplorer:
Closing: keys ≥256 bytes cannot reach proof encoding because the insert path (direct + batch via #521) enforces the 255-byte limit. The existing `debug_assert!` is sufficient as a development-time guard.

PR #521, opened 2026-03-05
fix: validate key length in batch operations to prevent Merk encoding corruption

QuantumExplorer:
## Summary

- **Audit finding H1**: The batch operation path (`apply_batch`) skipped key length validation, allowing keys >255 bytes to be inserted. Since Merk link encoding stores key length as a single `u8`, this would corrupt the encoding.
- Adds key length validation in `BatchStructure::continue_from_ops` (the central batch structure builder), returning `Error::InvalidInput` when any keyed operation has a key exceeding 255 bytes. Keyless operations (append-only tree ops) are not affected.
- Adds a test that verifies batch operations with 256-byte keys are rejected and 255-byte keys are accepted.

## Details

The direct insert path already validates key length via `validate_key_length()` in `grovedb/src/operations/insert/mod.rs`. The batch path builds operations through `BatchStructure::continue_from_ops` in `grovedb/src/batch/batch_structure.rs`, which previously had no such check. The fix adds the same validation at the point where each operation's key is extracted, before it enters the batch structure.

## Test plan

- [x] `cargo build` succeeds
- [x] New test `test_batch_rejects_key_longer_than_255_bytes` passes
- [x] All 296 existing batch tests pass (`cargo test -p grovedb batch`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Batch processing now validates key length, rejecting keys exceeding 255 bytes with an InvalidInput error indicating the maximum permitted length.

* **Tests**
  * Added test coverage verifying batch processing correctly rejects keys longer than 255 bytes while accepting valid keys up to the limit.

### DIFF
--- a/grovedb/src/batch/batch_structure.rs
+++ b/grovedb/src/batch/batch_structure.rs
@@ -133,11 +133,11 @@ where
             // Validate key length: Merk link encoding stores key length as a
             // single u8, so keys longer than 255 bytes would corrupt the
             // encoding.
-            if let KeyInfo::KnownKey(ref key_bytes) = key {
-                if key_bytes.len() > u8::MAX as usize {
-                    return Err(Error::InvalidInput("key length must be at most 255 bytes"))
-                        .wrap_with_cost(cost);
-                }
+            if let KeyInfo::KnownKey(ref key_bytes) = key
+                && key_bytes.len() > u8::MAX as usize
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Reviewed and tested, this is indeed a fix

PR #522, opened 2026-03-05
fix: prevent u16 limit underflow in proof verification

QuantumExplorer:
## Summary

- Replace all 6 instances of `*limit -= 1` with `*limit = limit.saturating_sub(1)` in `grovedb/src/operations/proof/verify.rs` to prevent `u16` underflow when the proof verification limit reaches 0
- When `decrease_limit_on_empty_sub_query_result` is `true` and multiple empty subtrees are encountered, the limit could be decremented past 0 — wrapping to 65535 in release mode (accepting oversized result sets) or panicking in debug mode
- Add regression tests for both v0 and v1 proof paths that exercise the edge case where `limit=1` and multiple empty subtrees are queried

## Context

Audit finding H4: when the merk proof returns an empty result set and `decrease_limit_on_empty_sub_query_result` is enabled, the code decrements `limit_left` without checking if it is already 0. This is a security issue because in release builds the `u16` wraps to 65535, causing the verifier to accept proofs with far more results than the query limit intended.

## Test plan

- [x] `cargo build` succeeds
- [x] `cargo test -p grovedb verify` — all 68 verify tests pass
- [x] `cargo test -p grovedb proof_coverage` — all 114 proof coverage tests pass
- [x] New test `verify_no_underflow_when_limit_zero_and_empty_subtree` exercises v0 path
- [x] New test `verify_v1_no_underflow_when_limit_zero_and_empty_subtree` exercises v1 path

### DIFF
--- a/grovedb/src/operations/proof/verify.rs
+++ b/grovedb/src/operations/proof/verify.rs
@@ -448,7 +448,9 @@ impl GroveDb {
 
         if merk_result.result_set.is_empty() {
             if prove_options.decrease_limit_on_empty_sub_query_result {
-                limit_left.iter_mut().for_each(|limit| *limit -= 1);
+                limit_left
+                    .iter_mut()
+                    .for_each(|limit| *limit = limit.saturating_sub(1));
             }
         } else {
             for proved_key_value in merk_result.result_set {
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Approved

PR #523, opened 2026-03-06
fix(batch): enforce internal-only GroveOp variants with #[non_exhaustive]

thepastaclaw:
## Issue

Audit finding A4: The `GroveOp` enum has four internal-only variants (`ReplaceTreeRootKey`, `InsertTreeWithRootHash`, `ReplaceNonMerkTreeRoot`, `InsertNonMerkTree`) that should not be constructed by external crates. PR #515 added documentation, but there was no compile-time enforcement.

## Fix

Added `#[non_exhaustive]` to each internal-only variant. In Rust, `#[non_exhaustive]` on a struct-like enum variant **prevents external crates from constructing it** while allowing internal (same-crate) code to work normally. External code can still pattern-match with `..`, but cannot create instances.

This provides compile-time enforcement on top of the existing runtime rejection in `batch_structure.rs`, at zero runtime cost.

## Why not split the enum?

- Minimal change (doc comments + 4 attributes vs. major refactor)
- No breaking changes to internal code or tests
- Idiomatic Rust pattern for this exact use case

## Validation

- `cargo build` ✅
- `cargo clippy --all-targets` ✅ (1 pre-existing warning, unrelated)
- `cargo test -p grovedb -- batch_rejection` ✅ (6 tests pass)
- Pre-commit hooks (fmt + check) ✅

## Summary by CodeRabbit

* **Breaking Changes**
  * Four core operation variants are now marked non-exhaustive, restricting external code from directly pattern matching or constructing these specific variants. All underlying behavior remains unchanged.

* **Documentation**
  * Documentation expanded with additional clarifications regarding the internal nature of affected operation variants and their construction restrictions.

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -170,7 +170,12 @@ impl NonMerkTreeMeta {
 /// User-facing variants: `InsertOnly`, `InsertOrReplace`, `Replace`, `Patch`,
 /// `RefreshReference`, `Delete`, `DeleteTree`, `CommitmentTreeInsert`,
 /// `MmrTreeAppend`, `BulkAppend`, `DenseTreeInsert`.
-/// Other variants are internal and produced by batch propagation.
+///
+/// Internal variants (`ReplaceTreeRootKey`, `InsertTreeWithRootHash`,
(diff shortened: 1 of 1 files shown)


thepastaclaw:
@coderabbitai review

QuantumExplorer (approved):

PR #524, opened 2026-03-06
test: cover empty tree early-return path in Merk::verify()

thepastaclaw:
## Summary

Adds a test covering the early-return path in `Merk::verify()` when called on an empty tree (no loaded root). This path was added in #497 to return empty maps instead of panicking, and was flagged as uncovered by codecov.

## Changes

- `merk/src/merk/mod.rs`: Added `test_verify_empty_tree_returns_empty_maps` test (+18 lines)

## Validation

- `cargo test` passes
- `cargo clippy` clean
- Pre-commit hooks pass


thepastaclaw:
@coderabbitai review

QuantumExplorer (approved):

PR #531, opened 2026-03-06
fix: prevent u32 overflow in epoch storage flags (audit M2)

QuantumExplorer:
## Summary

- **Audit finding M2**: `combine_with_higher_base_epoch` at `grovedb-epoch-based-storage-flags/src/lib.rs` performed unchecked `u32` addition (`original_bytes + added_bytes`) which could overflow/wrap when summing epoch byte counts.
- Replaced with `checked_add`, returning `StorageFlagsError::StorageFlagsOverflow` on overflow. This follows the same pattern already used in `combine_with_higher_base_epoch_remove_bytes` (which uses `checked_sub` with the same error variant).
- Added a regression test that constructs a `MultiEpoch` with a value near `u32::MAX` and verifies that adding bytes past the limit returns an error instead of panicking or wrapping.

## Test plan

- [x] `cargo build -p grovedb-epoch-based-storage-flags` passes
- [x] `cargo test -p grovedb-epoch-based-storage-flags` passes (42 tests including the new one)
- [x] `cargo clippy -p grovedb-epoch-based-storage-flags -- -D warnings` passes with no warnings

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced storage operations to detect and properly handle overflow conditions, preventing potential data corruption.

* **Tests**
  * Added test coverage for overflow edge cases.

### DIFF
--- a/grovedb-epoch-based-storage-flags/src/lib.rs
+++ b/grovedb-epoch-based-storage-flags/src/lib.rs
@@ -209,7 +209,12 @@ impl StorageFlags {
         match original_value {
             None => other_epoch_bytes.insert(*epoch_with_adding_bytes, added_bytes),
             Some(original_bytes) => {
-                other_epoch_bytes.insert(*epoch_with_adding_bytes, original_bytes + added_bytes)
+                let combined = original_bytes.checked_add(added_bytes).ok_or(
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Reviewed

PR #533, opened 2026-03-06
fix: replace expect with error return in range query iterator (audit M3)

QuantumExplorer:
## Summary

Replace `.expect()` calls on iterator `value()` and `key()` in `Element::query` range handling with proper `Error::CorruptedData` returns. If storage is corrupted, the iterator could return `None` even though `iter_is_valid_for_type` reports valid, causing a panic.

Audit finding M3.

## Test plan

- [x] `cargo build --features full` — clean
- [x] `cargo test -p grovedb --lib --features full -- query_tests` — 65 passed


QuantumExplorer (commented):
Reviewed

PR #534, opened 2026-03-06
fix: replace silent u8 truncation in build_prefix_body

QuantumExplorer:
## Summary
- `build_prefix_body` cast path segment lengths to `u8` via `as u8`, silently truncating segments > 255 bytes (e.g., 256 → 0), which could cause different paths to hash to the same prefix (collision)
- Replaced with `u8::try_from().expect()` so oversized segments produce a clear panic instead of silent data corruption
- In practice, this invariant is already enforced at insert time via `validate_key_length` (keys ≤ 255 bytes), so this is defense-in-depth

## Test plan
- [x] Added `test_build_prefix_rejects_oversized_segment` — verifies 256-byte segment panics with clear message
- [x] All existing storage tests pass (`cargo test -p grovedb-storage --features rocksdb_storage`)

Fixes audit finding M4.

### DIFF
--- a/storage/src/rocksdb_storage/storage.rs
+++ b/storage/src/rocksdb_storage/storage.rs
@@ -162,7 +162,10 @@ impl RocksDbStorage {
         for s in segments_iter {
             segments_count += 1;
             res.extend_from_slice(s);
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
approved

PR #536, opened 2026-03-06
fix: document Owner::own_result poisoning and improve panic message

QuantumExplorer:
## Summary
- When `own_result`'s closure returns `Err`, the contained value has been consumed and the Owner is left with `inner = None` (poisoned state)
- Previously, any subsequent access would hit `unreachable!("value should be Some")` — unhelpful for debugging
- Now: descriptive panic message explaining what happened and that the Owner must not be used after `own_result` error
- Added doc warning on `own_result` explaining the poisoning behavior
- Added unit tests for both success and error paths

## Test plan
- [x] `test_own_result_success_preserves_value` — verifies normal operation
- [x] `test_own_result_error_poisons_owner` — verifies the improved panic message
- [x] All merk tests pass (`cargo test -p grovedb-merk`)

Fixes audit finding M8.

### DIFF
--- a/merk/src/owner.rs
+++ b/merk/src/owner.rs
@@ -67,15 +67,18 @@ impl<T> Owner<T> {
     /// The function must return a result of the same type (the same value, or a
     /// new value to take its place).
     ///
-    /// Like `own`, but uses a tuple return type which allows specifying a value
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Reviewed

PR #537, opened 2026-03-06
fix: replace panics with error returns in TreeNode::load

QuantumExplorer:
## Summary
- `TreeNode::load()` panicked via `.expect("Expected link")` and `panic!("Expected Some(Link::Reference)")` for conditions that should return errors
- Both are now proper `Err(Error::CorruptedState(...))` returns, matching the function's existing `CostResult<(), Error>` return type
- Removes the TODO comment that requested this exact change

## Test plan
- [x] All 330 merk unit tests pass (`cargo test -p grovedb-merk --lib`)
- [x] `cargo build -p grovedb-merk` compiles cleanly

Fixes audit finding M9.

### DIFF
--- a/merk/src/tree/mod.rs
+++ b/merk/src/tree/mod.rs
@@ -1214,8 +1214,12 @@ impl TreeNode {
     where
         V: Fn(&[u8], &GroveVersion) -> Option<ValueDefinedCostType>,
     {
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Approved

PR #538, opened 2026-03-06
docs: document why all-0xFF prefix overflow in seek_to_last is not reachable

QuantumExplorer:
## Summary
- The prefix increment loop in `seek_to_last` would wrap to all zeros if every byte were `0xFF`, which would seek to the wrong position
- However, prefixes are 32-byte Blake3 hashes — an all-`0xFF` output has probability 1/2^256, which is effectively impossible
- Added doc comments explaining this rather than adding unreachable defensive code

## Test plan
- [x] All 25 storage tests pass (`cargo test -p grovedb-storage --features rocksdb_storage`)

Resolves audit finding L4 (not a bug — documented as unreachable).


QuantumExplorer (commented):
Approved

PR #539, opened 2026-03-06
fix: consistency check now catches InsertOnly and DeleteTree

QuantumExplorer:
## Summary
- `verify_consistency_of_operations` had two gaps in its "no inserts under deleted paths" check:
  1. Only `GroveOp::Delete` was considered when building deleted paths — `DeleteTree` (which removes everything under a path) was missed
  2. Only `InsertOrReplace` and `Replace` were checked as conflicting inserts — `InsertOnly` and `Patch` were missed
- Both are now included in the respective match arms

## Test plan
- [x] `test_consistency_insert_only_under_deleted_path` — verifies `InsertOnly` under a `Delete` path is flagged
- [x] `test_consistency_insert_under_delete_tree_path` — verifies `InsertOrReplace` under a `DeleteTree` path is flagged
- [x] All 4 consistency tests pass

Note: L16 from the audit ("`count > 1` instead of `count > 0`") is a false positive — the counter starts at 0 and increments per occurrence, so `count > 1` correctly catches ops appearing 2+ times.

Fixes audit finding L17.

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -902,7 +902,7 @@ impl QualifiedGroveDbOp {
         // Build a map of deleted_qualified_path -> indices of delete ops
         let mut deleted_path_to_op_indices: HashMap<KeyInfoPath, Vec<usize>> = HashMap::new();
         for (idx, op) in ops.iter().enumerate() {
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Reviewed

PR #542, opened 2026-03-06
fix: return error instead of panicking on None tree in rewrite_heights (M11)

QuantumExplorer:
## Summary
- Replace `.unwrap()` on `self.merk.tree.take()` in `Restorer::rewrite_heights` with `.ok_or()` error return
- Prevents panic if tree is unexpectedly None during chunk restoration

## Test plan
- [x] `cargo test -p grovedb-merk --lib -- restore` — 13 passed


QuantumExplorer (commented):
Approved

PR #544, opened 2026-03-06
docs: document why to_ne_bytes is intentional in build_prefix_body

QuantumExplorer:
## Summary

- Documents why `to_ne_bytes()` is used intentionally in `build_prefix_body` — changing to `to_be_bytes()` would be a breaking change since the output is Blake3-hashed into subtree prefixes stored in RocksDB
- Adds inline comment explaining the design decision and why portability is not a concern (GroveDB targets only little-endian platforms)

**Audit finding:** L5 — originally flagged as non-portable, resolved as documentation-only (not a bug)

## Test plan

- [x] No code change — documentation only
- [x] Existing tests pass unchanged

### DIFF
--- a/storage/src/rocksdb_storage/storage.rs
+++ b/storage/src/rocksdb_storage/storage.rs
@@ -168,7 +168,12 @@ impl RocksDbStorage {
             ));
         }
 
-        res.extend(segments_count.to_be_bytes());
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Approved

PR #545, opened 2026-03-06
fix: return error instead of silently dropping writes when batch is None (L6)

QuantumExplorer:
## Summary
- Transaction storage context write methods (`put`, `put_aux`, `put_root`, `put_meta`, `delete`, `delete_aux`, `delete_root`, `delete_meta`) silently discarded writes when `batch` was `None`, returning `Ok(())` as if the write succeeded
- Now returns `StorageError` so callers are immediately aware the write was not persisted
- 8 methods fixed in `PrefixedRocksDbTransactionContext`

## Test plan
- [x] `cargo build` — clean
- [x] `cargo test -p grovedb --lib` — 1274 passed


QuantumExplorer:
Closing — batch=None is the intentional read-only mode. All production write paths pass Some(&batch). The silent no-op is by design, not a bug.

PR #546, opened 2026-03-06
fix: use explicit .value for discarded costs in Restorer (L9)

QuantumExplorer:
## Summary
- Replace ambiguous `.unwrap()` calls on `CostContext` with explicit `.value` field access in `Restorer::write_chunk` and `Restorer::rewrite_heights`
- Add comments clarifying costs are intentionally discarded since the Restorer does not track operation costs

## Test plan
- [x] `cargo test -p grovedb-merk --lib -- restore` — 13 passed


QuantumExplorer (commented):
approved

PR #548, opened 2026-03-06
fix: replace unreachable!() with error returns in Walker::detach (L11)

QuantumExplorer:
## Summary
- Replace two `unreachable!()` calls in `Walker::detach` with `CorruptedState` error returns
- These branches could be reached with corrupted tree data (e.g., link claims tree is loaded but detach returns None, or link is not `Reference` variant when expected)
- Fails gracefully instead of panicking

## Test plan
- [x] `cargo test -p grovedb-merk --lib` — 332 passed


QuantumExplorer (commented):
Reviewed

PR #550, opened 2026-03-06
fix: return error on deserialization failure during proof generation (L14)

QuantumExplorer:
## Summary
- Replace silent `Err(_) => continue` on Element deserialization errors with `CorruptedData` error returns (2 locations in `generate.rs`)
- Silently skipping corrupted elements would produce incomplete proofs that omit data the verifier expects

## Test plan
- [x] `cargo test -p grovedb --lib -- proof` — 198 passed


QuantumExplorer (commented):
Reviewed

PR #552, opened 2026-03-06
fix: return error instead of panicking on missing flags in JIT update (L20)

QuantumExplorer:
## Summary
- Replace `.unwrap()` on `get_flags_mut().as_mut()` with `.ok_or()` error return in `process_old_element_flags`
- If an element has no flags during just-in-time reference update, returns a `CorruptedCodeExecution` error instead of panicking

## Test plan
- [x] `cargo test -p grovedb --lib -- batch` — 298 passed


QuantumExplorer (commented):
Approved

PR #557, opened 2026-03-06
fix: extend Tree::aggregate_data to handle all chunk node types (H2)

QuantumExplorer:
## Summary

**Severity: High** — Fixes multi-chunk GroveDB restoration failure for subtrees with >~15 elements.

### The Bug

`Tree::aggregate_data()` in `merk/src/proofs/tree.rs` only handled `KVValueHashFeatureType` nodes, returning an error for all other types. However, the chunk producer (`create_proof_node_for_chunk`) generates different node types based on the element type:
- **Items** → `KV` nodes
- **Trees/SumTrees** → `KVValueHash` nodes
- **Items in CountTree** → `KVCount` nodes
- **Trees in CountTree** → `KVValueHashFeatureType` nodes

During multi-chunk restoration, `restore.rs:383` calls `chunk_tree.aggregate_data()` on the root of each chunk. For any chunk whose root is a `KV`, `KVValueHash`, or `KVCount` node, this errored out — breaking restoration for any subtree large enough to require multiple chunks (~15+ elements).

Existing tests only used single-chunk subtrees or raw merk values (which always produce `KVValueHashFeatureType`), masking the bug.

### The Fix

Extended `aggregate_data()` to handle all chunk node types, matching the existing logic in `Child::as_link()`:

| Node Type | Aggregate Data |
|-----------|---------------|
| `KV`, `KVValueHash` | `NoAggregateData` |
| `KVCount` | `ProvableCount(count)` |
| `KVValueHashFeatureType` | from feature_type (unchanged) |
| Other (Hash, KVHash, etc.) | Error |

Updated the existing test that expected an error for `KV` nodes, and added a new test confirming `KV` returns `NoAggregateData`.

## Test plan

- [x] `cargo test -p grovedb-merk` — all 333 tests + 9 doc tests pass
- [x] `cargo test -p grovedb-merk -- restore` — all 13 restore tests pass

### DIFF
--- a/merk/src/proofs/tree.rs
+++ b/merk/src/proofs/tree.rs
@@ -385,10 +385,12 @@ impl Tree {
 
     #[cfg(feature = "minimal")]
     pub(crate) fn aggregate_data(&self) -> Result<AggregateData, Error> {
-        match self.node {
-            Node::KVValueHashFeatureType(.., feature_type) => Ok(feature_type.into()),
+        match &self.node {
+            Node::KVValueHashFeatureType(.., feature_type) => Ok((*feature_type).into()),
+            Node::KVCount(_, _, count) => Ok(AggregateData::ProvableCount(*count)),
+            Node::KV(..) | Node::KVValueHash(..) => Ok(AggregateData::NoAggregateData),
             _ => Err(Error::InvalidProofError(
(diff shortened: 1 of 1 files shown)


QuantumExplorer:
Okay

PR #562, opened 2026-03-06
fix: preserve query flags in merge operations (M5)

QuantumExplorer:
## Summary

Fixes audit finding **M5**: `Query::merge_multiple` and `merge_with` silently discard `left_to_right` and `add_parent_tree_on_subquery` flags from merged queries.

### Problem

Both `merge_multiple` and `merge_with` destructured the `Query` struct using `..`, which silently dropped the `left_to_right` and `add_parent_tree_on_subquery` fields. Only the first query's flags were preserved; all subsequent queries' flags were ignored.

This is particularly problematic for `add_parent_tree_on_subquery` (a v2 feature that includes parent tree elements like CountTree/SumTree in query results). If query A doesn't request parent trees but query B does, the merged result would incorrectly exclude parent trees.

### Fix

- Explicitly destructure all fields (no `..`) to ensure nothing is silently dropped
- `add_parent_tree_on_subquery`: Use OR semantics — if any merged query requests parent tree inclusion, the merged result does too
- `left_to_right`: Explicitly bound to `_` since the merged query can only iterate in one direction (the first query's direction is kept)

This makes the flag handling explicit rather than relying on a catch-all that silently discards fields. If new fields are added to `Query` in the future, the destructuring will produce a compile error rather than silently dropping the new field.

## Test plan

### DIFF
--- a/grovedb-query/src/merge.rs
+++ b/grovedb-query/src/merge.rs
@@ -444,8 +444,13 @@ impl Query {
                 mut items,
                 default_subquery_branch,
                 conditional_subquery_branches,
-                ..
+                left_to_right: _,
+                add_parent_tree_on_subquery,
             } = query;
+            // Preserve add_parent_tree_on_subquery if any query requests it
+            if add_parent_tree_on_subquery {
+                merged_query.add_parent_tree_on_subquery = true;
+            }
             // the searched for items are the union of all items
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Okay

PR #563, opened 2026-03-06
fix: enforce validate_insertion_does_not_override_tree in batch path (M6)

QuantumExplorer:
## Summary

Fixes audit finding **M6**: `validate_insertion_does_not_override_tree` was ignored in the batch execution path, allowing batch inserts to silently overwrite existing trees.

### Problem

The `validate_insertion_does_not_override_tree` option in `BatchApplyOptions` was defined and could be set to `true`, but the batch execution path (`execute_ops_on_path`) never checked it. Only `validate_insertion_does_not_override` was checked — and only for Item/SumItem insertions, not for tree element insertions.

This meant batch operations could overwrite existing trees even when the protection was explicitly enabled, while individual insert operations correctly blocked the overwrite.

### Fix

Added tree override validation in the batch path for three element categories:

1. **Tree elements** (Tree, SumTree, CountTree, etc.): When inserting a tree and `validate_insertion_does_not_override_tree` is set, checks if an existing tree element exists at the key and blocks the insertion if so.

2. **CommitmentTree elements**: Same check as regular tree elements.

3. **Item elements** (Item, SumItem, ItemWithSumItem): When `validate_insertion_does_not_override` is not set but `validate_insertion_does_not_override_tree` is, checks if the existing element is a tree before allowing the insertion. This matches the non-batch insert path behavior.

The validation follows the same pattern as the non-batch insert path: deserialize the existing element and check `is_any_tree()`.

### Test changes

Updated `test_batch_validate_insertion_does_not_override_tree` to correctly expect that overwriting a tree is blocked when the validation is enabled (previously the test incorrectly expected success). Added a second assertion verifying that the override succeeds when the validation is disabled.

## Test plan

- [x] Updated test correctly validates both the blocking and non-blocking cases

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -1664,252 +1664,179 @@ where
                 GroveOp::InsertOnly { element }
                 | GroveOp::InsertOrReplace { element }
                 | GroveOp::Replace { element }
-                | GroveOp::Patch { element, .. } => match &element {
-                    Element::Reference(path_reference, element_max_reference_hop, _) => {
-                        let merk_feature_type = cost_return_on_error_into!(
+                | GroveOp::Patch { element, .. } => {
+                    // Check tree-override protection for all non-reference elements.
+                    if batch_apply_options.validate_insertion_does_not_override_tree
+                        && !matches!(&element, Element::Reference(..))
+                    {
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Okay

PR #564, opened 2026-03-06
fix: document balance_factor overflow impossibility (M7)

QuantumExplorer:
## Summary

**Audit Finding M7**: The original audit flagged a potential arithmetic overflow in `TreeNode::balance_factor()` when casting `u8` child heights to `i8`.

After review, this overflow is **impossible in practice**: AVL tree heights are O(log n), so even with 2^127 elements the height would be ~127, well within `i8` range. A tree that could overflow `u8` child heights (>255) is physically impossible.

- Reverted the `i16` promotion approach
- Added a comment documenting why the `u8 as i8` cast is safe

## Test plan

- [x] `cargo build -p grovedb-merk` — compiles clean
- [x] `cargo test -p grovedb-merk` — all tests pass

## Summary by CodeRabbit

* **No user-facing changes**
  * Internal code improvements with no impact on functionality or user experience.

### DIFF
--- a/merk/src/tree/mod.rs
+++ b/merk/src/tree/mod.rs
@@ -740,9 +740,16 @@ impl TreeNode {
     /// subtree is 2 levels taller than the left subtree.
     #[inline]
     pub const fn balance_factor(&self) -> i8 {
-        let left_height = self.child_height(true) as i8;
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Approved

PR #566, opened 2026-03-06
fix: return accumulated costs from commit_batch (M9)

QuantumExplorer:
## Summary

Fixes audit finding **M9**: `PrefixedRocksDbImmediateStorageContext::commit_batch` discards all accumulated costs.

### Problem

`commit_batch` was returning `Default::default()` as its cost context, which is a zero-cost `OperationCost`. The batch's `cost_acc` field (containing seek counts, storage costs accumulated during batch operations) was silently dropped. This caused cost tracking to underestimate actual resource usage during replication.

### Fix

One-line change: use `batch.cost_acc` instead of `Default::default()` in the `wrap_with_cost` call.

## Test plan

- [x] All storage tests pass (`cargo test -p grovedb-storage`)


QuantumExplorer:
Verified that this does not break Platform.

PR #568, opened 2026-03-06
fix: replace unreachable!() with overwrite in Query::insert_item (M11)

QuantumExplorer:
## Summary

Fixes audit finding **M11**: `unreachable!()` in `Query::insert_item` and `AggregateSumQuery::insert_item` — `binary_search` finding a duplicate caused a panic.

### Fix

When `binary_search` finds an existing equal item (after retain/merge), overwrite it at that position instead of panicking. Applied to both `Query` and `AggregateSumQuery`.

## Test plan

- [x] All grovedb-query tests pass
- [x] All 1274 grovedb lib tests pass


QuantumExplorer (commented):
Approved

PR #569, opened 2026-03-06
fix: return Result from Link::into_reference instead of panicking

QuantumExplorer:
## Summary

**Audit Finding M13**: `Link::into_reference()` panicked on `Modified` and `Uncommitted` variants, which could crash the process if called in an unexpected state during tree pruning in `TreeNode::commit()`.

- Changed `Link::into_reference()` to return `std::result::Result<Self, crate::Error>` instead of panicking
- `Modified` and `Uncommitted` variants now return `Err(Error::CorruptedState(...))` instead of `panic!(...)`
- Updated both callers in `TreeNode::commit()` (`merk/src/tree/mod.rs`) to propagate errors using `cost_return_on_error_no_add!` with `.transpose()` on the `Option<Result<Link>>` 
- Converted two `#[should_panic]` tests (`modified_into_reference`, `uncommitted_into_reference`) to assert `Result::is_err()` instead
- Added `.unwrap()` to existing tests that call `into_reference()` on `Reference` and `Loaded` variants

### Why this matters

In production, a `panic!` in `into_reference()` would terminate the process. While `Modified` and `Uncommitted` links shouldn't normally reach the pruning step (they should be committed first), returning an error allows callers to handle the situation gracefully — e.g., log the corrupted state and recover — rather than crashing.

## Test plan

- [x] `cargo build -p grovedb-merk` — compiles clean
- [x] `cargo build` — full workspace compiles clean
- [x] `cargo test -p grovedb-merk` — all 332 tests pass
- [x] `modified_into_reference` test verifies `Err` is returned for `Modified` variant
- [x] `uncommitted_into_reference` test verifies `Err` is returned for `Uncommitted` variant

### DIFF
--- a/merk/src/tree/link.rs
+++ b/merk/src/tree/link.rs
@@ -218,25 +218,29 @@ impl Link {
         }
     }
 
-    /// Consumes the link and converts to variant `Link::Reference`. Panics if
-    /// the link is of variant `Link::Modified` or `Link::Uncommitted`.
+    /// Consumes the link and converts to variant `Link::Reference`. Returns an
+    /// error if the link is of variant `Link::Modified` or `Link::Uncommitted`.
     #[inline]
-    pub fn into_reference(self) -> Self {
+    pub fn into_reference(self) -> std::result::Result<Self, crate::Error> {
         match self {
-            Link::Reference { .. } => self,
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Approved

QuantumExplorer (commented):
I guess

PR #571, opened 2026-03-07
fix: validate version byte in Query::borrow_decode (L1)

QuantumExplorer:
## Summary

**Audit Finding L1**: `BorrowDecode` for `Query` read and discarded the version byte without validating it. Future v2+ encodings would be silently accepted and decoded incorrectly instead of producing a clean error.

- Added `version != 1` check matching the existing `Decode` impl (line 75)
- Returns `DecodeError::Other("unsupported Query encoding version")` on mismatch

## Test plan

- [x] `cargo build -p grovedb-query` compiles clean
- [x] `cargo test -p grovedb-query` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistently validate query encoding version across all processing paths; unsupported formats are now rejected with clear error messages.
* **Tests**
  * Added a test to ensure borrow-decoding rejects unsupported query encoding versions and surfaces the appropriate error.

### DIFF
--- a/grovedb-query/src/query.rs
+++ b/grovedb-query/src/query.rs
@@ -118,7 +118,10 @@ impl<'de, Context> BorrowDecode<'de, Context> for Query {
     fn borrow_decode<D: bincode::de::BorrowDecoder<'de, Context = Context>>(
         decoder: &mut D,
     ) -> Result<Self, DecodeError> {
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Approved

PR #573, opened 2026-03-07
fix: use parent tree type for DeleteTree batch operations (L3)

QuantumExplorer:
## Summary

**Audit Finding L3**: `GroveOp::DeleteTree(tree_type)` passed the deleted subtree's `tree_type` to `Element::delete_into_batch_operations()` instead of `in_tree_type` (the parent tree type). The `in_tree_type` parameter determines whether to emit `DeleteLayered` vs `DeleteLayeredMaybeSpecialized` merk operations.

- Changed from `tree_type` to `in_tree_type`, matching the `GroveOp::Delete` arm
- Currently non-exploitable since both code paths are handled identically downstream, but semantically incorrect

## Test plan

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -1895,13 +1895,13 @@ where
                         )
                     );
                 }
-                GroveOp::DeleteTree(tree_type) => {
+                GroveOp::DeleteTree(_tree_type) => {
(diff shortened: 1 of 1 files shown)


QuantumExplorer:
We are safe to merge this in because Platform never has deleted a tree with a different sub tree type. And has no code to do so yet.

PR #574, opened 2026-03-07
fix: add visited-set cycle detection to batch reference resolution (L4)

QuantumExplorer:
## Summary

**Audit Finding L4**: Batch reference resolution used only a hop counter (max 10) to detect reference cycles, unlike the non-batch path which uses both a hop counter and a visited-path set. Short cycles (e.g., A→B→A) would waste up to 10 hops of I/O before detection.

- Added `HashSet<Vec<Vec<u8>>>` visited set threaded through `follow_reference_get_value_hash`, `process_reference`, and `process_reference_with_hop_count_greater_than_one`
- Returns `Error::CyclicReference` immediately when a previously visited path is encountered, matching the non-batch behavior
- Each external entry point (2 call sites in `execute_ops_on_path`) creates a fresh `HashSet`

## Test plan

- [x] `cargo test -p grovedb --lib -- reference` — all 93 reference tests pass (including `test_cyclic_reference_detected`)

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -28,7 +28,7 @@ mod single_sum_item_insert_cost_tests;
 use core::fmt;
 use std::{
     cmp::Ordering,
-    collections::{btree_map::Entry, hash_map::Entry as HashMapEntry, BTreeMap, HashMap},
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Approved

PR #575, opened 2026-03-07
fix: add InsertOnly and DeleteTree to apply_operations_without_batching (L5)

QuantumExplorer:
## Summary

**Audit Finding L5**: `apply_operations_without_batching` returned `NotSupported` for `InsertOnly`, `Patch`, `RefreshReference`, and `DeleteTree` operations, even though `apply_batch` accepts them all.

- **InsertOnly**: Now supported — calls `self.insert()` with `validate_insertion_does_not_override = true`
- **DeleteTree**: Now supported — merged with the `Delete` arm since `self.delete()` handles both items and trees
- **Patch / RefreshReference**: Kept as `NotSupported` with an explicit error message — these are batch-only operations with no non-batch equivalent method
- Replaced the wildcard `_` catch-all with explicit match arms for remaining internal tree ops (`ReplaceTreeRootKey`, `InsertTreeWithRootHash`, etc.)

## Test plan

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -2720,7 +2720,33 @@ impl GroveDb {
                         )
                     );
                 }
-                GroveOp::Delete => {
+                GroveOp::InsertOnly { element } => {
+                    let path_slices: Vec<&[u8]> =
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Sure

PR #576, opened 2026-03-07
docs: document put-wins semantics in StorageBatch::delete (L6)

QuantumExplorer:
## Summary

**Audit Finding L6**: `StorageBatch::delete` silently drops delete operations when a put for the same key already exists in the batch. This behavior is tested and intentional (during tree rebalancing, a node may be deleted from one position and re-inserted at another within the same commit), but was undocumented.

- Added doc comments to `delete()`, `delete_aux()`, and `delete_root()` explaining the put-wins semantics
- No behavior change — documentation only

## Test plan

- [x] `cargo build -p grovedb-storage` compiles clean

### DIFF
--- a/storage/src/storage.rs
+++ b/storage/src/storage.rs
@@ -407,7 +407,12 @@ impl StorageBatch {
         );
     }
 
-    /// Add deferred `delete` operation
+    /// Add deferred `delete` operation.
+    ///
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Approved

PR #578, opened 2026-03-07
fix: remove eprintln in commit_session (L8)

QuantumExplorer:
## Summary

**Audit Finding L8**: `commit_session` used `inspect_err(|e| eprintln!(...))` to print errors to stderr. This leaks error details in production and is redundant since the error is already returned via `Result` for the caller to handle.

- Removed the `.inspect_err()` call, keeping the clean `session.commit()` return

## Test plan


QuantumExplorer (commented):
Approved

PR #580, opened 2026-03-07
fix: use try_from for path length comparison in delete_up_tree (L10)

QuantumExplorer:
## Summary

**Audit Finding L10**: `path.to_vec().len() as u16` in `add_delete_operations_for_delete_up_tree_while_empty` would silently truncate on paths with >65535 segments, causing incorrect comparison with `stop_path_height`.

- Changed to `u16::try_from(path.to_vec().len()).unwrap_or(u16::MAX)` — paths exceeding u16 range will never match a u16 stop height, which is correct behavior

## Test plan

- [x] `cargo test -p grovedb --lib -- delete` — all 119 delete tests pass


QuantumExplorer (commented):
Reviewed

PR #581, opened 2026-03-07
fix: propagate IO errors in Visualize for KeyInfoPath (L11)

QuantumExplorer:
## Summary

**Audit Finding L11**: `Visualize for KeyInfoPath` used `.unwrap()` on fallible `Write` operations. Since the method already returns `std::io::Result`, these should use `?` to propagate errors instead of panicking.

- Changed 3 `.unwrap()` calls to `?` in `KeyInfoPath::visualize()`

## Test plan


QuantumExplorer (commented):
Reviewed

PR #584, opened 2026-03-07
fix: distinguish sync-never-started from sync-completed (L14)

QuantumExplorer:
## Summary

**Audit Finding L14**: `is_sync_completed()` returned `true` when `current_prefixes` was empty, making it impossible to distinguish "sync completed" from "sync never started."

- Added early return `false` when both `current_prefixes` and `processed_prefixes` are empty
- A freshly created `MultiStateSyncSession` (before any `apply_chunk`) now correctly reports sync as not completed

## Test plan

- [x] `cargo test -p grovedb --lib -- replication` — all 39 replication tests pass
- [x] `is_sync_completed_transitions` test verifies the full lifecycle


QuantumExplorer (commented):
Approved

PR #585, opened 2026-03-07
fix: add u32 bounds checks to chunk blob serialization (L15)

QuantumExplorer:
## Summary

- `serialize_fixed` and `serialize_variable` in `grovedb-bulk-append-tree/src/chunk.rs` used `as u32` casts for entry count and entry sizes, which silently truncate on 64-bit platforms if values exceed `u32::MAX`
- Replaced with `try_into()` that returns `BulkAppendError::InvalidInput` instead of producing a corrupt blob with silently truncated length fields
- Changed `serialize_chunk_blob` return type from `Vec<u8>` to `Result<Vec<u8>, BulkAppendError>` and updated all callers

**Audit finding:** L15 — `serialize_variable` and `serialize_fixed` truncate entry length/count to u32

## Test plan

- [x] All 68 `grovedb-bulk-append-tree` tests pass
- [x] Build succeeds
- [x] Existing roundtrip tests validate that valid inputs still serialize correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling in chunk serialization to gracefully handle invalid input sizes instead of panicking.
  * Added validation for entry count and size parameters to prevent overflow conditions.

### DIFF
--- a/grovedb-bulk-append-tree/src/chunk.rs
+++ b/grovedb-bulk-append-tree/src/chunk.rs
@@ -33,9 +33,9 @@ const MAX_CHUNK_ENTRIES: usize = 1 << 20;
 /// - Otherwise -> variable-size format
 ///
 /// Returns an empty `Vec` for an empty slice (no header byte).
-pub fn serialize_chunk_blob(entries: &[Vec<u8>]) -> Vec<u8> {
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Okay, why not

PR #586, opened 2026-03-07
fix: enforce InsertOnly semantics in batch execution path

QuantumExplorer:
## Summary

**Audit finding M-NEW-1**: `InsertOnly` did not enforce insert-only semantics in the batch execution path.

- In `execute_ops_on_path`, `InsertOnly` was grouped with `InsertOrReplace`/`Replace`/`Patch` in a single match arm
- The check to use `insert_if_not_exists_into_batch_operations` depended solely on `batch_apply_options.validate_insertion_does_not_override`
- The non-batch fallback path (`apply_operations_without_batching`) correctly forced `validate_insertion_does_not_override = true` for `InsertOnly`
- **Result**: Batch `InsertOnly` silently overwrote existing elements unless the caller explicitly set the batch option

## Fix

Capture whether the op is `InsertOnly` before entering the match, and always use `insert_if_not_exists_into_batch_operations` for `InsertOnly` ops regardless of the batch option.

## Test plan

- [x] New test: `test_batch_insert_only_rejects_overwrite_without_explicit_validation_option` — verifies InsertOnly rejects overwrites with default batch options
- [x] All 10 existing InsertOnly tests pass
- [x] `cargo test -p grovedb --lib insert_only`

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -1674,6 +1674,7 @@ where
 
         let mut batch_operations: Vec<(Vec<u8>, Op)> = vec![];
         for (key_info, op) in ops_at_path_by_key.into_iter() {
+            let is_insert_only = matches!(&op, GroveOp::InsertOnly { .. });
             match op {
                 GroveOp::InsertOnly { element }
                 | GroveOp::InsertOrReplace { element }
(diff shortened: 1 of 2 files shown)


QuantumExplorer:
This needs to be versioned, otherwise platform will not work as intended

QuantumExplorer:
This is being replaced by #595

PR #588, opened 2026-03-07
fix: prevent u16 truncation in query limit tracking

QuantumExplorer:
## Summary

**Audit findings M-NEW-5 and M-NEW-6**: `usize`/`u64` values were cast to `u16` without bounds checking when decrementing query limits.

Affected locations:
- `grovedb/src/element/query.rs:403` — `sub_elements.len() as u16`
- `grovedb/src/operations/proof/generate.rs:1381` — BulkAppendTree range count `as u16`
- `grovedb/src/operations/proof/generate.rs:1294` — MMR proof leaves count `as u16`
- `grovedb/src/operations/proof/generate.rs:1538` — DenseTree proof entries count `as u16`

All now clamp to `u16::MAX` before casting, ensuring the limit is always decremented by at least `u16::MAX` (effectively zeroing it) rather than wrapping to a small value.

## Test plan

- [x] `cargo build -p grovedb`

### DIFF
--- a/grovedb/src/element/query.rs
+++ b/grovedb/src/element/query.rs
@@ -400,7 +400,8 @@ impl ElementQueryExtensions for Element {
                         // we should decrease by 1 in this case
                         *limit = limit.saturating_sub(1);
                     } else {
-                        *limit = limit.saturating_sub(sub_elements.len() as u16);
+                        *limit =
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Approved

PR #591, opened 2026-03-07
fix: enforce set_new_transaction precondition in release builds

QuantumExplorer:
## Summary

**Audit finding E-NEW-8**: The safety-critical precondition in `set_new_transaction` that `current_prefixes` must be empty before replacing the transaction was only enforced by `debug_assert!`, which is stripped in release builds.

This is an unsafe function where violating the precondition could cause use-after-free: if storage contexts in `current_prefixes` still hold references to the old transaction when it's replaced, those references become dangling.

## Fix

Replace `debug_assert!` with a runtime check that returns `Err(Error::InternalError(...))`.

## Test plan

- [x] `cargo build -p grovedb`
- [x] Existing replication tests pass


QuantumExplorer (commented):
Approved

PR #594, opened 2026-03-07
fix: version-gate InsertOnly batch semantics to prevent consensus breakage

QuantumExplorer:
## Summary

- Adds `execute_ops_on_path` `FeatureVersion` field to version structs (set to `0` for V1/V2/V3)
- Version-gates `InsertOnly` uniqueness enforcement: V0 uses `insert_into_batch_operations` (no seek, current behavior), V1 will use `insert_if_not_exists_into_batch_operations` (with seek)
- Prevents consensus breakage that would occur if InsertOnly enforcement were applied retroactively — the extra seek changes `OperationCost`, which changes fees and state hashes

## Context

PR #586 proposed enforcing InsertOnly semantics by routing through `insert_if_not_exists_into_batch_operations`. However, this adds a seek operation that changes `OperationCost`, breaking consensus on resync for nodes running v3.0.1+ (token total supply uses InsertOnly). This PR implements the same fix but version-gated, so it can be safely activated in a future protocol version.

## Test plan

- [x] `test_batch_insert_only_allows_overwrite_in_v0` — V0 allows InsertOnly overwrite (current behavior preserved)
- [x] `test_batch_insert_only_rejects_overwrite_in_v1` — V1 rejects InsertOnly overwrite with `OverrideNotAllowed`
- [x] `test_batch_insert_only_new_key_succeeds_in_v1` — V1 allows InsertOnly for new keys
- [x] All 1285 grovedb lib tests pass
- [x] Clippy clean

## Summary by CodeRabbit

* **New Features**
  * Added a new version counter to enable version-gated batch insertion behavior.

* **Behavior Changes**
  * InsertOnly semantics now depend on version: older versions may allow overwrites; newer versions enforce uniqueness by default.

* **Performance / Cost Estimates**
  * Cost accounting for InsertOnly updated to reflect existence checks in newer versions.

* **Tests**
  * Added tests covering InsertOnly overwrite and new-key scenarios across versions.

* **Documentation**
  * Updated inline docs to reflect the new insert path and behavior.

### DIFF
--- a/grovedb-version/src/version/grovedb_versions.rs
+++ b/grovedb-version/src/version/grovedb_versions.rs
@@ -50,6 +50,7 @@ pub struct GroveDBApplyBatchVersions {
     pub apply_batch_with_element_flags_update: FeatureVersion,
     pub apply_partial_batch_with_element_flags_update: FeatureVersion,
     pub estimated_case_operations_for_batch: FeatureVersion,
+    pub execute_ops_on_path: FeatureVersion,
 }
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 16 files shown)


QuantumExplorer:
Closing: InsertOnly semantics were wrong. Will split into insert_only_known_to_not_exist and insert_if_not_exists instead.

PR #599, opened 2026-03-08
fix: batch DeleteTree now enforces emptiness check and cleans up subtree storage (H1, H2)

QuantumExplorer:
## Summary

Two related fixes for batch DeleteTree operations:

**H2 — Emptiness check**: Before `apply_body` runs, each `DeleteTree` op is now checked for subtree emptiness when `allow_deleting_non_empty_trees` is false. Standard Merk trees use `is_empty_tree_except(batch_deleted_keys)` to correctly account for sibling deletes in the same batch. Non-Merk trees use `element.non_merk_entry_count()`. When a non-empty tree is found and `deleting_non_empty_trees_returns_error` is true, returns `Error::DeletingNonEmptyTree`; otherwise silently skips the op.

**H1 — Storage cleanup**: After `apply_body` completes, for each standard Merk tree deletion, `find_subtrees` recursively discovers all nested subtrees and `storage.clear()` is applied to each. This matches the non-batch delete behavior in `delete_internal_on_transaction`. Previously, only the parent key was removed, leaving orphaned storage data that could corrupt a later re-insert at the same path.

Both fixes apply to the full batch and partial batch paths.

## Test plan

- [x] `test_batch_delete_tree_non_empty_should_fail_when_not_allowed` — H2: error on non-empty tree delete
- [x] `test_batch_delete_tree_non_empty_succeeds_when_allowed` — H2: success when allowed
- [x] `test_batch_delete_empty_tree_succeeds` — H2: empty tree always works
- [x] `test_batch_delete_tree_cleans_up_subtree_storage` — H1: no orphaned storage after delete
- [x] `test_batch_delete_tree_then_reinsert_produces_clean_tree` — H1: clean re-insertion after delete
- [x] 3 existing tests updated to set `allow_deleting_non_empty_trees: true` (they intentionally delete non-empty trees)
- [x] All 1295 tests pass

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -5601,6 +5601,335 @@ mod tests {
         }
     }
 
+    // ===================================================================
+    // InsertIfNotExists and InsertWithKnownToNotAlreadyExist tests
+    // ===================================================================
+
+    #[test]
+    fn test_batch_insert_if_not_exists_succeeds_for_new_key() {
+        // InsertIfNotExists should succeed when the key does not exist.
+        let grove_version = GroveVersion::latest();
+        let db = make_test_grovedb(grove_version);
+
+        let ops = vec![QualifiedGroveDbOp::insert_if_not_exists_op(
+            vec![TEST_LEAF.to_vec()],
+            b"new_key".to_vec(),
+            Element::new_item(b"value".to_vec()),
+        )];
+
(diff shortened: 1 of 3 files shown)


QuantumExplorer (commented):
Approved, this would cause Platform to no longer work, but PR https://github.com/dashpay/platform/pull/3210 will solve that on Platform side.

PR #601, opened 2026-03-08
fix: add visited-set cycle detection to aggregate sum query reference following (M5)

QuantumExplorer:
## Summary

- Add `HashSet<Vec<Vec<u8>>>` visited set to the reference-following loop in `aggregate_sum_path_query_push`
- Detect cyclic references immediately with `Error::CyclicReference` instead of wasting reads until the hop limit triggers a misleading `ReferenceLimit` error
- Matches the pattern used in every other reference-following path (`follow_reference` in `operations/get/mod.rs`, `follow_reference_get_value_hash` in `batch/mod.rs`, etc.)

**Audit finding M5**: The aggregate sum query reference following relied solely on `MAX_AGGREGATE_REFERENCE_HOPS = 3` for termination. Unlike all other reference-following paths, it lacked explicit cycle detection via a visited set.

## Test plan

- [x] `test_cyclic_reference_detected_in_aggregate_sum_query` — two mutually-referencing elements, verifies `Error::CyclicReference`
- [x] `test_self_referencing_element_detected_in_aggregate_sum_query` — self-referencing element, verifies `Error::CyclicReference`
- [x] All 1284 grovedb tests pass
- [x] Clippy clean

### DIFF
--- a/grovedb/src/element/aggregate_sum_query/mod.rs
+++ b/grovedb/src/element/aggregate_sum_query/mod.rs
@@ -1,6 +1,8 @@
 //! Query
 //! Implements functions in Element for querying
 
+#[cfg(feature = "minimal")]
+use std::collections::HashSet;
 use std::fmt;
 
 use crate::element::SumValue;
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Reviewed

PR #602, opened 2026-03-08
fix: replace panic!() with error returns in TreeNode/Walker attach/detach_expect (M2)

QuantumExplorer:
## Summary

- `TreeNode::attach()`: returns `Result<Self, Error>` instead of panicking when child slot is occupied
- `TreeNode::detach_expect()`: returns `Result<(Self, Self), Error>` instead of panicking when expected child is `None`
- `Walker::detach_expect()`: returns error via `flat_map_ok` instead of panicking
- `Walker::attach()`: returns `Result<Self, Error>` to propagate attach errors
- Updated `walk()`, `walk_expect()` to propagate errors
- Updated `rotate()`, `promote_edge()`, `remove_edge()` in ops.rs to handle Result returns with `cost_return_on_error_no_add!`

**Audit finding M2**: These methods are called during core AVL tree operations (rotations, removes) on every insert/delete. If tree state is corrupted (e.g., storage returns bad data), the node would crash instead of returning a recoverable error.

## Test plan

- [x] `attach_existing` test changed from `#[should_panic]` to asserting `result.is_err()`
- [x] All 230 merk tests pass
- [x] Full workspace builds
- [x] Clippy clean

### DIFF
--- a/merk/src/proofs/query/merk_integration_tests.rs
+++ b/merk/src/proofs/query/merk_integration_tests.rs
@@ -39,10 +39,12 @@ fn make_3_node_tree() -> TreeNode {
             true,
             Some(TreeNode::new(vec![3], vec![3], None, BasicMerkNode).unwrap()),
         )
+        .unwrap()
         .attach(
             false,
             Some(TreeNode::new(vec![7], vec![7], None, BasicMerkNode).unwrap()),
-        );
+        )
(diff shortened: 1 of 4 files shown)


QuantumExplorer (commented):
Sure, looks like a good refactor

PR #604, opened 2026-03-08
fix: return error instead of silently discarding writes when storage batch is None (M3)

QuantumExplorer:
## Summary

- All 9 write operations in `PrefixedRocksDbTransactionContext` (`put`, `put_aux`, `put_root`, `put_meta`, `delete`, `delete_aux`, `delete_root`, `delete_meta`, `commit_batch`) now return `Error::StorageError` when `self.batch` is `None`, instead of silently returning `Ok(())`
- Fixed 4 merk tests that were unknowingly relying on the silent-discard behavior (writes were silently lost) — they now provide a `StorageBatch`
- Read operations continue to work correctly with `None` batch

**Audit finding M3**: All write operations silently succeeded and discarded data when no batch was provided. No legitimate code path writes through a batchless context — all batchless call sites only perform reads. The silent success made it impossible for callers to detect data loss.

## Test plan

- [x] `test_write_operations_error_when_batch_is_none` — verifies all 9 write operations return errors
- [x] `test_read_operations_succeed_when_batch_is_none` — verifies reads still work with None batch
- [x] 4 merk tests fixed to use proper StorageBatch
- [x] All 2000+ workspace tests pass
- [x] Clippy clean

### DIFF
--- a/merk/src/element/get.rs
+++ b/merk/src/element/get.rs
@@ -609,8 +609,9 @@ mod tests {
             .unwrap()
             .unwrap();
 
+        let batch2 = StorageBatch::new();
         let ctx = storage
-            .get_transactional_storage_context(SubtreePath::empty(), None, &transaction)
+            .get_transactional_storage_context(SubtreePath::empty(), Some(&batch2), &transaction)
             .unwrap();
         let mut merk = Merk::open_base(
             ctx,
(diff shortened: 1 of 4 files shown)


QuantumExplorer:
Okay makes sense

PR #606, opened 2026-03-08
fix: cap batch reference recursion to MAX_REFERENCE_HOPS (L2)

QuantumExplorer:
## Summary

- Cap `recursions_allowed` to `MAX_REFERENCE_HOPS` (10) at the entry point of `follow_reference_get_value_hash`, regardless of user-provided `element_max_reference_hop`
- Prevents a reference with `max_reference_hop = Some(255)` from creating 255 stack frames via mutual recursion between `follow_reference_get_value_hash`, `process_reference`, and `process_reference_with_hop_count_greater_than_one`
- Single-line fix: `let recursions_allowed = recursions_allowed.min(MAX_REFERENCE_HOPS as u8);`

**Audit finding L2**: The batch reference resolution used mutual recursion bounded by `recursions_allowed` which could be set up to 255 via `MaxReferenceHop`. The non-batch path already caps at `MAX_REFERENCE_HOPS = 10`.

## Test plan

- [x] `test_batch_reference_hop_count_capped_to_max` — chain of 11 refs with `max_hop=255` fails with `ReferenceLimit`; chain of 10 succeeds
- [x] All existing tests pass
- [x] Clippy clean

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -1470,6 +1470,9 @@ where
         ) -> Result<(StorageRemovedBytes, StorageRemovedBytes), Error>,
     {
         let mut cost = OperationCost::default();
+        // Cap recursion depth to MAX_REFERENCE_HOPS to prevent excessive stack
+        // depth even if the user-provided element_max_reference_hop is larger.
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
fair enough

PR #607, opened 2026-03-08
fix: align Link::encoding_cost() with actual varint encoding (L1)

QuantumExplorer:
## Summary

- Replace fixed-size assumptions in `Link::encoding_cost()` with delegation to `encoding_length()`, which correctly computes varint sizes
- Previously, Sum/Count/ProvableCount assumed 8 bytes fixed; actual varint is 1-10 bytes depending on value. This caused inaccurate cost/fee calculations.

**Audit finding L1**: `encoding_cost()` used fixed byte sizes while `encode_into()` and `encoding_length()` used varint encoding. Small values were overcharged by ~7 bytes; very large values could be undercharged by ~2 bytes.

## Test plan

- [x] `encoding_cost_matches_encoding_length_for_all_aggregate_types` — tests all 7 AggregateData variants with small and large values, verifies encoding_cost == encoding_length == actual encoded size
- [x] All 338 merk tests pass
- [x] Clippy clean

### DIFF
--- a/merk/src/tree/link.rs
+++ b/merk/src/tree/link.rs
@@ -270,74 +270,15 @@ impl Link {
         not_prefixed_key_len + HASH_LENGTH_U32 + 4 + sum_tree_cost
     }
 
-    /// The encoding cost is always 8 bytes for the sum instead of a varint
+    /// Returns the actual encoded size of this link in bytes.
+    ///
+    /// This delegates to `encoding_length()` to ensure consistency with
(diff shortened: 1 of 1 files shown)


QuantumExplorer:
This is Claude. Closing this PR — the fixed-size cost estimation in `Link::encoding_cost()` is intentional. Using exact varint sizes would be a breaking change to fee calculations, and the current fixed-size approach is preferred for simplicity and consistency. A documentation PR will be created to explain this design choice.

PR #613, opened 2026-03-09
fix: enforce depth limit during Query/SubqueryBranch deserialization

QuantumExplorer:
## Summary

- `Query` and `SubqueryBranch` form a mutual recursion during bincode deserialization (`Query` → `SubqueryBranch` → `Option<Box<Query>>` → repeat)
- No depth limit was enforced during decoding — only flat dimensions (item count, branch count) were bounded
- A crafted payload with ~19M nesting levels within 256MB could cause stack overflow
- Added `MAX_SUBQUERY_DECODE_DEPTH = 64` (matches existing `MAX_TERMINAL_KEYS_DEPTH`) enforced during deserialization via custom `Decode` implementations

## Test plan

- [x] `query_decode_rejects_excessive_subquery_nesting` — verifies decode rejects depth > 64
- [x] `query_borrow_decode_rejects_excessive_subquery_nesting` — same for BorrowDecode path
- [x] `query_decode_round_trips_at_valid_nesting_depth` — verifies depth 10 round-trips correctly
- [x] All existing query tests continue to pass (65 total)

### DIFF
--- a/grovedb-query/src/query.rs
+++ b/grovedb-query/src/query.rs
@@ -88,15 +88,25 @@ const MAX_QUERY_ITEMS: usize = 65_536;
 /// Prevents OOM from malicious inputs with inflated lengths.
 const MAX_CONDITIONAL_BRANCHES: usize = 1024;
 
-impl<Context> Decode<Context> for Query {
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Approved

PR #615, opened 2026-03-09
docs: document emptiness check limitation for same-batch inserts

QuantumExplorer:
## Summary

- Documents that the pre-deletion emptiness check for `DeleteTree` ops only considers Delete/DeleteTree ops when building the exception set — it does not account for Insert ops that would add new keys to the subtree
- This is safe in practice because the consistency check (`verify_consistency_of_operations`) rejects batches with conflicting insert + delete-tree ops before the emptiness check runs
- The limitation only applies when `disable_operation_consistency_check = true`, in which case the caller has accepted responsibility for ensuring no such conflicts exist
- Added documentation comments at both emptiness check sites in `batch/mod.rs` and on the `disable_operation_consistency_check` option in `batch/options.rs`

## Test plan

- [x] Documentation-only change — no behavioral changes
- [x] `cargo build` passes

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -3390,6 +3390,25 @@ impl GroveDb {
                         // Standard Merk trees: use is_empty_tree_except to
                         // account for other delete ops in the same batch that
                         // target this subtree.
+                        //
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Approved

PR #616, opened 2026-03-09
docs: document non-atomic delete cost computation (TOCTOU)

QuantumExplorer:
## Summary

- Documents the non-atomic freed-size cost computation in `RocksDbStorage::continue_write_batch` as a known design limitation rather than a bug
- Adds comprehensive doc comment to `continue_write_batch` explaining the 4 reasons the TOCTOU-style limitation is acceptable
- Replaces all `TODO: fix not atomic freed size computation` and `todo: improve deletion` comments with references to the rationale
- Documents `cost_info: None` usage at each call site (`Merk::clear`, `Merk::delete_meta`, `Merk::set_base_root_key`, `PrefixedRocksDbTransactionContext::clear`)

## Investigation findings

The `continue_write_batch` method reads committed DB state (`self.db.get()`) to estimate freed bytes when a delete operation has no pre-computed `cost_info`. Investigation confirmed this is **not exploitable** because:

1. **Main path always provides cost**: `Merk::commit()` always passes `Some(storage_cost)` via `KeyUpdates::deleted_keys`
2. **None fallback is cleanup-only**: Only reached by `clear()`, `delete_meta()`, `set_base_root_key(None)`, and `delete_aux()` without cost info
3. **Put-wins semantics**: `StorageBatch::delete()` drops deletes when a put for the same key already exists in the batch
4. **Safe direction**: Underestimation of freed bytes (when key exists only in uncommitted transaction) is the safe direction for cost accounting

## Test plan

- [x] `cargo check -p grovedb-storage -p grovedb-merk` passes
- [x] `cargo test -p grovedb-storage --features rocksdb_storage` passes (27 tests)
- [x] Pre-commit hooks pass (cargo fmt, typos, etc.)
- [ ] CI passes

### DIFF
--- a/merk/src/merk/clear.rs
+++ b/merk/src/merk/clear.rs
@@ -7,7 +7,12 @@ impl<'db, S> Merk<S>
 where
     S: StorageContext<'db>,
 {
-    /// Deletes tree data
+    /// Deletes all tree data (nodes) from storage.
+    ///
+    /// Deletes are issued with `cost_info: None`, so freed-bytes cost is
+    /// estimated from committed DB state. This is acceptable for a
+    /// bulk-clear operation. See `RocksDbStorage::continue_write_batch`
+    /// for the full rationale.
     pub fn clear(&mut self) -> CostResult<(), Error> {
         let mut cost = OperationCost::default();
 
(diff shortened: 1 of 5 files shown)


QuantumExplorer:
Okay

PR #617, opened 2026-03-09
docs: document single-writer transaction requirement

QuantumExplorer:
## Summary

- Documents the single-writer transaction constraint for GroveDb, addressing audit finding #9 about optimistic transaction isolation gaps
- Adds doc comments to `GroveDb` struct, `start_transaction`, `commit_transaction`, `rollback_transaction`, the `Storage` trait, and `RocksDbStorage` struct
- Explains why this is a documentation issue rather than a code fix: Rust's type system already prevents the worst concurrent access patterns (`!Sync` on `StorageBatch`/`MerkCache`), RocksDB catches conflicting writes at commit time, and Dash Platform processes blocks sequentially

## Analysis

GroveDb uses RocksDB's `OptimisticTransactionDB` which is `Send + Sync`, making `GroveDb` itself shareable across threads. However, there is no `Mutex` or `RwLock` enforcing single-writer access. The analysis found:

1. **RocksDB level**: `OptimisticTransactionDB` permits concurrent transactions but detects write conflicts at commit time (returning `Busy`/`TryAgain`)
2. **Rust type system**: `StorageBatch` uses `RefCell` (`!Sync`) and `MerkCache` uses `UnsafeCell` (`!Sync`), preventing sharing a single transaction's state across threads
3. **GroveDb level**: No application-level mutex, but in-memory Merk tree state (hashes, balancing, root propagation) built during a transaction cannot be cheaply unwound on commit failure
4. **Dash Platform**: Block processing is sequential, naturally satisfying the single-writer constraint

This is a **design constraint** that should be documented, not a code fix. Adding a `Mutex` would add unnecessary overhead for the primary use case and change the API contract.

## Test plan

- [x] `cargo check -p grovedb -p grovedb-storage` compiles cleanly
- [x] `cargo test --doc -p grovedb` passes (3 doc tests including the `start_transaction` example)
- [x] All pre-commit hooks pass (formatting, typos, etc.)

### DIFF
--- a/grovedb/src/lib.rs
+++ b/grovedb/src/lib.rs
@@ -262,7 +262,28 @@ use crate::Error::MerkError;
 #[cfg(feature = "minimal")]
 type Hash = [u8; 32];
 
-/// GroveDb
+/// GroveDb is a hierarchical authenticated data structure database.
+///
+/// # Concurrency and Transaction Safety
+///
+/// `GroveDb` is `Send + Sync` because the underlying RocksDB
+/// `OptimisticTransactionDB` is thread-safe at the storage level. However,
+/// **GroveDb is designed for single-writer access**. Callers must ensure that
+/// at most one write transaction is active at any given time.
+///
+/// While RocksDB's optimistic transaction mechanism will detect conflicting
+/// concurrent writes and fail one transaction at commit time (returning a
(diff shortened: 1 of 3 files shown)


QuantumExplorer:
Yep

PR #620, opened 2026-03-09
docs: document find_subtrees stale state limitation during batch delete

QuantumExplorer:
## Summary

- `find_subtrees` reads from committed transaction state only (passes `None` for the storage batch), so subtrees inserted by the same batch are invisible to it during `DeleteTree` cleanup
- This is safe in practice because `verify_consistency_of_operations` (enabled by default) rejects batches that insert under paths being deleted
- The stale-state window is only reachable when `disable_operation_consistency_check = true`
- Added documentation at:
  - `disable_operation_consistency_check` field: warning about potential storage leak
  - `find_subtrees` method: storage batch visibility section
  - Both `find_subtrees` call sites in batch processing: NOTE comments

## Test plan

- [x] Documentation-only change — no behavioral changes
- [x] All existing batch validation tests pass


QuantumExplorer (commented):
Approved

PR #623, opened 2026-03-09
fix: enforce InsertIfNotExists existence check for Tree and Reference elements

QuantumExplorer:
## Summary
- The batch `InsertIfNotExists` operation only checked for existing elements when inserting `Item`/`SumItem` types
- `Tree`, `SumTree`, `Reference`, and other element types bypassed the existence check entirely, allowing silent overwrites even when `insert_if_not_exists_op` was used
- Added existence checks before the Reference and Tree/SumTree/etc branches in `execute_ops_on_path`, using the same `element_at_key_already_exists` pattern already used for Item types

## Test plan
- [x] `test_batch_insert_if_not_exists_tree_errors_when_exists` — error mode rejects existing tree
- [x] `test_batch_insert_if_not_exists_or_skip_tree_skips_when_exists` — skip mode silently skips existing tree
- [x] `test_batch_insert_if_not_exists_tree_succeeds_for_new_key` — new tree inserts normally
- [x] `test_batch_insert_if_not_exists_reference_errors_when_exists` — error mode rejects existing reference
- [x] `test_batch_insert_if_not_exists_or_skip_reference_skips_when_exists` — skip mode silently skips existing reference
- [x] All existing InsertIfNotExists tests continue to pass

### DIFF
--- a/grovedb/src/batch/mod.rs
+++ b/grovedb/src/batch/mod.rs
@@ -52,8 +52,8 @@ use grovedb_costs::{
 use grovedb_merk::{
     element::{
         costs::ElementCostExtensions, delete::ElementDeleteFromStorageExtensions,
-        get::ElementFetchFromStorageExtensions, insert::ElementInsertToStorageExtensions,
-        tree_type::ElementTreeTypeExtensions,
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Approved

PR #625, opened 2026-03-09
ci: add shared build step to avoid 3x compilation

QuantumExplorer:
## Summary
- Add a `build-test` job that compiles all test binaries once with `cargo llvm-cov nextest --no-run`
- Share the instrumented build artifacts with the 3 test partitions via `actions/cache`
- Test partitions restore pre-built artifacts and only run their shard of tests
- Remove sccache from test jobs (ineffective with `cargo-llvm-cov`)
- Configure `rust-cache` to cache `target/llvm-cov-target/` for cross-run caching

## Problem

Currently, all 3 test partitions independently compile the full workspace with llvm-cov instrumentation (~7 minutes each). This means 3x redundant compilations per CI run (~21 minutes of total compute time for compilation alone).

## Solution

```
Before:  partition 1 ─── [compile 7m] ─── [test]
         partition 2 ─── [compile 7m] ─── [test]    (all parallel)
         partition 3 ─── [compile 7m] ─── [test]

After:   build-test ─── [compile 7m] ─── [cache save]
         partition 1 ─── [cache restore] ─── [test]
         partition 2 ─── [cache restore] ─── [test]  (parallel after build)
         partition 3 ─── [cache restore] ─── [test]
```

**Trade-off**: Wall-clock time increases slightly (build step is sequential before test shards), but total compute time drops from ~21 to ~7 minutes of compilation. Subsequent pushes to the same PR benefit from `rust-cache` making the build step incremental.

## Test plan
- [ ] Verify `build-test` job completes successfully
- [ ] Verify all 3 test partitions restore cache and run without recompilation
- [ ] Verify coverage reports are still uploaded correctly to Codecov
- [ ] Compare total CI time vs baseline on develop

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Restructured continuous integration workflow to improve test execution efficiency through optimized partitioning and artifact caching.
  * Enhanced code coverage instrumentation and reporting for better visibility into code quality metrics during development cycles.

PR #627, opened 2026-03-09
ci: fix build step

QuantumExplorer:
## Summary
- Fix `cargo llvm-cov nextest --no-run` which is deprecated and fails with "not found *.profraw files"

## Bug fix

The build step introduced in #625 used `cargo llvm-cov nextest --no-run`, which is deprecated. It fails because it tries to merge profraw files that don't exist (no tests were run):

```
warning: --no-run is deprecated, use `cargo llvm-cov report` subcommand instead
error: failed to merge profile data: not found *.profraw files
```

Replaced with `cargo llvm-cov build --workspace --all-features --tests`, which correctly builds instrumented test binaries without running them.

## Test plan
- [ ] Verify build-test job completes successfully
- [ ] Verify test partitions restore cache and pass

### DIFF
--- a/.github/workflows/grovedb.yml
+++ b/.github/workflows/grovedb.yml
@@ -34,9 +34,7 @@ jobs:
   build-test:
     name: Build Tests (instrumented)
     needs: detect-changes
-    if: >-
-      needs.detect-changes.outputs.any-code == 'true'
-      && !contains(github.event.pull_request.labels.*.name, 'skip-tests')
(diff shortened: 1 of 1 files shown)


QuantumExplorer:
Closing: reverting the shared build step approach entirely since cargo llvm-cov build --tests may not produce artifacts compatible with cargo llvm-cov nextest, causing recompilation anyway.

PR #632, opened 2026-03-09
test: demonstrate non-empty CountTree count swap vulnerability (C1)

QuantumExplorer:
## Summary

- Adds a test (`non_empty_count_tree_count_swap_in_result_set`) proving that a non-empty CountTree's count metadata can be swapped from 3 to 999 without the verifier detecting the tampering
- The vulnerability exists because when a tree element appears in a proof result set **without** a lower-layer subquery, no `combine_hash` verification is performed — the verifier trusts the `KVValueHash` node's value directly
- The test documents the known vulnerability by asserting the tampered count is accepted (flipping to `Err` expectation once a fix lands)

## Root Cause

In proof verification (`verify_layer_proof` V0/V1), non-empty trees that appear in the result set without being drilled into via subquery skip the `combine_hash(H(value_bytes), child_merk_root) == proof_hash` check. This allows an attacker to substitute the serialized element bytes (changing count, sum, or other metadata) while keeping the same `value_hash`, since `KVValueHash` nodes trust the provided hash.

## Fix Direction

The fix requires prover-side changes: include child merk root hashes for non-empty trees even when no subquery drills into them (either via a minimal lower layer or a new proof field). This enables the verifier to perform `combine_hash` checks on all tree elements.

## Test plan

- [x] Test compiles and passes (`cargo test -p grovedb --lib non_empty_count_tree_count_swap_in_result_set`)
- [x] Test asserts the vulnerability exists (tampered count=999 accepted)
- [x] Test will panic with clear message when the vulnerability is fixed

### DIFF
--- a/grovedb/src/tests/proof_coverage_tests.rs
+++ b/grovedb/src/tests/proof_coverage_tests.rs
@@ -6416,4 +6416,176 @@ mod tests {
             err
         );
     }
+
+    // =========================================================================
+    // Non-empty tree value swap investigation
+    //
+    // When a non-empty tree (e.g. CountTree with 3 items) appears in the
+    // result set WITHOUT the query drilling into it, there is no lower-layer
+    // combine_hash verification. The value bytes can be swapped (e.g. change
+    // the count from 3 to 1) without detection.
+    // =========================================================================
+
+    #[test]
+    fn non_empty_count_tree_count_swap_in_result_set() {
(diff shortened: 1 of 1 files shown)


QuantumExplorer:
Closing — this vulnerability demonstration is superseded by PR #633, which fixes the issue and includes a test (`non_empty_count_tree_count_swap_is_detected`) that verifies the tampering is now rejected.

PR #640, opened 2026-03-10
ci: compile tests once, share archive across shards

QuantumExplorer:
## Summary

- Split the monolithic test job into three stages: **build → test (3 shards) → coverage report**
- Uses `cargo llvm-cov nextest-archive` to compile instrumented binaries once and produce a nextest archive (`tests.tar.zst`)
- Each test shard downloads the archive and runs its partition without recompilation
- A final `coverage-report` job merges `.profraw` data from all shards and generates the combined lcov report for Codecov

This should reduce total CI compute time since compilation (the most expensive step) now happens once instead of three times.

## How it works

| Job | What it does | Depends on |
|-----|-------------|------------|
| `build-tests` | Compiles workspace with LLVM instrumentation, creates `tests.tar.zst` | `detect-changes` |
| `test` (x3) | Downloads archive, runs sharded tests (`--partition count:N/3`), uploads `.profraw` data | `build-tests` |
| `coverage-report` | Downloads archive + all profraw data, runs `cargo llvm-cov report`, uploads to Codecov | `test` |

## Test plan

- [ ] CI passes on this PR (the workflow runs against itself)
- [ ] Coverage report appears on Codecov
- [ ] Compare total CI wall time vs previous approach

## Release Notes

* **Chores**
  * Reorganized CI/CD infrastructure for streamlined test execution and coverage reporting
  * Optimized test artifact handling with partitioned execution for improved efficiency

### DIFF
--- a/.github/workflows/grovedb.yml
+++ b/.github/workflows/grovedb.yml
@@ -62,10 +62,39 @@ jobs:
             echo "needs-tests=${{ steps.filter.outputs.any-code }}" >> "$GITHUB_OUTPUT"
           fi
 
-  test:
-    name: Test (${{ matrix.partition }}/3)
+  build-tests:
+    name: Build Tests
     needs: detect-changes
     if: needs.detect-changes.outputs.needs-tests == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: "false"
+          workspaces: ". -> target/llvm-cov-target"
+      - uses: taiki-e/install-action@cargo-llvm-cov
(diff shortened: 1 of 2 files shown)


QuantumExplorer:
## Closing — artifact transfer overhead negates compilation savings

After testing, the compile-once-and-share-via-archive approach adds significant overhead from artifact upload/download between jobs:

1. **Upload** the nextest archive (~hundreds of MB of instrumented binaries) after build
2. **Download** the archive in each of 3 test shards
3. **Upload** profraw coverage data from each shard
4. **Download** all profraw data + archive again in the coverage-report job

The time spent transferring artifacts between jobs largely negates the compilation time saved, especially since GitHub Actions artifact transfer is not particularly fast.

### Alternative approaches worth exploring
- **Larger runner for the existing single-job-per-shard approach** (e.g., ARM-based `c8g.4xlarge` spot instances as self-hosted runners) — faster compilation without any artifact transfer overhead
- **Better Rust build caching** (sccache with S3 backend) so each shard's compilation is mostly cache hits
- **Reducing the workspace size** so compilation is inherently faster

PR #647, opened 2026-03-10
fix: remove ProveOptions from V1 proofs

QuantumExplorer:
## Summary

- Removes `prove_options` field from `GroveDBProofV1` struct — V1 proofs no longer embed prover-controlled options
- V1 verification now uses `ProveOptions::default()` instead of deserializing options from the untrusted proof
- V0 proofs are unchanged (backwards compatible)
- Updates debugger, depth-limit tests, and doc comments

## Background

`ProveOptions` (specifically `decrease_limit_on_empty_sub_query_result`) was embedded in the serialized proof and trusted by the verifier. A malicious prover could set this flag to `true` to make empty subquery results consume the query limit, causing the verifier to return fewer results than actually exist. Both honest and malicious proofs verified correctly — the verifier couldn't distinguish them.

Since all Platform queries use `decrease_limit_on_empty_sub_query_result = true` (the default), removing it from V1 proofs has no practical impact while closing the attack vector.

## Test plan

- [x] All 22 proof depth limit tests pass
- [x] All 133 proof coverage tests pass
- [x] All 5 trunk proof tests pass
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo build --no-default-features --features verify` builds

## Summary by CodeRabbit

* **Refactor**
  * Simplified V1 proof structure by removing embedded configuration options.
  * V1 proof verification now uses standardized default configuration settings instead of proof-embedded options.
  * Updated documentation to clarify V1 proof behavior.

### DIFF
--- a/grovedb/src/debugger.rs
+++ b/grovedb/src/debugger.rs
@@ -370,7 +370,8 @@ fn proof_to_grovedbg(proof: GroveDBProof) -> Result<grovedbg_types::Proof, crate
         }),
         GroveDBProof::V1(p) => Ok(grovedbg_types::Proof {
             root_layer: v1_proof_layer_to_grovedbg(p.root_layer)?,
-            prove_options: prove_options_to_grovedbg(p.prove_options),
+            // V1 proofs no longer embed ProveOptions; use default for debugger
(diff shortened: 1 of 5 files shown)


QuantumExplorer (commented):
Approved

PR #648, opened 2026-03-11
test: reproduce merge_multiple default_subquery_branch bug (M6)

QuantumExplorer:
## Summary

- Adds a **failing test** proving that `Query::merge_multiple` silently drops the `default_subquery_branch` from the 2nd+ query for overlapping items
- When two queries share `Key(2)` and each has a different `default_subquery_branch`, the merged result only retains the 2nd query's subquery for that key — the 1st query's default is lost entirely
- Root cause: `merge_multiple` (merge.rs:501-504) converts each subsequent query's items into conditional subqueries using that query's default, but never accounts for the fact that overlapping items from the 1st query were using its own default (which is never converted to a conditional for those items)

## Bug details

```
Query A: items=[Key(1), Key(2)], default_subquery selects key "a"
Query B: items=[Key(2), Key(3)], default_subquery selects key "b"

After merge_multiple:
  Key(1) → uses default (A's) → selects "a" ✓
  Key(3) → conditional (B's default) → selects "b" ✓
  Key(2) → conditional (B's default) → selects "b" only ✗
           Should select BOTH "a" AND "b"
```

## Test plan

- [ ] `cargo test -p grovedb-query --test merge_coverage merge_multiple_drops_default_subquery_branch` — currently **FAILS**, proving the bug
- [ ] Fix `merge_multiple` to merge default subquery branches for overlapping items
- [ ] Test should pass after fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed merge behavior so overlapping query items preserve existing conditionals and combine defaults correctly, while non-overlapping items keep their own defaults.

* **Tests**
  * Added extensive tests validating pairwise and multi-way overlaps, range overlap scenarios, empty/missing defaults, and ensuring defaults don’t leak into explicit conditionals.


grovedb-query/src/merge.rs
```diff
@@ -495,12 +478,85 @@ impl Query {
...
+                                item,
+                                default_subquery_branch.clone(),
+                            );
+                        }
+                    }
+
+                    // Items only in old (ours): no action needed, they use
+                    // merged_query's default_subquery_branch
+                }
             }
```
thepastaclaw: **🟡 Suggestion: Duplicated overlap-detection logic between merge_multiple and merge_with**

Lines 492–560 (merge_multiple) and 602–650 (merge_with) contain nearly identical overlap-detection and default-promotion logic, differing only in `merged_query` vs `self`. Could be extracted into a shared helper like `fn promote_default_for_overlapping_items(...)`. Reasonable to defer in a bugfix PR, but worth noting for future cleanup.

<sub>source: ['claude']</sub>

<details>
<summary>🤖 Fix this with AI agents</summary>

```
These findings are from an automated code review. Verify each finding against the current code and only fix it if needed.

In `grovedb-query/src/merge.rs`:
- [SUGGESTION] lines 492-560: Duplicated overlap-detection logic between merge_multiple and merge_with
  Lines 492–560 (merge_multiple) and 602–650 (merge_with) contain nearly identical overlap-detection and default-promotion logic, differing only in `merged_query` vs `self`. Could be extracted into a shared helper like `fn promote_default_for_overlapping_items(...)`. Reasonable to defer in a bugfix PR, but worth noting for future cleanup.
```
</details>

grovedb-query/src/merge.rs
```diff
@@ -495,12 +478,85 @@ impl Query {
...
+                        let remaining = QueryItem::intersect_many_ordered(
+                            &mut old_default_items,
+                            vec![conditional_item],
+                        );
+                        old_default_items = remaining.ours.unwrap_or_default();
+                        if old_default_items.is_empty() {
+                            break;
+                        }
+                    }
+                }
```
thepastaclaw: **🟡 Suggestion: Pre-existing: incoming explicit conditional causes old default to be dropped for overlapping items**

When B has `conditional Key(2)→'x'` and A covers Key(2) via `default='a'`, the merged result gives Key(2) only `'x'` — A's default `'a'` is lost. Under strict union semantics, Key(2) should return BOTH. This is **pre-existing** behavior (not introduced by this PR): the old code subtracted conditional items from `items` identically. The test at `merge_coverage.rs:1251` explicitly documents this as intentional "explicit overrides implicit" semantics. Worth revisiting as a separate issue if strict union behavior is desired.

<sub>source: ['claude', 'codex']</sub>

<details>
<summary>🤖 Fix this with AI agents</summary>

```
These findings are from an automated code review. Verify each finding against the current code and only fix it if needed.

In `grovedb-query/src/merge.rs`:
- [SUGGESTION] lines 496-509: Pre-existing: incoming explicit conditional causes old default to be dropped for overlapping items
  When B has `conditional Key(2)→'x'` and A covers Key(2) via `default='a'`, the merged result gives Key(2) only `'x'` — A's default `'a'` is lost. Under strict union semantics, Key(2) should return BOTH. This is **pre-existing** behavior (not introduced by this PR): the old code subtracted conditional items from `items` identically. The test at `merge_coverage.rs:1251` explicitly documents this as intentional "explicit overrides implicit" semantics. Worth revisiting as a separate issue if strict union behavior is desired.
```
</details>

grovedb-query/tests/merge_coverage.rs
```diff
@@ -843,6 +843,505 @@ fn merge_with_overlapping_conditional_items_intersect() {
...
+    let key4_sq = conds
+        .get(&QueryItem::Key(k(4)))
+        .unwrap()
+        .subquery
+        .as_ref()
+        .unwrap();
+    assert!(key4_sq.items.contains(&QueryItem::Key(vec![b'c'])));
+    assert!(!key4_sq.items.contains(&QueryItem::Key(vec![b'a'])));
+    assert!(!key4_sq.items.contains(&QueryItem::Key(vec![b'b'])));
+}
```
thepastaclaw: **🟡 Suggestion: Missing test: key present in all three queries**

The three-query test has pairwise overlaps (Key(2) in A+B, Key(3) in B+C) but no key shared by all three queries. Adding a key in all three would exercise the sequential merge across multiple iterations more thoroughly. Low risk since it works correctly, but strengthens coverage.

<sub>source: ['claude']</sub>

<details>
<summary>🤖 Fix this with AI agents</summary>

```
These findings are from an automated code review. Verify each finding against the current code and only fix it if needed.

In `grovedb-query/tests/merge_coverage.rs`:
- [SUGGESTION] lines 1009-1088: Missing test: key present in all three queries
  The three-query test has pairwise overlaps (Key(2) in A+B, Key(3) in B+C) but no key shared by all three queries. Adding a key in all three would exercise the sequential merge across multiple iterations more thoroughly. Low risk since it works correctly, but strengthens coverage.
```
</details>

thepastaclaw (commented):
## Code Review

The PR correctly fixes the M6 bug (default-vs-default overlap silently dropping one query's default_subquery_branch). The fix is sound, well-tested, and passes clippy cleanly. A pre-existing semantic gap where explicit conditionals override another query's default for the same item is documented by the new test but not introduced by this PR.

*Reviewed commit: 7f8e82f2*

🟡 3 suggestion(s)

<details>
<summary>🤖 Prompt for all review comments with AI agents</summary>

```
These findings are from an automated code review. Verify each finding against the current code and only fix it if needed.

In `grovedb-query/src/merge.rs`:
- [SUGGESTION] lines 492-560: Duplicated overlap-detection logic between merge_multiple and merge_with
  Lines 492–560 (merge_multiple) and 602–650 (merge_with) contain nearly identical overlap-detection and default-promotion logic, differing only in `merged_query` vs `self`. Could be extracted into a shared helper like `fn promote_default_for_overlapping_items(...)`. Reasonable to defer in a bugfix PR, but worth noting for future cleanup.
- [SUGGESTION] lines 496-509: Pre-existing: incoming explicit conditional causes old default to be dropped for overlapping items
  When B has `conditional Key(2)→'x'` and A covers Key(2) via `default='a'`, the merged result gives Key(2) only `'x'` — A's default `'a'` is lost. Under strict union semantics, Key(2) should return BOTH. This is **pre-existing** behavior (not introduced by this PR): the old code subtracted conditional items from `items` identically. The test at `merge_coverage.rs:1251` explicitly documents this as intentional "explicit overrides implicit" semantics. Worth revisiting as a separate issue if strict union behavior is desired.

In `grovedb-query/tests/merge_coverage.rs`:
- [SUGGESTION] lines 1009-1088: Missing test: key present in all three queries
  The three-query test has pairwise overlaps (Key(2) in A+B, Key(3) in B+C) but no key shared by all three queries. Adding a key in all three would exercise the sequential merge across multiple iterations more thoroughly. Low risk since it works correctly, but strengthens coverage.
```
</details>

grovedb-query/src/merge.rs
QuantumExplorer: Addressed in the latest branch: the duplicated default-promotion logic has been extracted into a shared helper used by both `merge_multiple` and `merge_with`.

grovedb-query/src/merge.rs
QuantumExplorer: Leaving this behavior unchanged intentionally for this PR. As noted, the explicit-conditional override behavior is pre-existing and is now documented by the regression test, so changing it should be handled separately if strict union semantics are desired.

grovedb-query/tests/merge_coverage.rs
QuantumExplorer: Addressed in the latest branch: the three-query merge test now includes a key present in all three queries and asserts that the merged subquery contains all three defaults.

QuantumExplorer (commented):
Approved

PR #654, opened 2026-05-09
feat: add Element::NonCounted wrapper to opt-out of count propagation

QuantumExplorer:

GroveDB count-bearing trees (`CountTree`, `ProvableCountTree`, `CountSumTree`, `ProvableCountSumTree`) currently aggregate every child element into the parent's count. There's no way to insert a value that participates in storage and proofs but is exempt from the count — e.g. housekeeping rows, schema markers, or auxiliary indexes inside a count tree.

This adds that escape hatch.

## What was done?

Added a new `Element::NonCounted(Box<Element>)` variant that:
- behaves identically to its inner element for storage, hashing, sum propagation, and internal aggregation,
- contributes **0** to the parent count tree's aggregate count,
- still contributes the inner sum to a parent sum tree (only count is suppressed),
- may **only** be inserted into a count-bearing tree.

### Design choice: single wrapper variant (vs. ~15 explicit variants)

A wrapper means one new discriminant and one dispatch point for the \"is non-counted?\" question. Existing `Element` predicates (\`is_tree\`, \`is_sum_item\`, \`is_any_item\`, \`is_non_empty_tree\`, ...) keep working by looking *through* the wrapper via a new \`underlying()\` helper. Pattern-match sites in cost, proof, query, batch, and insert paths dispatch on the underlying element.

### Cryptographic notes

The wrapper byte is included in the serialized value, so the value hash distinguishes \`NonCounted(X)\` from a bare \`X\`. The parent's \`TreeFeatureType\` for a non-counted child is the inner element's feature type with its count zeroed (via a new \`TreeFeatureType::zero_count()\` helper); no new \`TreeFeatureType\` variants are introduced.

Nested \`NonCounted(NonCounted(...))\` is rejected at construction, serialization, and deserialization to close a stack-overflow vector.

### Scope

Touched 18 files across \`grovedb-element\`, \`grovedb-query\`, \`grovedb-merk\`, and \`grovedb\`:
- enum + discriminant + constructor + helpers + serialization
- \`merk\` insert validation, feature-type dispatch, get/cost paths
- \`grovedb\` batch / insert / query / proof / lib pattern-match sites updated to dispatch on the underlying element

## How Has This Been Tested?

New tests:
- \`grovedb-element\` (10 tests): constructor (wrap, reject-nested, idempotent \`into_non_counted\`); helper passthrough (\`is_*\`, \`count_value_or_default\` returns 0, \`sum_value_or_default\` keeps inner sum, flag accessors); bincode round-trip; reject nested wrapper at deserialize and at serialize.
- \`grovedb-merk\` (5 tests): insert into \`NormalTree\` and \`SumTree\` is rejected; insert into \`CountTree\` succeeds and contributes 0; \`NonCounted(SumItem(10))\` in a \`ProvableCountSumTree\` yields count=0, sum=10; a \`NonCounted(ProvableCountTree)\` whose internal count is 5 contributes 0 to a parent \`CountTree\`.

Full workspace test suite still passes:
- \`cargo test -p grovedb-element\` — 39 passed
- \`cargo test -p grovedb-merk --features minimal,test_utils,full\` — 360 passed
- \`cargo test -p grovedb\` — 1449 passed

\`cargo clippy\` is clean for the changed code (3 pre-existing warnings in unrelated files).

## Breaking Changes

None for existing callers. The change adds a new \`Element\` variant and is additive at the API surface. The on-disk serialization format gains a new bincode discriminant (15) — older code reading data written with this variant would fail to deserialize, so producer/consumer pairs need to be upgraded together. This matches the existing \"ONLY APPEND TO THIS LIST\" comment on the \`Element\` enum.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Support for non-counted elements that preserve storage/hashing and sums while contributing zero to parent counts
  * New helper APIs for transparent access to underlying elements

* **Behavior Changes**
  * Visualization and display now indicate non-counted wrappers
  * Queries, proofs, batch execution, and cost/accounting now treat non-counted elements as transparent while accounting for wrapper overhead

* **Validation & Safety**
  * Reject nested non-counted wrappers during serialization/deserialization
  * Prevent inserting non-counted elements into non-count-bearing trees

* **Tests**
  * Added end-to-end and unit coverage for non-counted semantics and safeguards

### DIFF
--- a/grovedb-commitment-tree/Cargo.toml
+++ b/grovedb-commitment-tree/Cargo.toml
@@ -17,7 +17,7 @@ client = ["shardtree"]
 sqlite = ["client", "rusqlite"]
 
 [dependencies]
-orchard = { git = "https://github.com/dashpay/orchard.git", rev = "41c8f7169f2683c99cf0e0c63e8d25ec12c47a79", features = ["circuit"] }
+orchard = { git = "https://github.com/dashpay/orchard.git", rev = "898258d76aab2822249492aede59a02d49278fff", features = ["circuit"] }
 incrementalmerkletree = "0.8"
 shardtree = { version = "0.6", optional = true }
 rusqlite = { version = "0.38", features = ["bundled"], optional = true }
--- a/grovedb-element/src/element/constructor.rs
+++ b/grovedb-element/src/element/constructor.rs
@@ -383,4 +383,28 @@ impl Element {
     pub fn new_dense_tree(count: u16, height: u8, flags: Option<ElementFlags>) -> Self {
         Element::DenseAppendOnlyFixedSizeTree(count, height, flags)
     }
+
+    /// Wrap an element in `NonCounted` so it contributes 0 to its parent count
+    /// tree's aggregate count when inserted. Sums (if any) still propagate.
+    ///
+    /// Returns `InvalidInput` if `inner` is itself a `NonCounted`. Wrapping
+    /// is not idempotent at the type level — use `into_non_counted` to wrap
+    /// without that risk.
+    pub fn new_non_counted(inner: Element) -> Result<Self, ElementError> {
+        if matches!(inner, Element::NonCounted(_)) {
+            return Err(ElementError::InvalidInput(
+                "NonCounted cannot wrap another NonCounted",
+            ));
+        }
+        Ok(Element::NonCounted(Box::new(inner)))
+    }
+
+    /// Wrap `self` in `NonCounted`. If `self` is already `NonCounted`, returns
+    /// `self` unchanged (idempotent).
+    pub fn into_non_counted(self) -> Self {
+        match self {
+            Element::NonCounted(_) => self,
+            other => Element::NonCounted(Box::new(other)),
+        }
+    }
 }
--- a/grovedb-element/src/element/helpers.rs
+++ b/grovedb-element/src/element/helpers.rs
@@ -11,10 +11,48 @@ use crate::{
 };
 
 impl Element {
+    /// Returns `true` if this element is wrapped in `Element::NonCounted`.
+    /// The wrapper suppresses count propagation to the parent count tree but
+    /// leaves all other behavior (storage, hashing, sum propagation, internal
+    /// aggregation) unchanged.
+    pub fn is_non_counted(&self) -> bool {
+        matches!(self, Element::NonCounted(_))
+    }
+
+    /// Returns the wrapped element if `self` is `NonCounted`, else `self`.
+    /// Use this when you need to inspect the actual element type and don't
+    /// care whether it is wrapped.
+    pub fn underlying(&self) -> &Element {
+        match self {
+            Element::NonCounted(inner) => inner,
+            other => other,
+        }
+    }
+
+    /// Mutable variant of [`underlying`].
+    pub fn underlying_mut(&mut self) -> &mut Element {
+        match self {
+            Element::NonCounted(inner) => inner,
+            other => other,
+        }
+    }
+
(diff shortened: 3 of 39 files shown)


QuantumExplorer:
Codex review findings:

1. **[P1] Reject nested `NonCounted` before decode**  
   `grovedb-element/src/element/serialize.rs:52`  
   The nested-wrapper guard runs only after `bincode::decode_from_slice` has decoded the recursive `Box<Element>`. A malicious value starting with repeated `NonCounted` discriminants still recurses before this check, so the stack-exhaustion vector is not actually closed. Pre-check the raw bytes for a nested wrapper, or call the serialized-type parser, before bincode decoding.

2. **[P1] Batch propagation drops the wrapper**  
   `grovedb/src/batch/mod.rs:2625`  
   This unwraps `NonCounted(Tree)` to choose the internal `InsertTreeWithRootHash`/`InsertNonMerkTree` op, but the internal op stores no information that the original element was wrapped. When a batch inserts a `NonCounted` tree and also writes under it, execution later reconstructs a bare tree, so the parent count includes it and the on-disk element loses the wrapper.

3. **[P1] Existing wrapped trees cannot propagate**  
   `merk/src/element/reconstruct.rs:27`  
   `reconstruct_with_root_key` has no `Element::NonCounted` arm. Existing `NonCounted` tree elements can be opened because the tree-type helpers unwrap them, but any later subtree mutation that propagates a new root key reaches this helper and returns `None`, causing `update_tree_item_preserve_flag` to fail instead of re-wrapping the reconstructed inner tree.

4. **[P2] Batch insert bypasses the parent-type guard**  
   `grovedb/src/batch/mod.rs:1867`  
   Direct Merk insertion rejects `NonCounted` children outside count-bearing parents, but this batch path never checks `element.is_non_counted() && !in_tree_type.is_count_bearing()` before adding the put op. User-facing batch inserts can therefore persist `NonCounted` elements into normal or sum trees, violating the new invariant.

5. **[P2] NonCounted references are not followed**  
   `grovedb/src/operations/get/mod.rs:83`  
   The get path only treats a bare `Element::Reference` as a reference. A valid `NonCounted(Reference)` inserted into a count-bearing tree falls through and is returned as a value, and the loop below has the same issue for reference chains. This contradicts the wrapper-transparency contract and differs from the query/proof paths that unwrap before dispatch.

Verification performed:

- `cargo test -p grovedb-element non_counted --all-features` passed.
- `cargo test -p grovedb-merk non_counted --all-features` passed.
- `cargo test -p grovedb non_counted --all-features` compiled and ran 0 matching tests, with existing warnings.

QuantumExplorer:
Codex follow-up review:

I re-fetched the updated PR at `10d38210` and rechecked the five prior findings. The fixes address my original concerns:

- nested `NonCounted` is rejected before bincode decode
- batch propagation preserves and re-applies the wrapper
- `reconstruct_with_root_key` re-wraps `NonCounted` trees
- batch insert now enforces count-bearing parents
- `get` / `follow_reference` handle `NonCounted(Reference)`

Focused verification run:

- `cargo test -p grovedb-element deserialize_rejects_long_nested_wrapper_chain_without_recursion --all-features`
- `cargo test -p grovedb-merk reconstruct_preserves_non_counted_wrapper --all-features`
- `cargo test -p grovedb non_counted --all-features`

All passed. The `grovedb` run still emits existing test warnings, but I do not see remaining blockers from my previous review.

QuantumExplorer (commented):
Reviewed

QuantumExplorer:
Self Reviewed

PR #656, opened 2026-05-09
feat: add AggregateCountOnRange query for provable count trees

QuantumExplorer:

Counting a sub-range of keys in a `ProvableCountTree` was previously forced through a regular query, paying `O(result-size)` proof bytes for an answer that's already cryptographically committed at every internal node via `node_hash_with_count`. This PR adds a dedicated count-only proof shape with `O(log n)` size regardless of how many keys fall in the range.

## What was done?

A new `QueryItem` variant `AggregateCountOnRange(Box<QueryItem>)` that wraps an inner range and asks for a single `u64` count + cryptographic proof, instead of returning the elements themselves. Restricted to `ProvableCountTree` and `ProvableCountSumTree` (plus their `NonCounted*` wrapper variants); rejected for any other tree type at proof time.

**Cryptographic core:**
- New proof node `Node::HashWithCount(kv_hash, left_child_hash, right_child_hash, count)` which is **self-verifying**: the verifier recomputes `node_hash_with_count(...)` from the four committed fields. A forged count diverges from the parent's expected hash and the Merkle-root chain check fails — so the count is bound by the proof, not trusted on faith. (An earlier draft used `(node_hash, count)` only, which would have been forge-able; that version was rejected during design review.)

**Proof generation (merk):**
- `Merk::prove_aggregate_count_on_range` walks the AVL tree, classifying each subtree as `Disjoint` / `Contained` / `Boundary` using inherited exclusive key-bound windows. Disjoint short-circuits at the link level (no I/O); Contained walks one level for its kv_hash + grandchild hashes; Boundary recurses. Emits `Hash` / `HashWithCount` / `KVDigestCount` accordingly.

**Proof verification (merk):**
- `verify_aggregate_count_on_range_proof` replays Ops via the new `execute_with_options(.., verify_avl_balance=false, ..)`. The AVL-balance check is bypassed only here because count proofs intentionally collapse one side of the tree into a single op while descending the other — the resulting reconstructed tree is unbalanced by design. All existing callers keep the balance check via the unchanged `execute()` wrapper.
- Allowlists `Hash` / `HashWithCount` / `KVDigestCount`; anything else is `InvalidProofError`.

**GroveDB integration:**
- One `prove_query` entry — short-circuits inside both `prove_subqueries` (v0) and `prove_subqueries_v1` when the leaf-merk's items contain `AggregateCountOnRange`. Envelope structure is unchanged so existing serialization works.
- New `GroveDb::verify_aggregate_count_query(proof, path_query, version) -> Result<(CryptoHash, u64), Error>` walks the multi-layer envelope, verifies single-key existence proofs at each non-leaf layer, delegates to the merk count verifier at the leaf, and enforces `combine_hash(H(value), lower_root) == parent_proof_hash` for the chain.

**Validation:**
Enforced at `Query::validate_aggregate_count_on_range` / `SizedQuery::validate_aggregate_count_on_range` / `PathQuery::validate_aggregate_count_on_range`:
1. Must be the only item in the query
2. Inner item not `Key` (use `has_raw` / `get_raw`)
3. Inner item not `RangeFull` (read the parent `Element::ProvableCountTree` bytes)
4. Inner item not nested `AggregateCountOnRange`
5. No `default_subquery_branch`
6. No `conditional_subquery_branches`
7. No `SizedQuery::limit` / `offset`
8. `left_to_right` ignored

## How Has This Been Tested?

**Unit (10 tests):** classification of every Disjoint / Contained / Boundary case across all range-bound shapes.

**Merk integration (13 tests):** every allowed range variant (`Range`, `RangeInclusive`, `RangeFrom`, `RangeAfter`, `RangeTo`, `RangeToInclusive`, `RangeAfterTo`, `RangeAfterToInclusive`) + range-below-all-keys + range-above-all-keys + empty merk + tree-type rejection + a **count-forgery test** that tampers with the count in a `HashWithCount` op and asserts the reconstructed root diverges from the expected one.

**GroveDB integration (11 tests, [`grovedb/src/tests/aggregate_count_query_tests.rs`](grovedb/src/tests/aggregate_count_query_tests.rs)):** end-to-end `prove → encode → decode → verify` on `ProvableCountTree` and `ProvableCountSumTree` at multi-layer paths, plus validation rejections (Key inner, RangeFull inner, NormalTree target) and a GroveDB-level forgery test that flips a count byte in the encoded proof and asserts the verifier rejects it via the layer-chain check.

**Workspace state:** `cargo build --workspace`, `cargo clippy --workspace --all-targets`, and `cargo test --workspace` all pass with no regressions (1465+ grovedb lib tests + 387+ merk lib tests). `cargo fmt` clean.

## Documentation

New chapter at [`docs/book/src/aggregate-count-queries.md`](docs/book/src/aggregate-count-queries.md) covering the contract, allowed range variants, validation rules, and per-case proof shapes (open and closed ranges, with mermaid diagrams using a 15-key example tree). Cross-linked from the query-system chapter.

## Breaking Changes

None for callers — the new variant is additive and existing `prove_query` / `verify_query` paths are untouched. Sites that pattern-match exhaustively on `QueryItem` or `Node` were updated (one `&'static str` rejection arm each in dense-tree, bulk-append-tree, and the grovedb proof generator/verifier — none of those tree types support count queries).

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added aggregate count queries: count elements in bounded key ranges with verifiable cryptographic proofs and accompanying prove/verify support.

* **Documentation**
  * New "Aggregate Count Queries" doc and SUMMARY entry; added client-side JS fix to ensure Mermaid diagrams render correctly in the HTML docs.

* **Tests**
  * Extensive unit and end-to-end tests covering aggregate count proofs across tree types.

### DIFF
--- a/grovedb/src/operations/proof/generate.rs
+++ b/grovedb/src/operations/proof/generate.rs
@@ -2165,7 +2165,7 @@ impl GroveDb {
 mod tests {
     use grovedb_merk::proofs::query::QueryItem;
 
-    use crate::GroveDb;
+    use crate::{Error, GroveDb};
 
     /// Helper: encode a u16 as big-endian bytes.
     fn be_u16(v: u16) -> Vec<u8> {
@@ -2302,5 +2302,60 @@ mod tests {
             start,
             end
         );
+    }
+
+    // -----------------------------------------------------------------------
+    // AggregateCountOnRange rejection on non-provable-count tree types.
+    //
+    // `AggregateCountOnRange` is only meaningful against `ProvableCountTree`
+    // and `ProvableCountSumTree` (their nodes commit a count via
+    // `node_hash_with_count`). Dense, MMR, and BulkAppendTree have no such
+    // commitment, so the index-resolution helpers must reject the variant
+    // outright rather than silently fall through.
+    // -----------------------------------------------------------------------
+
+    #[test]
+    fn dense_tree_rejects_aggregate_count_on_range() {
+        let inner = QueryItem::RangeInclusive(be_u16(0)..=be_u16(5));
+        let items = vec![QueryItem::AggregateCountOnRange(Box::new(inner))];
+        let err = GroveDb::query_items_to_positions(&items, 100)
+            .expect_err("dense tree must reject AggregateCountOnRange");
+        match err {
+            Error::InvalidInput(msg) => assert!(
+                msg.contains("dense fixed-size") || msg.contains("provable count"),
+                "unexpected message: {msg}"
+            ),
+            other => panic!("expected InvalidInput, got {:?}", other),
+        }
+    }
+
+    #[test]
+    fn mmr_tree_rejects_aggregate_count_on_range() {
+        let inner = QueryItem::RangeInclusive(be_u64(0)..=be_u64(5));
+        let items = vec![QueryItem::AggregateCountOnRange(Box::new(inner))];
+        let err = GroveDb::query_items_to_leaf_indices(&items, 7)
+            .expect_err("MMR must reject AggregateCountOnRange");
+        match err {
+            Error::InvalidInput(msg) => assert!(
+                msg.contains("MMR") || msg.contains("provable count"),
+                "unexpected message: {msg}"
+            ),
+            other => panic!("expected InvalidInput, got {:?}", other),
+        }
+    }
+
+    #[test]
+    fn bulk_append_tree_rejects_aggregate_count_on_range() {
+        let inner = QueryItem::RangeInclusive(be_u64(0)..=be_u64(5));
+        let items = vec![QueryItem::AggregateCountOnRange(Box::new(inner))];
+        let err = GroveDb::query_items_to_range(&items, 100)
+            .expect_err("BulkAppendTree must reject AggregateCountOnRange");
+        match err {
+            Error::InvalidInput(msg) => assert!(
+                msg.contains("BulkAppendTree") || msg.contains("provable count"),
+                "unexpected message: {msg}"
+            ),
+            other => panic!("expected InvalidInput, got {:?}", other),
+        }
     }
 }
--- a/grovedb/src/operations/proof/generate.rs
+++ b/grovedb/src/operations/proof/generate.rs
@@ -269,6 +269,29 @@ impl GroveDb {
             *overall_limit
         };
 
+        // Aggregate-count short-circuit: when the query items at this level
+        // are a single AggregateCountOnRange, we skip the regular merk proof
+        // path entirely and emit a count-only merk proof. Count queries are
+        // leaf-only — `lower_layers` stays empty.
+        if let Some(inner_range) = query.items.first().and_then(|qi| match qi {
+            QueryItem::AggregateCountOnRange(inner) => Some(inner.as_ref()),
+            _ => None,
+        }) {
+            let (count_ops, _count) = cost_return_on_error!(
+                &mut cost,
+                subtree
+                    .prove_aggregate_count_on_range(inner_range, grove_version)
+                    .map_err(Error::MerkError)
+            );
+            let mut serialized = Vec::with_capacity(128);
+            encode_into(count_ops.iter(), &mut serialized);
+            return Ok(MerkOnlyLayerProof {
+                merk_proof: serialized,
+                lower_layers: BTreeMap::new(),
+            })
+            .wrap_with_cost(cost);
+        }
+
         let mut merk_proof = cost_return_on_error!(
             &mut cost,
             self.generate_merk_proof(
@@ -1011,6 +1034,29 @@ impl GroveDb {
         } else {
             *overall_limit
         };
+
+        // Aggregate-count short-circuit (v1 path). Identical logic to v0:
+        // a single AggregateCountOnRange item routes to the count-proof
+        // generator; lower_layers is empty. The count-proof bytes are wrapped
+        // in `ProofBytes::Merk` since they share the merk Op stream encoding.
+        if let Some(inner_range) = query.items.first().and_then(|qi| match qi {
+            QueryItem::AggregateCountOnRange(inner) => Some(inner.as_ref()),
+            _ => None,
+        }) {
+            let (count_ops, _count) = cost_return_on_error!(
+                &mut cost,
+                subtree
+                    .prove_aggregate_count_on_range(inner_range, grove_version)
+                    .map_err(Error::MerkError)
+            );
+            let mut serialized = Vec::with_capacity(128);
+            encode_into(count_ops.iter(), &mut serialized);
+            return Ok(LayerProof {
+                merk_proof: ProofBytes::Merk(serialized),
+                lower_layers: BTreeMap::new(),
+            })
+            .wrap_with_cost(cost);
+        }
 
         let mut merk_proof = cost_return_on_error!(
             &mut cost,
@@ -1862,6 +1908,12 @@ impl GroveDb {
                         }
                     }
                 }
+                QueryItem::AggregateCountOnRange(_) => {
+                    return Err(Error::InvalidInput(
+                        "AggregateCountOnRange is only supported on provable count trees, \
+                         not on dense fixed-size merkle trees",
+                    ));
+                }
             }
         }
 
@@ -1980,6 +2032,12 @@ impl GroveDb {
                         }
                     }
                 }
+                QueryItem::AggregateCountOnRange(_) => {
+                    return Err(Error::InvalidInput(
+                        "AggregateCountOnRange is only supported on provable count trees, \
+                         not on MMR trees",
+                    ));
+                }
             }
         }
 
@@ -2047,6 +2105,12 @@ impl GroveDb {
                     let e = Self::decode_be_u64(range.end())?;
                     min_start = min_start.min(s.saturating_add(1));
                     max_end = max_end.max(e.saturating_add(1));
+                }
+                QueryItem::AggregateCountOnRange(_) => {
+                    return Err(Error::InvalidInput(
+                        "AggregateCountOnRange is only supported on provable count trees, \
+                         not on BulkAppendTree",
+                    ));
                 }
             }
         }
--- a/grovedb/src/operations/proof/generate.rs
+++ b/grovedb/src/operations/proof/generate.rs
@@ -269,18 +269,26 @@ impl GroveDb {
             *overall_limit
         };
 
-        // Aggregate-count short-circuit: when the query items at this level
-        // are a single AggregateCountOnRange, we skip the regular merk proof
-        // path entirely and emit a count-only merk proof. Count queries are
-        // leaf-only — `lower_layers` stays empty.
-        if let Some(inner_range) = query.items.first().and_then(|qi| match qi {
-            QueryItem::AggregateCountOnRange(inner) => Some(inner.as_ref()),
-            _ => None,
-        }) {
+        // Aggregate-count short-circuit: if any item at this level is an
+        // `AggregateCountOnRange`, the surrounding `PathQuery` must validate
+        // as a well-formed aggregate-count query. We do **not** route on a
+        // partial match (e.g. a query with extra items, subqueries, or an
+        // illegal inner) — those would silently produce a count proof for
+        // the wrong shape. Instead we run the same validation the verifier
+        // runs and let it surface the precise error.
+        if query
+            .items
+            .iter()
+            .any(QueryItem::is_aggregate_count_on_range)
+        {
+            let inner_range = cost_return_on_error_no_add!(
+                cost,
+                path_query.validate_aggregate_count_on_range().cloned()
+            );
             let (count_ops, _count) = cost_return_on_error!(
                 &mut cost,
                 subtree
-                    .prove_aggregate_count_on_range(inner_range, grove_version)
+                    .prove_aggregate_count_on_range(&inner_range, grove_version)
                     .map_err(Error::MerkError)
             );
             let mut serialized = Vec::with_capacity(128);
@@ -1035,18 +1043,24 @@ impl GroveDb {
             *overall_limit
         };
 
-        // Aggregate-count short-circuit (v1 path). Identical logic to v0:
-        // a single AggregateCountOnRange item routes to the count-proof
-        // generator; lower_layers is empty. The count-proof bytes are wrapped
-        // in `ProofBytes::Merk` since they share the merk Op stream encoding.
-        if let Some(inner_range) = query.items.first().and_then(|qi| match qi {
-            QueryItem::AggregateCountOnRange(inner) => Some(inner.as_ref()),
-            _ => None,
-        }) {
+        // Aggregate-count short-circuit (v1 path). Same validation contract
+        // as v0: any AggregateCountOnRange at this level requires the
+        // surrounding PathQuery to validate as a well-formed aggregate-count
+        // query. The count-proof bytes are wrapped in `ProofBytes::Merk`
+        // since they share the merk Op stream encoding.
+        if query
+            .items
+            .iter()
+            .any(QueryItem::is_aggregate_count_on_range)
+        {
+            let inner_range = cost_return_on_error_no_add!(
+                cost,
+                path_query.validate_aggregate_count_on_range().cloned()
+            );
             let (count_ops, _count) = cost_return_on_error!(
                 &mut cost,
                 subtree
-                    .prove_aggregate_count_on_range(inner_range, grove_version)
+                    .prove_aggregate_count_on_range(&inner_range, grove_version)
                     .map_err(Error::MerkError)
             );
             let mut serialized = Vec::with_capacity(128);
--- a/grovedb/src/operations/proof/generate.rs
+++ b/grovedb/src/operations/proof/generate.rs
@@ -109,6 +109,22 @@ impl GroveDb {
         prove_options: Option<ProveOptions>,
         grove_version: &GroveVersion,
     ) -> CostResult<GroveDBProof, Error> {
+        // Aggregate-count gate: validate at entry so malformed ACOR
+        // queries (invalid inner range, ACOR-hidden-in-subquery, etc.) are
+        // rejected up front instead of being skipped when the recursive
+        // prover never reaches the ACOR-bearing leaf — for example because
+        // the path doesn't exist. Without this gate, `prove_query` would
+        // happily return a regular path/absence proof for an invalid
+        // aggregate-count request.
+        if path_query
+            .query
+            .query
+            .has_aggregate_count_on_range_anywhere()
+            && let Err(e) = path_query.validate_aggregate_count_on_range()
+        {
+            return Err(e).wrap_with_cost(OperationCost::default());
+        }
+
         match grove_version
             .grovedb_versions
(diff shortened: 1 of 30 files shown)


QuantumExplorer:
Review findings:

1. **[P0] Count verifier accepts forged proof shapes**  
   `merk/src/proofs/query/aggregate_count.rs:475-488`

   The verifier only allowlists node types; it never proves that a `Hash` is actually for a disjoint subtree or that `HashWithCount` is a leaf in a contained subtree. A prover can send `Push(Hash(expected_root))` for a non-empty tree and get the trusted root back with count 0 for any range. They can also replace in-range `HashWithCount` subtrees with `Hash` to undercount, or attach extra counted children under keyless `Hash`/`HashWithCount` nodes to overcount while preserving the root hash. The verifier needs to replay the aggregate-count traversal against `inner_range` and inherited subtree bounds, not just execute arbitrary allowed ops.

2. **[P2] Prover bypasses aggregate-count validation**  
   `grovedb/src/operations/proof/generate.rs:276-284`

   The short-circuit checks only whether the first item is `AggregateCountOnRange`. That lets invalid queries through proof generation: extra items are ignored, subquery branches are ignored, limits can be ignored, and invalid inner items such as `Key` or `RangeFull` still produce a count proof. This should use `PathQuery::validate_aggregate_count_on_range()` before routing to the count prover, or at least enforce the same single-item/no-subquery/no-pagination/inner-range rules here.

3. **[P2] Recursive `QueryItem` decoding has no depth limit**  
   `grovedb-query/src/query_item/mod.rs:357-359`

   `AggregateCountOnRange` is invalid when nested, but decoding recursively accepts another `QueryItem` before validation runs. A small malicious payload made of repeated variant-10 bytes can recurse deeply enough to overflow the stack during bincode or borrow decoding. Either reject an inner aggregate variant during decode or add a bounded `decode_with_depth` path like the existing Query/Subquery decoding guards.

QuantumExplorer:
Suggested fix direction for the findings above:

1. **Fix the aggregate-count verifier by replaying the proof shape, not just allowlisting node types.**

   I would make `verify_aggregate_count_on_range_proof` two-phase:

   - Decode/execute the proof into a `ProofTree` as today, but do not accumulate `count` in the `visit_node` closure.
   - Walk the reconstructed proof tree with the same inherited exclusive subtree bounds used by the prover: `(subtree_lo_excl, subtree_hi_excl)`.
   - At each proof-tree node, call `classify_subtree(bounds, inner_range)` and require the node shape to match the classification:
     - `Disjoint` => node must be a leaf `Node::Hash(_)`, contributes `0`.
     - `Contained` => node must be a leaf `Node::HashWithCount(.., count)`, contributes `count`.
     - `Boundary` => node must be `Node::KVDigestCount(key, ..)`, the key must be strictly inside the inherited bounds, then recurse left with `(lo, key)` and right with `(key, hi)`, plus `1` if `inner_range.contains(key)`.
   - Reject children under `Hash` or `HashWithCount`; their hash computation ignores attached proof-tree children, so accepting them lets a proof overcount without changing the root hash.
   - Use checked addition rather than `saturating_add`; overflow should be an invalid proof.
   - Optionally reject inverted structural ops for this proof flavor, or at least rely on the shape walk to validate the final left/right tree topology.

   In pseudocode, the important part is roughly:

   ```rust
   fn verify_count_shape(
       tree: &ProofTree,
       range: &QueryItem,
       lo: Option<&[u8]>,
       hi: Option<&[u8]>,
   ) -> Result<u64, Error> {
       match classify_subtree(lo, hi, range) {
           Disjoint => match &tree.node {
               Node::Hash(_) if tree.left.is_none() && tree.right.is_none() => Ok(0),
               _ => Err(invalid("expected disjoint Hash leaf")),
           },
           Contained => match &tree.node {
               Node::HashWithCount(_, _, _, count)
                   if tree.left.is_none() && tree.right.is_none() => Ok(*count),
               _ => Err(invalid("expected contained HashWithCount leaf")),
           },
           Boundary => match &tree.node {
               Node::KVDigestCount(key, _, _) => {
                   ensure_key_inside_bounds(key, lo, hi)?;
                   let left = tree.left.as_ref()
                       .map_or(Ok(0), |c| verify_count_shape(&c.tree, range, lo, Some(key)))?;
                   let right = tree.right.as_ref()
                       .map_or(Ok(0), |c| verify_count_shape(&c.tree, range, Some(key), hi))?;
                   checked_sum(left, right, u64::from(range.contains(key)))
               }
               _ => Err(invalid("expected boundary KVDigestCount")),
           },
       }
   }
   ```

   Then `verify_aggregate_count_on_range_proof` returns `verify_count_shape(&tree, inner_range, None, None)` alongside `tree.hash()`. That rejects the single-`Hash(expected_root)` undercount and the keyless-node child overcount cases.

2. **Fix proof generation by centralizing validation before the short-circuit.**

   The `prove_subqueries` / `prove_subqueries_v1` short-circuit should not look only at `query.items.first()`. I would make it trigger only when the current level contains an aggregate item anywhere, then call `path_query.validate_aggregate_count_on_range()` and use the returned inner range. That makes invalid shapes fail instead of being partially interpreted.

   Concretely, change the short-circuit from “first item is aggregate” to “any aggregate item means this path query must validate as the exact aggregate-count form”:

   ```rust
   if query.items.iter().any(QueryItem::is_aggregate_count_on_range) {
       let inner_range = path_query.validate_aggregate_count_on_range()?;
       // prove aggregate count with inner_range
   }
   ```

   This catches extra query items, nested/subquery aggregate usage, limits/offsets, and invalid inner `Key` / `RangeFull` before the merk prover is called. The same guard should be applied in both v0 and v1 proof-generation paths.

3. **Fix recursive `QueryItem` decoding with a bounded decode helper.**

   I would add `QueryItem::decode_with_depth(decoder, depth)` and `borrow_decode_with_depth(decoder, depth)` with a small max depth, similar to the existing `Query` / `SubqueryBranch` decode-depth guard. The public `Decode` impl calls depth `0`; variant `10` decodes the inner item at `depth + 1` and errors once the max is exceeded.

   Since nested `AggregateCountOnRange` is invalid anyway, it is also reasonable to reject an inner `AggregateCountOnRange` immediately after decoding the inner item. The depth guard is still useful so maliciously deep variant-10 payloads fail before stack exhaustion.

QuantumExplorer:
This is Claude. Pushed [`5989cc49`](https://github.com/dashpay/grovedb/pull/656/commits/5989cc497) addressing all three findings:

**[P0] Count verifier shape walk** — `verify_aggregate_count_on_range_proof` is now two-phase. Phase 1 reconstructs the proof tree via `execute_with_options` (no counting in the visit_node closure). Phase 2 walks the reconstructed tree with the same inherited exclusive subtree-key bounds the prover used and binds each position's node type to the classification:

| Classification | Required node | Contribution |
|---|---|---|
| Disjoint  | leaf `Hash`           | 0 |
| Contained | leaf `HashWithCount`  | committed `count` |
| Boundary  | `KVDigestCount` with key strictly inside bounds | recurse + (1 if `range.contains(key)`) |

Counts use `checked_add`; overflow is treated as proof corruption. Three new tests in [`merk/src/proofs/query/aggregate_count.rs`](https://github.com/dashpay/grovedb/blob/feat/aggregate-count-query-item/merk/src/proofs/query/aggregate_count.rs) prove each attack you described is rejected:
- `shape_walk_rejects_single_hash_undercount` — `Push(Hash(expected_root))` for a non-empty tree.
- `shape_walk_rejects_hash_swap_for_contained_subtree` — replacing an in-range `HashWithCount` with a `Hash` carrying the same `node_hash`.
- `shape_walk_rejects_keyless_node_with_attached_children` — splicing a fake `+1` child under a `HashWithCount` (whose `hash()` ignores attached children).

**[P2] Prover validation** — both `prove_subqueries` (v0) and `prove_subqueries_v1` short-circuits now fire on \`query.items.iter().any(QueryItem::is_aggregate_count_on_range)\` and immediately call \`PathQuery::validate_aggregate_count_on_range()\` to extract the inner range. Extra items, subquery branches, limits/offsets, and invalid inner items are now rejected before the count prover runs.

**[P2] QueryItem decode depth + nested rejection** — added `QueryItem::decode_with_depth` and `borrow_decode_with_depth` mirroring the existing Query/SubqueryBranch depth-guard pattern. `MAX_QUERY_ITEM_DECODE_DEPTH = 4` (one legal level + headroom). Defense-in-depth: an inner `AggregateCountOnRange` is also rejected immediately at decode time. Three new tests cover the legal one-level shape, the nested rejection, and a malicious deep payload.

CI:
- `cargo clippy --workspace --all-features -D warnings`: clean
- `cargo fmt --all`: clean
- `cargo test --workspace`: all green (1475+ grovedb lib + 27 merk aggregate_count + 3 query_item decode)

QuantumExplorer:
Re-reviewed the latest head (`5989cc49`). The three earlier findings are materially addressed:

- The aggregate-count verifier now reconstructs the proof tree first, then shape-walks it against inherited subtree bounds. The previous single-`Hash(expected_root)`, `HashWithCount -> Hash`, and keyless-node child-smuggling attacks are covered by tests and rejected.
- The v0/v1 proof-generation short-circuits now call `PathQuery::validate_aggregate_count_on_range()` instead of routing on `items.first()`.
- The bincode `Decode` / `BorrowDecode` paths for `QueryItem` now have bounded recursive decode helpers and reject nested `AggregateCountOnRange`.

One remaining item I would still fix before merging:

**[P2] Serde `QueryItem` deserialization still recursively accepts nested `AggregateCountOnRange`**  
`grovedb-query/src/query_item/mod.rs:225-227`

The bincode decode path is now depth-bounded, but the optional `serde` path still does `let inner: QueryItem = variant_access.newtype_variant()?` and then wraps it. With the `serde` feature enabled, a nested `AggregateCountOnRange` payload still recurses through `QueryItem::deserialize` before any validation runs, so serde-backed clients keep the invalid-nesting/DoS class that the bincode fix closed. Since nested aggregate count is invalid anyway, the serde variant should deserialize the inner value through a non-recursive helper that accepts only non-aggregate `QueryItem` variants, or otherwise enforce a depth limit before recursing. Please add a serde-feature test for nested `AggregateCountOnRange` rejection too.

Verification I ran locally:

- `cargo test -p grovedb-merk aggregate_count --features minimal`
- `cargo test -p grovedb aggregate_count --features minimal`
- `cargo test -p grovedb-query aggregate_count --features serde`
- `cargo test -p grovedb-query query_item::test --features serde`

QuantumExplorer:
This is Claude. Pushed [`fd272c7b`](https://github.com/dashpay/grovedb/pull/656/commits/fd272c7b) addressing the remaining serde-feature finding.

**Fix**: when `Field::AggregateCountOnRange` dispatches in the `QueryItem` deserialize visitor, the inner item is now deserialized via a new `NonAggregateInner(QueryItem)` newtype wrapper. Its `Deserialize` impl mirrors the QueryItem variant set but **omits** `AggregateCountOnRange` from its `Field` enum entirely — so a nested-aggregate payload is rejected by serde's enum dispatcher immediately, with no recursion through `QueryItem::deserialize`. This is the serde counterpart to the bincode `decode_with_depth` + nested-rejection in 5989cc49.

**Tests** (gated on `feature = "serde"`, using `serde_test`'s token-level driver):
- `serde_decode_rejects_nested_aggregate_count_on_range` — token stream for `AggregateCountOnRange(AggregateCountOnRange(...))` produces an \`unknown field \`aggregate_count_on_range\`\` error from `NonAggregateInner`'s field dispatcher.
- `serde_decode_accepts_valid_one_level_aggregate_count_on_range` — token stream for the only legal shape (one wrap around a non-aggregate range) deserializes successfully.

Added `serde_test = \"1.0\"` as a dev-dependency. Used token-level testing rather than a textual format because while writing this I noticed a pre-existing PascalCase/snake_case mismatch between the existing `Serialize` impl (variant tags emitted as PascalCase) and the existing `Field` enum (\`#[serde(field_identifier, rename_all = \"snake_case\")]\` — expects snake_case). That mismatch breaks JSON round-trip but is invisible to formats that don't carry variant names textually. Happy to fix it as a small follow-up commit on this PR if you'd like — it's unrelated to this work but worth surfacing.

Workspace state: \`cargo clippy --workspace --all-features -D warnings\` clean, \`cargo test --workspace\` green.

QuantumExplorer:
This is Claude. Pushed [\`02e75c15\`](https://github.com/dashpay/grovedb/pull/656/commits/02e75c15) addressing the CodeRabbit findings.

**[Major #1] Detect ACOR presence even on malformed queries** — Fixed. \`Query::aggregate_count_on_range\` now scans the entire \`items\` vec via \`iter().find\`, so queries with extra items or with the ACOR not at index 0 are still detected. Shape enforcement remains in \`validate_aggregate_count_on_range\`. Updated the \`aggregate_count_on_range_helper_*\` test to cover ACOR-not-at-index-0 and ACOR-with-extra-item cases.

**[Major #3] Reject HashWithCount in the plain query verifier** — Fixed. Real attack: \`Tree::hash()\` for \`HashWithCount\` recomputes from its embedded \`(kv_hash, l, r, count)\` and ignores any reconstructed children, so a malicious prover could include a \`HashWithCount\` in a regular query proof, attach fake KV pushes under it (which \`execute_node\` would credit as query results), and the parent's hash chain would still verify. Now \`Node::HashWithCount(..)\` has its own match arm in the plain verifier that fails fast with \`\"HashWithCount node is only valid in aggregate-count proofs; encountered in regular query verification\"\`. New test \`regular_query_verifier_rejects_hash_with_count_node\` exercises this end-to-end.

**[Major #2] is_provable_count_bearing missing NonCounted variants** — Skipped, not a bug. CodeRabbit suggested normalizing \`tree_type\` for \`NonCounted*\` wrapper variants, but there's no such \`TreeType\` variant — \`NonCounted\` is an \`ElementType\` wrapper that the merk-open path strips via \`Element::NonCounted(inner) => inner.root_key_and_tree_type_owned()\` ([merk/src/element/tree_type.rs:63](https://github.com/dashpay/grovedb/blob/feat/aggregate-count-query-item/merk/src/element/tree_type.rs#L63)). The merk's \`tree_type\` for any NonCounted-wrapped provable count tree element is already the bare \`TreeType::ProvableCountTree\` / \`ProvableCountSumTree\`, so the existing gate is correct. (The book wording \"NonCounted* wrapper variants\" was the source of confusion; it referred to ElementType wrappers, not TreeType.)

\`cargo clippy --workspace --all-features -D warnings\` clean; full workspace test suite green including the two new tests.

QuantumExplorer (commented):
Reviewed

QuantumExplorer:
This is Claude. Addressed both CodeRabbit comments in abf86d0d:

**1. Excluded-subtree examples (`docs/book/src/aggregate-count-queries.md`)**
Updated the per-case diagrams (Case 1 RangeFrom + RangeAfter, Case 2 RangeInclusive) so out-of-range leaves are shown as `HashWithCount` instead of plain `Hash`. The role table and verifier-walk section were already consistent — only the diagrams contradicted the security model. Also added the outside subtrees (with `+0` contribution) to each verifier total table for clarity, and updated the gray legend entry to reflect the new shape.

**2. Forgery test (`grovedb/src/tests/aggregate_count_query_tests.rs`)**
Replaced the byte-scan with a proper parse:
- Decode the `GroveDBProof` envelope via bincode (same config as the verifier).
- Walk down `lower_layers` along the path keys to the leaf merk proof bytes.
- Use `grovedb_merk::proofs::Decoder` to parse the merk proof at true op boundaries, find the first `Op::Push(Node::HashWithCount(_, _, _, count))` (or `PushInverted`), increment the count, re-encode with `encode_into`.
- Re-encode the envelope.

This guarantees we mutate an actual count varint and not a `0x1e` byte that happens to live inside an embedded 32-byte hash. The test name promises that a forged count is rejected; the new flow actually proves that.

All 19 grovedb-level + 30 merk-level aggregate-count tests pass. `cargo fmt --check` and the `typos`/`cargo fmt` pre-commit hooks all clean.

grovedb/src/operations/proof/generate.rs
```diff
@@ -269,6 +269,37 @@ impl GroveDb {
...
+        // the wrong shape. Instead we run the same validation the verifier
+        // runs and let it surface the precise error.
+        if query
+            .items
+            .iter()
+            .any(QueryItem::is_aggregate_count_on_range)
+        {
+            let inner_range = cost_return_on_error_no_add!(
+                cost,
+                path_query.validate_aggregate_count_on_range().cloned()
```
QuantumExplorer, line 302: [P2] Aggregate-count validation still runs too late

This validation only runs after traversal reaches the target query level. If the path is missing, the recursive prover never sees the `AggregateCountOnRange` item, so malformed aggregate-count queries can still return a regular path/absence proof instead of being rejected. I reproduced this with `PathQuery::new_aggregate_count_on_range([TEST_LEAF, "missing"], QueryItem::Key(...))`: `prove_query` returned `Ok(Some(proof))` even though the inner `Key` is invalid for aggregate count.

I would fix this by validating aggregate-count path queries at the `prove_query` / v1 entry point before `prove_subqueries` starts, for example: `if path_query.has_aggregate_count_on_range() { path_query.validate_aggregate_count_on_range()?; }`. If aggregate-count items can be constructed inside nested subquery branches, make the detector recursive and reject those shapes too, since aggregate-count is terminal and should not be hidden under a regular subquery path.

QuantumExplorer:
This is Claude. Real bug — thanks for the catch and the precise reproducer. Fixed in c13b5f4d.

**Reproduction confirmed first.** Added `aggregate_count_with_missing_path_and_invalid_inner_is_rejected_at_entry`, which builds the exact shape you described (`PathQuery::new_aggregate_count_on_range([TEST_LEAF, "missing"], QueryItem::Key(...))`). Before the fix, `prove_query` returned `Ok(86)` (an 86-byte proof for the malformed query). Test went red, fix made it green.

**Fix follows your recommended structure:**

1. New `Query::has_aggregate_count_on_range_anywhere` — recursive detector. Walks the top-level items, then `default_subquery_branch.subquery`, then every `conditional_subquery_branches[*].subquery`. Returns true if any `AggregateCountOnRange` is reachable from this query.

2. New gate at the entry of `GroveDb::prove_query_non_serialized` (before V0/V1 dispatch):
   ```rust
   if path_query.query.query.has_aggregate_count_on_range_anywhere()
       && let Err(e) = path_query.validate_aggregate_count_on_range()
   {
       return Err(e).wrap_with_cost(OperationCost::default());
   }
   ```
   `validate_aggregate_count_on_range` already enforces the canonical single-ACOR-no-subqueries shape, so detection-anywhere + validate-strict together close the surface.

**Coverage of the broader concern.** Added `aggregate_count_hidden_in_subquery_branch_is_rejected_at_entry` — constructs a query whose top-level items are a regular `Range` and whose `default_subquery_branch.subquery` carries the `AggregateCountOnRange`. Top-level `aggregate_count_on_range()` misses it, but the new recursive detector finds it; validate then rejects because the surrounding query isn't the canonical shape. Plus `has_aggregate_count_on_range_anywhere_walks_subqueries` in grovedb-query covers default+conditional subquery branches at the unit level.

Workspace: 1486 grovedb lib tests + 28 aggregate_count grovedb + 4 grovedb-query aggregate_count tests pass. `cargo fmt`/typos hooks clean.

PR #659, opened 2026-05-10
feat: add Element::NotSummed wrapper to opt-out of sum propagation

QuantumExplorer:

Symmetric counterpart to [Element::NonCounted](https://github.com/dashpay/grovedb/pull/654) (#654). Where NonCounted opts a child out of **count** propagation in count-bearing trees, NotSummed opts a sum-bearing subtree out of **sum** propagation in sum-bearing trees.

Use case: a SumTree of category subtotals, where one inner sum-tree should be informational only — it tracks its own sum but does not bubble up to the parent total.

## What was done?

Added a new `Element::NotSummed(Box<Element>)` variant that:
- behaves identically to its inner element for storage, hashing, and the inner sum-tree's own internal sum aggregate,
- contributes **0** to the parent sum tree's running sum,
- still contributes the inner element's count to a parent count tree (only sum is suppressed),
- may **only** wrap one of the four sum-tree variants (`SumTree`, `BigSumTree`, `CountSumTree`, `ProvableCountSumTree`),
- may **only** be inserted into a sum-bearing parent (`SumTree`, `BigSumTree`, `CountSumTree`, `ProvableCountSumTree`).

### Design choice: stricter inner-type whitelist than NonCounted

NonCounted can wrap any element (including items, references, all tree types). NotSummed restricts the inner to the four sum-tree variants because that is where the wrapper is semantically meaningful — a sum-bearing subtree that retains its own internal sum but contributes nothing to the parent. Wrapping items or non-sum trees would have no effect distinct from using the bare type.

The constructor `Element::new_not_summed` returns `Result` and rejects everything else with `InvalidInput`. Serialization and deserialization enforce the same whitelist, plus an O(1) two-byte pre-check that rejects all four wrapper-on-wrapper combinations (`[15,15]`, `[15,16]`, `[16,15]`, `[16,16]`) before bincode can recurse — closing a stack-exhaustion vector.

### Discriminant scheme

| | bincode (`Element`) | `ElementType` flag bit | twins |
|---|---|---|---|
| NonCounted (existing) | 15 | `0x80` | 128..=142 (15 twins) |
| **NotSummed (new)** | **16** | **`0x40`** | **68, 69, 71, 74** (only 4 twins) |

The disjoint flag bits keep `is_non_counted()` and `is_not_summed()` independently testable on `ElementType`. `base()` strips whichever flag is set. The two wrappers are mutually exclusive in practice — constructors and (de)serializers reject any nesting, including cross-nesting `NotSummed(NonCounted(_))` and `NonCounted(NotSummed(_))`.

### Cryptographic notes

The wrapper byte is part of the serialized value, so the value hash distinguishes `NotSummed(X)` from a bare `X`. The parent's `TreeFeatureType` for a not-summed child is the inner element's feature type with its sum component zeroed (via a new `TreeFeatureType::zero_sum()` helper, symmetric to `zero_count()`); no new `TreeFeatureType` variants are introduced.

### Scope

Touched 27 files across `grovedb-element`, `grovedb-query`, `grovedb-merk`, and `grovedb`. Mirrors PR #654's wiring throughout the stack:
- enum + discriminant + constructor + helpers + serialization
- merk: `TreeType::is_sum_bearing()`, insert validation, feature-type dispatch, get/cost paths, reconstruct preserves wrapper
- grovedb: batch propagation tracks `not_summed` flag alongside `non_counted` in `InsertTreeWithRootHash`, parent-type guard, estimated cost overhead, debugger transparency
- existing tests in `tests/{batch_unit,batch_coverage,batch_rejection}_tests.rs` updated for the new field

## How Has This Been Tested?

New tests:
- **grovedb-element** (10 tests): constructor whitelist (rejects items, sum items, plain trees, count trees, all wrapper nesting; accepts the four sum-tree variants); helper passthrough (`is_*`, `sum_value_or_default → 0`, `count_value_or_default` still propagates, `count_sum_value_or_default → (count, 0)`, flag accessors); bincode round-trip; reject nested wrappers at deserialize and at serialize; long-chain pre-check stack-overflow guard.
- **grovedb-element discriminant pinning**: pins `NOT_SUMMED_WRAPPER_DISCRIMINANT = 16` and the four `0x40 | base` twin discriminants; rejects all illegal inner bytes including the wrapper bytes themselves and the synthetic twin range.
- **grovedb-merk** (5 tests): insert into `NormalTree` and `CountTree` is rejected; insert into `SumTree` succeeds and contributes 0; `NotSummed(SumTree(_,100,_))` in a `ProvableCountSumTree` yields count=1 sum=0; `reconstruct_with_root_key` preserves the wrapper through propagation; constructor rejects non-sum-tree inner.
- **grovedb** (`tests/not_summed_tests.rs`, 5 tests): batch insert rejected into NormalTree and CountTree; direct insert in SumTree excludes subtree sum from parent aggregate; **batch propagation preserves wrapper** through `InsertTreeWithRootHash` (with both wrapper insert AND child insert in the same batch); `check_subtree_exists` works through the wrapper.

Full workspace test suite passes:
- `cargo test -p grovedb-element` — 74 passed (57 unit + 17 integration)
- `cargo test -p grovedb-query` — 235 passed
- `cargo test -p grovedb-merk --features minimal,test_utils,full` — 416 passed
- `cargo test -p grovedb` — 1495 passed

## Breaking Changes

None for existing callers. The change adds a new `Element` variant and is additive at the API surface. The on-disk serialization format gains bincode discriminant 16 — older code reading data written with this variant would fail to deserialize, so producer/consumer pairs need to be upgraded together. This matches the existing "ONLY APPEND TO THIS LIST" comment on the `Element` enum.

The internal `GroveOp::InsertTreeWithRootHash` variant gains a `not_summed: bool` field; the variant is `#[non_exhaustive]` and only constructed inside this crate, but the four in-tree test sites that build it directly are updated.

## Checklist:

### DIFF
--- a/grovedb-element/Cargo.toml
+++ b/grovedb-element/Cargo.toml
@@ -27,3 +27,6 @@ verify = []
 constructor = []
 serde = ["dep:serde"]
 visualize = ["dep:grovedb-visualize"]
+
+[dev-dependencies]
+serde_json = "1"
--- a/grovedb-element/src/element/constructor.rs
+++ b/grovedb-element/src/element/constructor.rs
@@ -439,17 +439,19 @@ impl Element {
         }
     }
 
-    /// Wrap `self` in `NotSummed` without re-validating the inner element's
-    /// type. Panics if `self` is not a sum-tree variant. Used by batch
-    /// propagation paths that have already established the element will be
-    /// a sum-tree type.
-    pub fn into_not_summed_unchecked(self) -> Self {
-        match Self::new_not_summed(self) {
-            Ok(wrapped) => wrapped,
-            Err(_) => panic!(
-                "into_not_summed_unchecked called on non-sum-tree element — caller must ensure \
-                 the element is one of SumTree/BigSumTree/CountSumTree/ProvableCountSumTree"
-            ),
+    /// Wrap `self` in `NotSummed`. If `self` is already `NotSummed`, returns
+    /// it unchanged (idempotent on `NotSummed`).
+    ///
+    /// Returns `InvalidInput` if `self` is `NonCounted` (the two wrappers
+    /// are mutually exclusive) or any non-sum-tree variant. Mirrors
+    /// [`Element::into_non_counted`].
+    pub fn into_not_summed(self) -> Result<Self, ElementError> {
+        match self {
+            Element::NotSummed(_) => Ok(self),
+            Element::NonCounted(_) => Err(ElementError::InvalidInput(
+                "cannot wrap NonCounted in NotSummed; wrappers are mutually exclusive",
+            )),
+            other => Self::new_not_summed(other),
         }
     }
 }
--- a/grovedb-element/src/element/constructor.rs
+++ b/grovedb-element/src/element/constructor.rs
@@ -407,4 +407,39 @@ impl Element {
             other => Element::NonCounted(Box::new(other)),
         }
     }
+
+    /// Wrap a sum-tree variant in `NotSummed` so it contributes 0 to its
+    /// parent sum tree's running sum when inserted. Counts (if any) still
+    /// propagate.
+    ///
+    /// Only the four sum-tree variants are accepted: `SumTree`, `BigSumTree`,
+    /// `CountSumTree`, `ProvableCountSumTree`. Any other element — including
+    /// items, sum items, references, non-sum trees, and any wrapper
+    /// (`NonCounted`, `NotSummed`) — is rejected with `InvalidInput`.
+    pub fn new_not_summed(inner: Element) -> Result<Self, ElementError> {
+        match inner {
+            Element::SumTree(..)
+            | Element::BigSumTree(..)
+            | Element::CountSumTree(..)
+            | Element::ProvableCountSumTree(..) => Ok(Element::NotSummed(Box::new(inner))),
+            _ => Err(ElementError::InvalidInput(
+                "NotSummed inner element must be a sum-tree variant (SumTree, BigSumTree, \
+                 CountSumTree, or ProvableCountSumTree)",
+            )),
+        }
+    }
+
+    /// Wrap `self` in `NotSummed` without re-validating the inner element's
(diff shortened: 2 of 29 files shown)


QuantumExplorer:
@coderabbitai review

This is Claude. The PR is intentionally in draft while we wait for the first review pass — please review it now.

QuantumExplorer (commented):
Codex review findings:

1. **[P1] Reject cross-wrapper construction** (`grovedb-element/src/element/constructor.rs:393-407`)

`new_non_counted` only rejects `NonCounted`, so it can still construct `NonCounted(NotSummed(...))` even though serialization rejects cross-wrapper nesting. That value can reach batch execution with `is_non_counted() == true` and `underlying() == NotSummed`, then hit the supposedly-unreachable wrapper arm instead of returning a typed error.

Suggested fix: make `new_non_counted` reject both wrapper variants, and make `into_non_counted` avoid wrapping `NotSummed` as well, either by making it fallible or by adding a separate checked helper for external callers. Add a batch regression for `NonCounted(NotSummed(SumTree))` to confirm it returns an `InvalidInput` / `InvalidBatchOperation` instead of panicking.

2. **[P2] Add serde-side wrapper validation** (`grovedb-element/src/element/mod.rs:45-47`)

The bincode serialize/deserialize paths now reject nested wrappers and `NotSummed` with a non-sum-tree inner, but the `serde` feature still derives recursive `Deserialize` for `Element`. With serde enabled, a payload can construct invalid `NotSummed(Item)` / cross-wrapper values, and deeply nested wrapper payloads still recurse before any validation runs.

Suggested fix: replace the derived serde deserialize path with a manual visitor or serde helper wrappers that enforce the same invariants as `Element::deserialize`: wrapper inners must not be wrappers, and `NotSummed` may only wrap `SumTree`, `BigSumTree`, `CountSumTree`, or `ProvableCountSumTree`. Add serde-feature tests for nested wrapper rejection and `NotSummed(non_sum_tree)` rejection.

QuantumExplorer:
Reviewed

PR #16, opened 2021-12-28
feat: multi path proof verification

iammadab:

PR #661, opened 2026-05-11
feat: add Element::ProvableSumTree + AggregateSumOnRange query

QuantumExplorer:

GroveDB had a symmetry gap in its aggregate trees:

| Variant | Aggregate | In node hash? |
|---|---|---|
| SumTree | sum (i64) | ❌ |
| CountTree | count (u64) | ❌ |
| ProvableCountTree | count (u64) | ✅ |
| ProvableSumTree (new) | sum (i64) | ✅ |
| BigSumTree | sum (i128) | ❌ |
| CountSumTree | count + sum | ❌ (count only on ProvableCountSumTree) |

This PR adds `Element::ProvableSumTree` — the missing parallel to `Element::ProvableCountTree`. It bakes the per-node aggregate sum into the node hash via `node_hash_with_sum(kv, l, r, sum)`, making **aggregate-sum range queries cryptographically verifiable** in O(log n + |boundary|) proof size.

The marquee user-visible feature is `AggregateSumOnRange` — ask "what is the total sum of children with keys in [a, b]?" against a ProvableSumTree, with a verifiable proof that a peer can check from the root hash alone, without ever seeing the underlying `SumItem` values.

`ProvableBigSumTree` (i128 variant) is intentionally **out of scope** — documented as a follow-up.

## What was done?

Seven commits, each a clean Phase boundary:

| # | Commit | Phase |
|---|---|---|
| 1 | [c95cf749](https://github.com/dashpay/grovedb/commit/c95cf749) | Types: `Element::ProvableSumTree`, `ElementType`, `TreeType::ProvableSumTree`, `TreeFeatureType::ProvableSummedMerkNode`, `NodeType::ProvableSumNode`, `zero_sum()` helper |
| 2 | [3364f08c](https://github.com/dashpay/grovedb/commit/3364f08c) | Merk: `node_hash_with_sum`, `AggregateData::ProvableSum`, new proof Node variants `KVSum`/`KVHashSum`/`KVRefValueHashSum`/`KVDigestSum`/`HashWithSum` at wire tags `0x30..=0x3D` |
| 3 | [46466d1c](https://github.com/dashpay/grovedb/commit/46466d1c) | Refactor: NotSummed twin reverted to `0xb0` prefix with **explicit per-variant mapping** rather than the `prefix \| base` formula. `NotSummedProvableSumTree = 177` hand-assigned. |
| 4 | [73697a8b](https://github.com/dashpay/grovedb/commit/73697a8b) | Wire ProvableSumTree through insert/read/batch paths |
| 5 | [40b3c168](https://github.com/dashpay/grovedb/commit/40b3c168) | `verify_grovedb` aggregate-consistency check across all 7 aggregate-bearing tree variants |
| 6 | [9c7d5e18](https://github.com/dashpay/grovedb/commit/9c7d5e18) | `QueryItem::AggregateSumOnRange` + prove/verify machinery with i128 accumulator and i64-narrowing overflow check |
| 7 | [e49fa100](https://github.com/dashpay/grovedb/commit/e49fa100) | Documentation: new `aggregate-sum-on-range-queries.md` chapter plus updates to `element-system`, `hashing`, `appendix-a`, and `aggregate-sum-queries` (disambiguation banner) |

### Discriminants (current state)

- `Element::ProvableSumTree` → bincode disc **17**
- `ElementType::ProvableSumTree` → **17**
- `ElementType::NonCountedProvableSumTree` → **145** (`0x80 | 17`)
- `ElementType::NotSummedProvableSumTree` → **177** (explicit slot, NOT `0xb0 | 17` — that would collide back to `Reference` under the legacy mask)
- `TreeType::ProvableSumTree` → **11**

**Note on PR #657 (CountIndexedTree).** This PR lands FIRST; PR #657 will need to rebase against develop once this merges and shift its discriminants up by 1: `CountIndexedTree` 17 → 18 and `ProvableCountIndexedTree` 18 → 19, with the NonCounted twins shifting 145/146 → 146/147 accordingly. The renumber is mechanical, same pattern as the historical [2498b97f](https://github.com/dashpay/grovedb/commit/2498b97f) merge resolution — just in the opposite direction this time.

### NotSummed twin encoding change

The existing `NotSummed` twin family used a `0xb0 | base` formula with a 4-bit base mask `0x0F`. ProvableSumTree at base 17 doesn't fit (mask collapses 17 → 1 = Reference). Rather than widen the mask (which would rebase every existing discriminant), this PR keeps the family range at `0xB0..=0xBF` and switches to **explicit per-variant slot assignment**. Existing twins keep their values (180, 181, 183, 186); ProvableSumTree gets the new explicit slot 177.

### Signed-sum correctness

Two arithmetic differences from the count machinery, documented inline:

- **No `if sum == 0 → short-circuit`.** Count can safely skip a 0-count subtree; sum can be 0 with non-zero children (e.g., +5 and -5). The prover descends regardless.
- **i128 accumulator with i64 narrowing at entry points.** The prover/verifier carry `i128` end-to-end and narrow once at the outermost call via `i64::try_from`. `i64::MAX + i64::MAX` correctly returns `InvalidProofError`; `i64::MAX + i64::MIN = -1` correctly fits.

## How Has This Been Tested?

`cargo test --workspace --all-features`: **2938 passed, 0 failed** (40 new tests across phases 3/4/5).

New test files / modules:

- `grovedb/src/tests/provable_sum_tree_tests.rs` — 10 round-trip + wrapper + nested-propagation tests + `integrity_walk_tests` module (7 tampering tests for Phase 4)
- `grovedb/src/tests/aggregate_sum_query_tests.rs` — 21 end-to-end query tests including negative sums, i64::MAX overflow rejection, tampering detection, NotSummed exclusion, validation rejection paths
- `merk/src/proofs/query/aggregate_sum.rs` tests — 14 Merk-level prover/verifier internals

Mandatory test coverage from the original spec:

- ✅ Direct insert + read round-trips with positive, negative, and i64::MIN/i64::MAX sums
- ✅ Aggregation up the tree (nested ProvableSumTree)
- ✅ `prove_aggregate_sum_query` round-trip
- ✅ Mixed positive/negative sums net correctly
- ✅ Overflow path (two i64::MAX children → `InvalidProofError`)
- ✅ `verify_grovedb` catches deliberately-corrupted ProvableSumTree node
- ✅ Existing SumTree tests still pass
- ✅ `Element::ProvableSumTree` byte-identical serialization round-trip
- ✅ `NotSummed(ProvableSumTree)` constructible + serializable + correctly suppresses parent sum

`cargo clippy --workspace --all-features --no-deps`: **zero warnings**.
`cargo fmt --all -- --check`: clean.

## Breaking Changes

**Pre-shipping V1 only.** The NotSummed twin discriminants did NOT change in this PR (they were temporarily widened to `0xA0..=0xBF` during one work-in-progress phase, then reverted in commit [46466d1c](https://github.com/dashpay/grovedb/commit/46466d1c) — existing twin values 180/181/183/186 are unchanged from before this PR).

The only new on-disk byte ranges are V1-only additions:
- Base discriminant 17 (`ProvableSumTree`)
- NonCounted twin 145
- NotSummed twin 177
- Proof Node wire tags `0x30..=0x3D` (V1 proofs only)

No V0 wire format changes.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas (twin scheme rationale, i128 narrowing, NotSummed wrapper semantics)

## Summary by CodeRabbit

* **New Features**
  * Added `ProvableSumTree` element type for cryptographic commitment of aggregate sums in node hashes
  * Introduced `AggregateSumOnRange` queries to prove and verify signed aggregate sums over key ranges without revealing individual values
  * Added sum-bearing proof nodes with deterministic encoding for verifiable range aggregation

* **Documentation**
  * Added guides for aggregate sum range queries, sum hashing semantics, and element system integration

* **Tests**
  * Added comprehensive test suites covering sum proof generation, verification, tampering detection, and cross-layer path validation

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/grovedb/pull/661)


QuantumExplorer:
This is Claude addressing the review feedback. Pushed 2b8490e3, ae7b971e, 8cec7b56, fce71323, ff5645b9.

**Fixed:**
- 2b8490e3 (Critical #1): reject `KVRefValueHashSum` in `extract_elements_and_leaf_keys` alongside `KVRefValueHash` and `KVRefValueHashCount`. Without this, the trunk/branch verifier would deserialize and insert the value bytes from a forged `KVRefValueHashSum` node — the embedded `node_value_hash` is opaque (`combine_hash` of node_value_hash and referenced_value_hash) and can't be recomputed at the trunk layer. Added regression test in `trunk_proof_tests.rs`.
- ae7b971e (Major #3): `has_aggregate_sum_on_range_anywhere` now also inspects the conditional-branch *selector* (map key) for `AggregateSumOnRange`, not only `branch.subquery`. The selector is itself a `QueryItem` and the type permits ASOR there. Added a regression test covering the selector case.
- 8cec7b56 (Minor #7): `verify_merk_and_submerks_in_transaction` uses `.entry().or_insert(...)` so the aggregate-consistency placeholder doesn't clobber a cryptographic `combined_value_hash != element_value_hash` entry that was inserted earlier for the same path.
- 8cec7b56 (Major #8): three `Error::InvalidProofError` sites in `merk/src/proofs/query/aggregate_sum.rs` reclassified to `Error::CorruptedData` — the prover walks our own merk, so an aggregate_data failure is local state corruption per the repo convention.
- 8cec7b56 (Nitpick #18): both ASOR `prove_aggregate_sum_on_range` call sites in `grovedb/src/operations/proof/generate.rs` wrap their MerkError with contextual `CorruptedData(format!("prove_aggregate_sum_on_range failed: {}", e))` per the repo's error-context convention.
- fce71323 (Minor #9): the populated-tree round-trip loop now tracks a running `expected_sum` and asserts the ProvableSumTree's aggregate matches it after every insert (7 → 20 → 40), not just the final value.
- fce71323 (Minor #10): `direct_insert_provable_sum_tree_with_root_key_and_sum` now actually performs the direct-insert it was named for — the non-batch insert path rejects non-empty Tree elements ("a tree should be empty…"), so the test uses `apply_batch` with `insert_only_known_to_not_already_exist_op` to write a ProvableSumTree carrying the captured `root_key` + `sum` at a fresh top-level key, then asserts both fields round-trip.
- fce71323 (Nitpick #16): the `NotSummed` wrapper zero-sum regression now also exercises `TreeType::ProvableSumTree`, asserting `ProvableSummedMerkNode(0)`.
- fce71323 (Nitpick #17): new round-trip test for the wire tag 7 introduced for `AggregateData::ProvableSum` in `merk/src/tree/link.rs` — encodes a negative-sum reference link, asserts tag 7 is present, decodes and checks all fields round-trip.
- fce71323 (Nitpick #19): `aggregate_sum_with_subquery_is_rejected_at_validation` now also calls `prove_query(...)` against the real 15-key fixture and asserts `Err`, so the prover-side gate (not just verifier-side) is regression-tested.
- ff5645b9 (Minor #12): refreshed the stale "Phase 1 / Phase 2 will diverge" comment on `TreeType::ProvableSumTree` and the matching comment in `element/tree_type.rs` to describe the actual post-Phase-2 semantics (sum baked into node hash via `node_hash_with_sum`).

**Skipped (false positives / out of scope):**
- #2 (serde variant name PascalCase vs snake_case): pre-existing intentional mismatch already documented in `query_item/mod.rs:1427` — `AggregateCountOnRange` has the same shape; the project uses `serde_test` token streams for round-trip testing precisely because of this. Fixing only the new variant would diverge it from the existing one. Skipping for parity.
- #4 (V0 hardcode in `verify_single_key_layer_proof_v0`): identical V0 hardcode in `aggregate_count.rs` at the same call sites. This is a pre-existing pattern, not introduced by this PR — the "v0" suffix in the function name is intentional. Fixing only the sum side would create count/sum inconsistency.
- #5 (sum variants collapse to `NoAggregateData` / `InvalidProofError` in `proofs/tree.rs`): the count counterparts (`KVHashCount`, `KVDigestCount`, `KVRefValueHashCount`) collapse the same way. The variants in question aren't produced by ProvableSumTree chunk generation (chunks emit `KVSum` / `KVValueHashFeatureType`), and the only `aggregate_data()` caller during restore (`chunk_tree` root) is a `KVValueHashFeatureType`. Pre-existing parallel-to-count pattern, no observable bug from this PR.
- #6 (`proof_node_type` fallback drops sum-awareness for raw Merk): same fallback drops count-awareness for raw Merk too (pre-existing). The fallback documents itself as "Default to tamper-resistant for raw Merk" — fixing only the sum side would diverge from count. Out of scope to expand to both.
- #11 (overflow test no-op early return): the test author already acknowledges the apply-side may or may not catch overflow with the comment "this scenario is additionally exercised at the verify layer via fabricated proofs"; the verify-path test handles the gate. Strengthening the apply-path assertion is low value.
- #13 (`appendix-a.md` TreeType discriminant): the "11" is the `TreeType` enum's own discriminant (see `merk/src/tree_type/mod.rs:72`), distinct from the Element discriminant. CodeRabbit's claim that it should be "17" is wrong — confirmed correct as-is.
- #14, #15 (deduplication / `cost_return_on_error!` reshape of pre-existing code): low value churn touching code outside this PR's strict scope.

**Test totals:** 2959 passing (was 2956). 3 new tests added across the fixes above. No regressions.

Per project memory rule: identifying as Claude on this comment. The branch `claude/distracted-margulis-99463b` only; no force pushes, no changes to `develop`.

QuantumExplorer:
This is Claude. Pushed [da53ef07](https://github.com/dashpay/grovedb/commit/da53ef07) addressing the Codex finding about the aggregate-sum verifier's loose tree-type check at the terminal layer.

**Verified the finding is real.** With the leaf at an empty Merk-backed tree of *any* type (not just empty `NormalTree` — also empty `SumTree`, `CountTree`, `ProvableCountTree`, etc.), every such tree's stored `value_hash = combine_hash(H(bytes), NULL_HASH)` matches the verifier's recomputation when paired with a forged empty leaf proof. The bypass returns `sum = 0` (the correct answer for an empty tree of any type), but the implicit claim "the leaf is a `ProvableSumTree`" is deceptive — a soundness gap, not a numeric forgery.

**Same bug existed on the count side.** Pre-existing this PR; fixed in the same commit.

**Fix.** `enforce_lower_chain` now takes `is_terminal: bool`. At intermediate depths the existing `is_any_tree()` check still applies (intermediate path elements may be any tree type). At the terminal depth the verifier requires:
- aggregate-sum: `matches!(element, Element::ProvableSumTree(..))`
- aggregate-count: `matches!(element, Element::ProvableCountTree(..) | Element::ProvableCountSumTree(..))`

Wrapper variants (`NonCounted`, `NotSummed`) continue to work transparently via the existing `into_underlying()` strip.

**Path == [] case.** Not vulnerable. At depth 0 the chain check is bypassed (no parent layer), and the merk-level hash divergence between `node_hash` and `node_hash_with_sum`/`node_hash_with_count` makes it computationally infeasible to forge a proof against the trusted root.

**Regression tests** (3 new, all use V0 envelope via `GROVE_V2` for surgical proof reconstruction):
- `empty_leaf_type_confusion_forgery_rejected` (sum, NormalTree leaf)
- `empty_provable_count_tree_at_leaf_rejected_for_sum` (sum, ProvableCountTree leaf — confirms type-specificity)
- `empty_leaf_type_confusion_forgery_rejected` (count, NormalTree leaf)

Each test honest-probes the leaf path with a single-key query to harvest valid layer-0 and layer-1 merk proof bytes, builds a forged `GroveDBProofV0` with an empty leaf, encodes, and verifies the new gate rejects it. Without the fix all three would accept the forgery.

Workspace: `cargo test --workspace --all-features` → 2962 pass / 0 fail (was 2959 / 0).

QuantumExplorer:
This is Claude. Pushed [cc1828dd](https://github.com/dashpay/grovedb/commit/cc1828dd) addressing CodeRabbit's empty-path follow-up on both aggregate verifiers.

**Verification of the finding.** Real concern. `verify_v0_layer` / `verify_v1_layer` short-circuit to the leaf when `depth == path_keys.len()`, which is true at depth 0 when `path == []`. The new terminal-type gate in `enforce_lower_chain` is never invoked in that case.

In practice the empty-path case is already protected by hash divergence: the GroveDB root merk is always a `NormalTree` (API guarantees this), so its `root_hash = node_hash(...)`. A forged proof of `HashWithSum` / `HashWithCount` ops would reconstruct via `node_hash_with_sum` / `node_hash_with_count` — distinct hash functions, no collision. So the caller's trusted-root compare catches the forgery cryptographically. But defense-in-depth: don't rely on hash divergence implicitly. Reject up-front.

**Fix.** `PathQuery::validate_aggregate_{sum,count}_on_range` now check `self.path.is_empty()` and return a clear `InvalidQuery` error naming why (root is always `NormalTree`, no valid `Provable*` target at root). The check fires at the entry of `verify_aggregate_{sum,count}_query` (validators are the first call) and at proof generation (the prover also validates the path query before dispatch).

**Tests** (2 new, pin rejection at both the `PathQuery` validator and the `verify_*_query` entrypoint):
- `empty_path_aggregate_sum_rejected_at_validation`
- `empty_path_aggregate_count_rejected_at_validation`

Workspace: 2964 pass / 0 fail (was 2962).

**On the V1-hardcoded-V0 duplicate.** Still skipping in this PR for the reason given in the prior thread — `verify_single_key_layer_proof_v0` is shared with the count side, and the same hardcoded `execute_proof(..., 0)` exists in the pre-existing `aggregate_count.rs` (introduced in PR #656). Fixing only the sum copy diverges them; fixing both is appropriately a follow-up PR scoped to "PROOF_VERSION_LATEST plumbing for layer-walking helpers" rather than landed inside this feature PR.

QuantumExplorer:
This is Claude. Pushed [b1027dcb](https://github.com/dashpay/grovedb/commit/b1027dcb) addressing all 4 CodeRabbit nitpicks:

- **Stale comment** in `NonAggregateInner::deserialize`: now names both aggregate variants.
- **Redundant disjunction** in the empty-leaf forgery test: dropped the looser `msg.contains("ProvableSumTree")` clause; pinned the exact phrase.
- **Silent-no-op overflow test**: replaced the early-return-on-both-inserts-accepted with an explicit "at least one of {insert, prove, verify} must reject" assertion. The test now fails if all three silently accept the i64::MAX + i64::MAX overflow.
- **MerkError context wrap**: `query_aggregate_sum` (and the count sibling, as you offered) now wrap the merk-level error with `Error::CorruptedData(format!("query_aggregate_{sum,count} at path {:?}: {}", path_slices, e))` per the repo's error-handling convention. Two test assertions updated from `MerkError(_)` to `CorruptedData(_)` to match the new contract.

Workspace `cargo test --all-features`: 3102 pass / 0 fail (unchanged).

QuantumExplorer:
This is Claude. Going back through the CodeRabbit review thread to make sure nothing was silently dismissed.

**Just-pushed [86b527be](https://github.com/dashpay/grovedb/commit/86b527be)** addressing the [appendix-a TreeType column](https://github.com/dashpay/grovedb/pull/661#discussion_r3219632656). CodeRabbit's specific claim ("TreeType should be 17") was wrong (11 is the correct `TreeType` enum discriminant per `merk/src/tree_type/mod.rs:77`), BUT the confusion was fair — the column header was ambiguous and I noticed the table had several pre-existing wrong values in adjacent rows. Fixed all of it:

- Column header renamed `TreeType` → `TreeType disc` with an explanatory intro paragraph.
- Added `N (VariantName)` labels to every tree row for unambiguous reading.
- Corrected three pre-existing wrong TreeType discriminants (BigSumTree, CountTree, CountSumTree).
- Swapped rows 8 and 9 — Element disc 8 is `ProvableCountTree`, 9 is `ItemWithSumItem` (the original table had them backwards).
- Tightened the `ProvableCountSumTree` Purpose to "only count in hash" since sum is unbound metadata in that variant.

**Re: the two remaining `Major`-tagged findings I'm still skipping:**

1. [PascalCase vs snake_case serde for `AggregateSumOnRange`](https://github.com/dashpay/grovedb/pull/661#discussion_r3219632669) — verified via JSON round-trip that `AggregateCountOnRange` has the exact same shape and round-trips correctly today; the `field_identifier`-only `rename_all` applies to keys, not variant names. Fixing only the new variant would diverge it from the count side, and the count side ships in production. Out of scope.

2. [V0 hardcode in `verify_single_key_layer_proof_v0` called from `verify_v1_layer`](https://github.com/dashpay/grovedb/pull/661#discussion_r3219632713) — same helper exists in `aggregate_count.rs` from PR #656 with the same hardcoded `execute_proof(..., 0)` for V1 layer steps. Fixing only the sum copy diverges the two. Appropriate as a follow-up PR scoped to "PROOF_VERSION_LATEST plumbing for aggregate-query layer-walking helpers" that touches both sides.

Both are documented for the cidx rebase / future PR. Everything else from the review has been addressed.

QuantumExplorer (commented):
reviewed

PR #663, opened 2026-05-14
feat(grovedb,query): allow AggregateCountOnRange as carrier subquery

QuantumExplorer:
## Summary

- Allow `AggregateCountOnRange` (ACOR) as a subquery item under an outer `Keys`/`Range*` carrier query — the natural multi-outer-key extension of the existing single-leg ACOR. Returns one `(outer_key, u64)` pair per matched outer key while leaving the leaf wire format byte-identical.
- Split `Query::validate_aggregate_count_on_range` into a leaf validator (existing semantics, renamed) and a new carrier validator, with the top-level dispatcher routing by query shape and returning the leaf inner range either way.
- Add `GroveDb::verify_aggregate_count_query_per_key(...) -> Result<(RootHash, Vec<(Vec<u8>, u64)>), Error>` that accepts both leaf and carrier shapes (leaf collapses to a single-entry Vec with an empty key). The legacy `verify_aggregate_count_query` now strictly validates the leaf shape.

## Why

Dash Platform's count-query module needs to answer
```
SELECT brand, COUNT(*) FROM widgets
WHERE brand IN ('brand_000', 'brand_001') AND color > 'color_00000500'
GROUP BY brand
```
against a `byBrand / <brand> / color / <ProvableCountTree>` layout. Today grovedb rejects the nested shape at `Query::validate_aggregate_count_on_range`, forcing callers to fan out one proof per outer key (k× larger proofs) or fall back to the unproven path. With this change the same GroveDB proof contains all k counts and is verified in one pass.

Asymptotic complexity for `|outer matches| = k`, outer tree `B`, leaf tree `C`: `O(k · (log B + log C))`.

## Wire-format compat

- Existing leaf-only ACOR proofs are byte-identical: drive's existing `aggregate_count_path_query` + `verify_aggregate_count_query` round-trip works without changes.
- The new carrier shape is structurally different in the `GroveDBProof` envelope (multi-key outer merk proof + per-outer-key `lower_layers`) — it cannot be confused for a leaf proof.

## What's NOT in this PR

- Drive-side consumer changes (separate PR in `dashpay/platform`).
- Conditional-subquery carriers (out of scope per spec).
- Multi-In on prefix (`IN × IN`) — the carrier accepts ONE outer fan-out dimension.

## Test plan

[...]

PR #664, opened 2026-05-15
feat(grovedb,query): allow SizedQuery::limit on carrier AggregateCountOnRange

QuantumExplorer:

Follow-up to #663 (allow `AggregateCountOnRange` as carrier subquery). That change deliberately accepted `Range*` outer items on carriers, but the `SizedQuery::limit` / `offset` constraint inherited from leaf-ACOR was still applied uniformly — which makes carrier ACOR with `Range*` outer items practically unusable above a certain dataset size (the proof grows linearly in the number of in-range outer keys, with no way for the caller to cap it).

The concrete upstream use case from `dashpay/platform` is a "Q8 with outer Range" `GROUP BY` shape:

```sql
SELECT brand, COUNT(*) FROM widgets
WHERE brand > 'brand_050' AND color > 'color_00000500'
GROUP BY brand
LIMIT 20
```

which maps to a carrier ACOR: outer item `RangeAfter("brand_050"..)`, `subquery_path = ["color"]`, subquery `AggregateCountOnRange(RangeAfter("color_00000500"..))`, capped at 20 outer matches via `SizedQuery::limit`. Today this `PathQuery` is rejected at validation with `InvalidQuery("AggregateCountOnRange queries may not set SizedQuery::limit")` even though the semantic is well-defined.

## What was done?

Split the SizedQuery-level size-constraint check by shape:

- **Leaf** (single `AggregateCountOnRange(_)` item, no subqueries): keep rejecting both `limit` and `offset`. A leaf returns a single `u64`; pagination would silently change the answer.
- **Carrier** (outer `Key`/`Range*` items routing to a leaf-ACOR subquery): accept `SizedQuery::limit` — it caps the number of outer-key matches the carrier walks. Each matched outer key still produces a complete leaf-ACOR `u64` (the inner range is *not* capped). `SizedQuery::offset` is still rejected on carriers; the error message documents that skipping outer matches changes which `(outer_key, u64)` pairs end up in the proof and the use case for that hasn't been designed yet.

Changes:

- `grovedb/src/query/mod.rs`: `SizedQuery::check_aggregate_count_size_constraints` is replaced by `check_leaf_aggregate_count_size_constraints` (strict) and `check_carrier_aggregate_count_size_constraints` (offset-only). `validate_aggregate_count_on_range` classifies the inner query first and then dispatches the appropriate check. `validate_leaf_aggregate_count_on_range` still uses the leaf-strict check.
- `grovedb/src/operations/proof/aggregate_count/{helpers,per_key}.rs`: threaded `SizedQuery::limit` through to the carrier verifier so the merk walker's `execute_proof(... limit, ...)` stops at the same boundary the prover's truncation produced. Without this the verifier rejects truncated proofs with "Proof is missing data for query."
- `grovedb/src/operations/get/query.rs`: `query_aggregate_count_per_key` (no-proof path) propagates `SizedQuery::limit` to the shallow outer-walk query.

The merk-level hash composition is unchanged; the prover's existing `Merk::prove_unchecked_query_items(query, limit, ...)` call already truncates correctly when the carrier merk's `limit` arrives non-`None`.

## How Has This Been Tested?

`cargo test --workspace --features 'minimal verify'` passes (all 1543 grovedb tests + the rest of the workspace).

Replaced `carrier_pagination_is_rejected_at_entry` with `carrier_aggregate_count_rejects_offset` (limit is no longer rejected; offset still is). Added five new tests in `grovedb/src/tests/aggregate_count_query_tests.rs`:

- `leaf_aggregate_count_still_rejects_limit` — the leaf strict path is byte-identical to before this PR.
- `carrier_keys_outer_with_limit_caps_results` — outer `Keys` + `limit=2` against 4 brands returns exactly 2 `(brand, u64)` pairs in lex-asc order, proof verifies.
- `carrier_range_outer_with_limit_caps_results` — outer `RangeAfter` + `limit=2` against 4 in-range brands behaves identically (the upstream Q8-with-Range shape).
- `carrier_range_outer_with_limit_zero_returns_no_results` — `limit=0` walks zero outer matches; exercised via the no-proof per-key API because the existing `prove_query_non_serialized` entry-point gate has a long-standing unconditional rejection of `limit == 0` for proved queries (`"proved path queries can not be for limit 0"`), which is outside the scope of this PR.
- `carrier_range_outer_with_limit_exceeding_available_walks_all` — `limit=100` against 2 in-range brands returns all 2, matching the no-proof per-key result byte-for-byte.

Also added two SizedQuery-level unit tests in `grovedb/src/query/mod.rs`: `sized_query_validate_leaf_acor_rejects_limit_and_offset` and `sized_query_validate_carrier_acor_accepts_limit_rejects_offset`.

`cargo clippy -p grovedb --lib --features minimal,verify -- -D warnings` is clean. `cargo fmt --check` is clean.

## Breaking Changes

None. Wire format is unchanged — `SizedQuery::limit` was already part of the serialized PathQuery and just rejected client-side at validation. Any previously-valid query keeps its byte-identical behavior. The relaxed rule applies unconditionally (no per-feature grove-version gate added); if maintainers want a version gate for the new carrier-with-limit acceptance so older verifiers continue to reject these queries, happy to add one — flagging here per the original spec.

Out of scope (intentionally deferred):

- `SizedQuery::offset` on carriers — the rationale is in the validator's error message; adding it later is a separate, well-scoped change.
- Limit on leaf ACOR queries (still rejected — a leaf returns one `u64` and pagination doesn't apply).
- The drive-side wire-up in `dashpay/platform` — a separate PR there consumes the new grovedb behavior.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed aggregate count proof verification to properly respect limit boundaries for carrier-shaped queries.

* **Enhancements**
  * Carrier-shaped aggregate count queries now support pagination limits to cap results.
  * Improved validation to distinguish between leaf and carrier query shapes with appropriate constraint rules.
  * Expanded test coverage for aggregate count query pagination behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/grovedb/pull/664)

### DIFF
--- a/grovedb/src/operations/get/query.rs
+++ b/grovedb/src/operations/get/query.rs
@@ -772,8 +772,15 @@ where {
     ///
     /// `path_query` must satisfy
     /// [`PathQuery::validate_aggregate_count_on_range`] in either
-    /// shape. Pagination is rejected. Each leaf subtree the walk
-    /// terminates in must be a `ProvableCountTree` or
+    /// shape. Pagination rules differ by shape: for **leaf** queries
+    /// both `SizedQuery::limit` and `SizedQuery::offset` are rejected
+    /// (a leaf returns a single `u64` and pagination would silently
+    /// change the answer); for **carrier** queries `SizedQuery::limit`
+    /// is accepted and caps the number of outer-key matches the walk
+    /// returns (each matched outer key still produces a complete
+    /// leaf-ACOR `u64`, the inner range is not capped), while
+    /// `SizedQuery::offset` is still rejected. Each leaf subtree the
+    /// walk terminates in must be a `ProvableCountTree` or
     /// `ProvableCountSumTree` — the merk-level walk rejects any other
     /// tree type.
     ///
--- a/grovedb/src/operations/get/query.rs
+++ b/grovedb/src/operations/get/query.rs
@@ -20,7 +20,7 @@ use crate::{
     query_result_type::{QueryResultElement, QueryResultElements, QueryResultType},
     reference_path::ReferencePathType,
     util::TxRef,
-    Element, Error, GroveDb, PathQuery, TransactionArg,
+    Element, Error, GroveDb, PathQuery, SizedQuery, TransactionArg,
 };
 use grovedb_costs::cost_return_on_error_default;
 #[cfg(feature = "minimal")]
@@ -833,9 +833,18 @@ where {
         // outer items at `path_query.path` without descending into the
         // subquery — we want just the matched outer keys, not the
         // (unproven) results of the leaf aggregate-count.
+        //
+        // Propagate `SizedQuery::limit` (validated as carrier-only
+        // above): it caps the number of outer-key matches the walk
+        // returns. Each matched outer key still produces a complete
+        // leaf-ACOR `u64` below. `offset` is rejected at validation, so
+        // we don't propagate it here.
         let mut shallow_query = grovedb_query::Query::new_with_direction(left_to_right);
         shallow_query.items = outer_items;
-        let shallow_pq = PathQuery::new_unsized(path_query.path.clone(), shallow_query);
+        let shallow_pq = PathQuery::new(
+            path_query.path.clone(),
+            SizedQuery::new(shallow_query, path_query.query.limit, None),
+        );
 
         let (matched, _skipped) = cost_return_on_error!(
             &mut cost,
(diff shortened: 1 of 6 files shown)


QuantumExplorer (commented):
Reviewed

PR #666, opened 2026-05-16
feat(grovedb,merk,element): add Element::NotCountedOrSummed wrapper for combined sum+count opt-out

QuantumExplorer:
## Summary

Adds a third element wrapper that suppresses **both** count **and** sum contributions when inserted into a count-and-sum-bearing parent. The existing wrappers each cover one axis: `NonCounted` suppresses count, `NotSummed` suppresses sum.

```rust
Element::NotCountedOrSummed(Box<Element>)
```

- **Inner element** MUST be one of the four sum-bearing tree variants:
  `SumTree`, `BigSumTree`, `CountSumTree`, `ProvableCountSumTree`.
  Same allow-list as `NotSummed` — for `SumTree` / `BigSumTree` the wrapper suppresses the implicit count-of-one a subtree would contribute to its parent count tree.
- **Parent at insert time** MUST be `CountSumTree` or `ProvableCountSumTree` — the only tree types where suppressing both axes is meaningful. Any other parent rejects the wrapper at insert with `InvalidInputError` (per-merk) or `InvalidBatchOperation` (batch).
- The three wrappers are mutually exclusive — constructors and (de)serializers reject any nesting in any direction.

Aggregation suppression flows through the standard helpers:
- `count_value_or_default` returns 0
- `sum_value_or_default` / `big_sum_value_or_default` return 0
- `count_sum_value_or_default` returns `(0, 0)`

## Discriminant layout

| | NonCounted | NotSummed | **NotCountedOrSummed (new)** |
|---|---|---|---|
| Wrapper byte | 15 | 16 | **17** |
| Twin prefix | `0x80` | `0xb0` | **`0xc0`** |
| Synthetic twins | 128–142 | 180, 181, 183, 186 | **196, 197, 199, 202** |
| `disc & 0xf0` | `0x80` | `0xb0` | `0xc0` |

The deserialize pre-check rejects every ordered pair of wrapper bytes to bound bincode recursion — no path can construct a nested wrapper from on-disk bytes.

## Wiring

- **`grovedb-element`**: new `Element::NotCountedOrSummed` variant, `ElementType` twins, constructor (`new_not_counted_or_summed`, `into_not_counted_or_summed`), helpers (`is_not_counted_or_summed`, `is_wrapped`), display, type-twin mapping, serde shadow, and serialize/deserialize guards.
[...]

PR #672, opened 2026-05-17
feat(merk,grovedb): reject NonCounted/NotCountedOrSummed in provable count parents

QuantumExplorer:
## Why

`ProvableCountTree` and `ProvableCountSumTree` bind their aggregate count into every node hash via `node_hash_with_count`. That committed count IS the answer to an `AggregateCountOnRange` query.

Today, `NonCounted` and `NotCountedOrSummed` children are accepted in those parents (via `is_count_bearing` / `is_count_and_sum_bearing`). They contribute `own_count = 0`, so the cryptographically-committed count silently disagrees with the actual number of stored elements — confusing for callers, and a footgun for proof-driven readers who reasonably assume "if I iterate this range I'll get the same number the proof returned."

This PR makes that shape impossible going forward, while leaving the wrappers fully usable in the non-provable count-bearing parents where the count is plain bookkeeping rather than a cryptographic commitment.

## Rule

| Wrapper | Allowed parent(s) | Newly rejected parent(s) |
|---|---|---|
| `NonCounted` | `CountTree`, `CountSumTree` | `ProvableCountTree`, `ProvableCountSumTree` |
| `NotCountedOrSummed` | `CountSumTree` | `ProvableCountSumTree` |
| `NotSummed` | unchanged (uses `is_sum_bearing`) | — |

`is_count_bearing` and `is_count_and_sum_bearing` are unchanged — they still answer the structural question "does this tree track count/sum?" for the aggregate logic that legitimately needs that. The new wrapper-acceptance rule lives in two new dedicated predicates:

```rust
TreeType::accepts_non_counted_children()          // CountTree | CountSumTree
TreeType::accepts_not_counted_or_summed_children() // CountSumTree
```

## Guard sites updated

- `merk/src/element/insert.rs` — `insert`, `insert_reference`, `insert_subtree`.
- `grovedb/src/batch/mod.rs` — batch validation guard, plus the `RefreshReference` trusted path.

Direct (per-merk) and batch surfaces both now reject the disallowed combinations with `InvalidInputError` / `InvalidBatchOperation`.

## Tests

- 6 existing acceptance tests in `merk/src/element/insert.rs`, `grovedb/src/tests/not_counted_or_summed_tests.rs`, and `grovedb/src/tests/aggregate_count_query_tests.rs` flipped to assert rejection.
- 4 new direct + batch rejection tests in `non_counted_tests.rs` pin both wrappers against both `Provable*` count parents.
- `non_counted_still_accepted_in_count_sum_tree` pins the deliberate remaining acceptance so a future tightening (e.g., "only `CountTree`") surfaces here.
- TreeType-level unit tests for the two new predicates.

Full local results:
- `cargo test -p grovedb-merk --lib` → 497 / 497 pass.
- `cargo test -p grovedb --lib` → 1718 / 1718 pass.
- `cargo test -p grovedb-element` → 4 / 4 pass.
- `cargo fmt --all -- --check` → clean.

## Compatibility

No grove-version gate — same approach used when the wrappers were originally added in [#654](https://github.com/dashpay/grovedb/pull/654) and [#666](https://github.com/dashpay/grovedb/pull/666). Platform upgrades are coordinated, and no production data currently uses the now-rejected shape (these wrappers landed recently). The dead verifier paths that handled `own_count = 0` in proof walks remain in place as defensive code for any historical reader.

### DIFF
--- a/grovedb-element/src/element/constructor.rs
+++ b/grovedb-element/src/element/constructor.rs
@@ -468,6 +468,10 @@ impl Element {
     /// Use `into_non_counted` to wrap idempotently when `inner` may already
     /// be `NonCounted`; use that helper's `Result` return for the
     /// cross-wrapper case.
+    ///
+    /// Note: at insert time the parent must be `CountTree` or
+    /// `CountSumTree` (the non-provable count-bearing variants). Provable
+    /// count parents reject the wrapper at the merk-layer insert guard.
     pub fn new_non_counted(inner: Element) -> Result<Self, ElementError> {
         if matches!(
             inner,
@@ -566,8 +570,10 @@ impl Element {
     /// `NotSummed`, `NotCountedOrSummed`) — is rejected with
     /// `InvalidInput`.
     ///
-    /// Note: at insert time the parent must be `CountSumTree` or
-    /// `ProvableCountSumTree`. The merk-layer insert guard enforces that.
+    /// Note: at insert time the parent must be `CountSumTree`. Provable
+    /// count-and-sum parents (`ProvableCountSumTree`) reject the wrapper
+    /// at the merk-layer insert guard — they cryptographically commit the
+    /// count into every node hash and must reflect actual contents.
     pub fn new_not_counted_or_summed(inner: Element) -> Result<Self, ElementError> {
         match inner {
             Element::SumTree(..)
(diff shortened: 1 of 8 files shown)


QuantumExplorer (commented):
Reviewed

PR #673, opened 2026-05-18
feat(element): add required_{item,reference}_with_sum_item_space helpers

QuantumExplorer:

`grovedb-element` exposes a worst-case storage-sizing helper for plain `Element::Item` (`Element::required_item_space`), but PR #670 (`ItemWithSumItem`) and #667 (`ReferenceWithSumItem`) shipped their sum-bearing siblings without equivalent helpers.

Drive callers in [dash-platform](https://github.com/dashpay/platform) need them for stateless-cost / dry-run fee estimation on summable-index inserts. Without them the estimate undercharges the bytes the `i64 sum_value` adds to the serialized element, and dry-run fees diverge from applied fees on every `documentsSummable` / `summable` / `rangeSummable` write. Three call sites in `rs-drive` currently carry `TODO(sum-feature, cost):` comments pointing at this gap.

## What was done?

Added two helpers on `impl Element` in `grovedb-element/src/element/helpers.rs`, paralleling `required_item_space`:

```rust
pub fn required_item_with_sum_item_space(len: u32, flag_len: u32, gv: &GroveVersion) -> Result<u32, ElementError>
pub fn required_reference_with_sum_item_space(path_len: u32, flag_len: u32, gv: &GroveVersion) -> Result<u32, ElementError>
```

Both add `+10` on top of the plain-item formula for the worst-case `i64` `sum_value`. bincode 2.x varint maxes at 9 bytes for a zigzag-encoded `u64`; `+10` is a deliberate safety margin so callers using this for stateless-cost / dry-run fee estimation **never undercharge**. The contract is a strict upper bound — exact equality is not required.

Wired two new `FeatureVersion` fields through `grovedb-version/src/version/grovedb_versions.rs::GroveDbVersionElement`:

```rust
pub required_item_with_sum_item_space: FeatureVersion,
pub required_reference_with_sum_item_space: FeatureVersion,
```

…and initialized them to `0` alongside the existing `required_item_space: 0` in `v1.rs`, `v2.rs`, and `v3.rs`.

## How Has This Been Tested?

Four new tests in `grovedb-element/tests/element_constructors_helpers.rs`:

- `required_item_with_sum_item_space_matches_manual_formula` — mirrors the existing `required_item_space_matches_manual_formula` shape.
- `required_item_with_sum_item_space_is_upper_bound` — sweeps payload sizes (`0`, `1`, `250`, `65_500` bytes), flag variants, and boundary sum values (`0`, `±1`, `±250`, `i64::MAX`, `i64::MIN`, `i32::MAX`), asserting `required >= serialize().len()` for every combination.
- `required_reference_with_sum_item_space_matches_manual_formula`.
- `required_reference_with_sum_item_space_is_upper_bound` — sweeps path variants, `max_hop` values, flag variants, and boundary sums; uses bincode to compute the worst-case path payload size the way real callers do (e.g. `add_document_to_primary_storage`).

Verified:
- `cargo test -p grovedb-element --test element_constructors_helpers` — all 24 tests pass.
- `cargo build` — full workspace builds cleanly.

## Breaking Changes

None. Two new `pub fn` helpers, two new fields appended to `GroveDBElementMethodVersions` (initialized to `0` in every version table), and no changes to existing behavior.

## Checklist:

## Follow-up

Dash-platform PR can now switch from `Element::required_item_space(...)` to the new sum-aware helpers in three `rs-drive` call sites currently carrying `TODO(sum-feature, cost):`:
- `packages/rs-drive/src/drive/document/insert/add_document_to_primary_storage/v0/mod.rs:591`
- `packages/rs-drive/src/drive/document/insert/add_reference_for_index_level_for_contract_operations/v0/mod.rs:236, :280`

Once both are wired through, `add_estimation_costs_for_contract_insertion/v1/mod.rs:154` can set non-zero `sum_trees_weight` / `count_sum_trees_weight` weights for the estimation layers (currently hardcoded to 0).

### DIFF
--- a/grovedb-element/src/element/helpers.rs
+++ b/grovedb-element/src/element/helpers.rs
@@ -622,6 +622,70 @@ impl Element {
         Ok(len + len.required_space() as u32 + flag_len + flag_len.required_space() as u32 + 1)
     }
 
+    /// Worst-case serialized-storage cost of an
+    /// [`Element::ItemWithSumItem`]. Same shape as
+    /// [`Self::required_item_space`] with an extra worst-case allowance
+    /// for the `i64` sum_value field.
+    ///
+    /// The sum_value is bincode varint-encoded (with zigzag for negative
+    /// values) inside `Element::serialize`. The bincode 2.x varint encoding
+    /// for a u64 is at most 9 bytes (1 marker + 8 bytes). We use 10 here
+    /// as a deliberate upper-bound margin so callers using this for
+    /// stateless-cost / dry-run fee estimation never undercharge.
+    pub fn required_item_with_sum_item_space(
+        len: u32,
+        flag_len: u32,
+        grove_version: &GroveVersion,
+    ) -> Result<u32, ElementError> {
+        check_grovedb_v0!(
+            "required_item_with_sum_item_space",
+            grove_version
+                .grovedb_versions
+                .element
+                .required_item_with_sum_item_space
+        );
+        Ok(
+            len + len.required_space() as u32
+                + flag_len
+                + flag_len.required_space() as u32
+                + 10
+                + 1,
+        )
+    }
+
+    /// Worst-case serialized-storage cost of an
+    /// [`Element::ReferenceWithSumItem`]. Parallels
+    /// [`Self::required_item_with_sum_item_space`] with the reference path
+    /// as the variable-length payload.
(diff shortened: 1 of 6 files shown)


QuantumExplorer (commented):
Reviewed

PR #724, opened 2026-05-20
fix(element): reject trailing bytes during deserialization

### LINKED ISSUE #677: [audit][high] Element::deserialize accepts trailing bytes

## Summary
Element decoding uses `bincode::decode_from_slice` but ignores the returned consumed-byte count.

## Impact / failure scenario
`serialize(element) || junk` decodes as the same logical element. This makes element bytes non-canonical and affects batch, proof, and cost paths that call `Element::deserialize`.

## References
- `grovedb-element/src/element/serialize.rs:129`
- `grovedb-element/src/element/serialize.rs:133`
- `grovedb/src/batch/mod.rs:2857`
- `merk/src/element/costs.rs:111`

## Suggested fix
Capture `(elem, consumed)` and reject inputs where `consumed != bytes.len()` unless a specific caller is explicitly using framed data.

## Suggested tests
Append one trailing byte to serialized item/tree/reference elements and assert deserialization fails. Add proof-path regression coverage for element values with trailing bytes.

QuantumExplorer:
## Summary
- reject bincode element decodes that leave unconsumed bytes
- add regression coverage for item, tree, sum item, and reference elements with appended junk

Fixes #677.

## Verification
- cargo test -p grovedb-element deserialize_rejects_trailing_bytes --lib
- cargo test -p grovedb-element serialization --lib
- cargo test -p grovedb-element

### DIFF
--- a/grovedb-element/src/element/serialize.rs
+++ b/grovedb-element/src/element/serialize.rs
@@ -126,11 +126,17 @@ impl Element {
             ));
         }
         let config = config::standard().with_big_endian().with_no_limit();
-        let elem: Element = bincode::decode_from_slice(bytes, config)
+        let (elem, consumed): (Element, usize) = bincode::decode_from_slice(bytes, config)
             .map_err(|e| {
                 ElementError::CorruptedData(format!("unable to deserialize element {}", e))
-            })?
-            .0;
+            })?;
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Approved

PR #726, opened 2026-05-20
fix(costs): version default-section storage removal accounting

### LINKED ISSUE #683: [audit][medium] Storage removal addition drops default-section removals

## Summary
Combining basic and sectioned storage removals removes/updates the default identifier entry but does not reinsert it.

## Impact / failure scenario
Removal bytes from the default identifier can be lost, undercounting storage removals and any costs derived from them.

## References
- `costs/src/storage_cost/removal.rs:83`
- `costs/src/storage_cost/removal.rs:102`
- `costs/src/storage_cost/removal.rs:148`

## Suggested fix
Reinsert the updated default epoch map before returning or assigning the combined `SectionedStorageRemoval`.

## Suggested tests
Assert `Basic + Sectioned(default)`, `Sectioned(default) + Basic`, and `AddAssign` all preserve both removal amounts.

---

## September 2026 audit addendum — C008

Audit group: **C008**. Classification: **correctness**. Provisional severity: **medium**.

The audited mixed Basic/Sectioned removal branches detach and update the default-owner epoch map without reinserting it. This loses existing epoch attribution and incoming basic bytes. The audit also traces KeyValueStorageCost::combined_removed_bytes and final storage accumulation. Sectioned += Basic is already the effective sibling control and must not be regressed. This is the existing #683 defect, not unauthorized credit creation.

Expected contract and correction: Apply the version-selected correction proposed in #726 while preserving legacy V1-V3 results as required. Cover every mixed branch and aggregation order. The PR body reports that callers aggregating outside GroveDB must select the corrected behavior separately; that downstream integration is not independently verified here.

### Validation to complete

- Pin legacy outputs and corrected mixed-order/default-epoch totals.
- Keep the always-correct Sectioned += Basic branch unchanged on legacy versions.
- Verify GroveDB entry-point guards and document external aggregation requirements.

### Limits and existing work

- PR #726 claims downstream refund relevance; the audit did not execute or independently verify a monetary integration.

Related tracking: [issue #683](https://github.com/dashpay/grovedb/issues/683) (open), [PR #726](https://github.com/dashpay/grovedb/pull/726) (open).

Scope: saved GroveDB worktree with revision context `2fa0f133877420a0d9c91ba7bc51b1775ab8c783`. This report does not establish that current `develop` or any deployed application is affected. Focused runtime validation remains outstanding.

<details>
<summary>Audit source and canonical finding identifiers</summary>

Source status: snapshot-backed (`git_worktree`); plain source locations are used because this is not a sealed commit-only scan.

Audited revision context: `2fa0f133877420a0d9c91ba7bc51b1775ab8c783`.

The findings were manually reconciled from a preserved scan bundle. The native scan ended before final completion; these are provisional source-review findings, not a completed native scan certification.

Canonical finding ID: `csf_d0684b045cb881581559eaaa`

Primary fingerprint: `codex-security/v1:sha256:5cac73b5027bfe0390f2ba6beb3ca2b251729abd9efff2e3f012f1a997fe145d`

Source locations:

- Location (root_control): `costs/src/storage_cost/removal.rs:65-119`
- Location (implementation): `costs/src/storage_cost/removal.rs:140-183`
- Location (implementation): `grovedb-epoch-based-storage-flags/src/lib.rs:745-767`
- Location (implementation): `storage/src/rocksdb_storage/storage.rs:389-404`
- Location (root_control): `costs/src/storage_cost/removal.rs:76-90`
- Location (alternate_route): `costs/src/storage_cost/removal.rs:95-109`
- Location (alternate_route): `costs/src/storage_cost/removal.rs:146-160`
- Location (entrypoint): `costs/src/storage_cost/key_value_cost.rs:113-116`
- Location (sink): `storage/src/rocksdb_storage/storage.rs:398-401`

</details>

QuantumExplorer:
## Summary

Fixes #683. Combining a `BasicStorageRemoval` with a `SectionedStorageRemoval` can discard the default identifier's entire epoch map, losing both its existing bytes and the basic bytes being added. Preserve that section under GROVE_V4 while retaining exact GROVE_V1–V3 cost results.

Three operator arms had the defect: `Basic + Sectioned`, `Sectioned + Basic`, and `Basic += Sectioned`. `Sectioned += Basic` was already correct and remains unchanged on every version.

## Version selection and scope

- Add `grovedb_versions.storage_costs.add_basic_storage_removal_to_sectioned_storage_removal`, set to `0` for V1–V3 and `1` for V4.
- Carry the version through the existing cost operators with a thread-local selector and scoped guard. The selector defaults to legacy behavior. GroveDB sectional-delete and batch entry points, and the innermost Merk apply entry point, select the requested version for synchronous execution and storage-cost accumulation.
- Keep the guard `!Send` and `!Sync`, so its destructor restores the originating thread's selector. Mark it `must_use` and document synchronous scope use, reverse-order drops for nested guards, and the prohibition on holding it across `.await`. The closure helper does not extend its scope to futures returned by the closure.

Aggregation outside a GroveDB call, including Drive's `combine_cost_operations`, still uses legacy arithmetic unless the consumer installs the guard using the applicable GroveVersion's storage-cost slot. That downstream integration remains a separate change; upgrading GroveDB alone does not correct every external cost sum.

## Production impact

Drive's storage-flags callback returns basic/basic removals for unflagged elements and sectioned/sectioned removals for flagged elements, with no-removal results where appropriate for zero bytes. Mixed basic/sectioned arithmetic arises when those results are aggregated. The GroveDB deletion tests deliberately use custom basic/sectioned callbacks to exercise the affected arithmetic.

Only the default identifier's section is dropped. Drive separates that section into `FeeResult.removed_bytes_from_system` before calculating identity refunds; identity-owned sections survive in both legacy and corrected arithmetic. The 0-of-155 and 0-of-302 test results below demonstrate missing removal accounting, not missing identity refunds on each Drive deletion. See [Platform's fee conversion](https://github.com/dashpay/platform/blob/a57d7c609e/packages/rs-drive/src/fees/op.rs#L268). Preserving V1–V3 outputs maintains historical cost compatibility without asserting a demonstrated identity-refund or consensus defect.

## Validation

- Exact-map regressions cover all four mixed operator arms with the default section present with `UNKNOWN_EPOCH`, present without it, and absent. They preserve other owners and epochs, pin legacy losses, and assert that `Sectioned += Basic` is identical unguarded, under version 0, and under version 1.
- Version tests pin the selector to `0/0/0/1` for V1–V4.
- A custom-callback single deletion reports exactly 155 removed bytes under V4 and an empty map under V3. A two-item batch reports exactly 302 bytes with both epochs preserved under V4 and an empty map under V3.
- Compile-fail regressions reject moving the guard between threads or requiring it to be `Sync`. Runtime tests check thread isolation, nesting, restoration to legacy, and restoration after a nested panic.

Local checks:

```text
cargo test --locked --offline -p grovedb-costs -p grovedb-version
cargo test --locked --offline -p grovedb --all-features --lib basic_plus_default_sectioned_removal_cost
cargo test --locked --offline -p grovedb --all-features --lib batch_delete_
cargo clippy --locked --offline --workspace --all-features -- -D warnings
cargo fmt --all --check
```

### DIFF
--- a/costs/src/storage_cost/removal.rs
+++ b/costs/src/storage_cost/removal.rs
@@ -28,6 +28,7 @@
 
 use std::{
     borrow::BorrowMut,
+    cell::Cell,
     cmp::Ordering,
     collections::BTreeMap,
     ops::{Add, AddAssign},
@@ -60,6 +61,83 @@ pub enum StorageRemovedBytes {
     SectionedStorageRemoval(StorageRemovalPerEpochByIdentifier),
 }
 
+thread_local! {
+    static BASIC_SECTIONED_REMOVAL_ADDITION_VERSION: Cell<u16> = const { Cell::new(0) };
+}
+
+/// Guard that restores the previous storage-removal arithmetic version when
+/// dropped.
+pub struct BasicSectionedRemovalAdditionVersionGuard {
+    previous_version: u16,
+}
+
+impl Drop for BasicSectionedRemovalAdditionVersionGuard {
+    fn drop(&mut self) {
+        BASIC_SECTIONED_REMOVAL_ADDITION_VERSION.with(|current_version| {
+            current_version.set(self.previous_version);
+        });
+    }
+}
+
+/// Use a storage-removal arithmetic version until the returned guard is
+/// dropped.
+pub fn use_basic_sectioned_removal_addition_version(
+    version: u16,
+) -> BasicSectionedRemovalAdditionVersionGuard {
+    let previous_version = BASIC_SECTIONED_REMOVAL_ADDITION_VERSION
+        .with(|current_version| current_version.replace(version));
+    BasicSectionedRemovalAdditionVersionGuard { previous_version }
+}
+
+/// Run storage-removal arithmetic using a specific version.
+pub fn with_basic_sectioned_removal_addition_version<T>(version: u16, f: impl FnOnce() -> T) -> T {
+    let _guard = use_basic_sectioned_removal_addition_version(version);
+    f()
+}
+
+fn basic_sectioned_removal_addition_version() -> u16 {
+    BASIC_SECTIONED_REMOVAL_ADDITION_VERSION.with(Cell::get)
+}
+
+fn add_basic_removal_to_sectioned_map(
+    map: &mut StorageRemovalPerEpochByIdentifier,
+    removed_bytes: u32,
+) {
+    let epoch_map = map.entry(Identifier::default()).or_default();
+    let old_value = epoch_map.remove(UNKNOWN_EPOCH).unwrap_or_default();
+    epoch_map.insert(UNKNOWN_EPOCH, old_value.saturating_add(removed_bytes));
+}
+
+fn legacy_add_basic_removal_to_sectioned_map(
+    map: &mut StorageRemovalPerEpochByIdentifier,
+    removed_bytes: u32,
+) {
+    let default = Identifier::default();
+    if let std::collections::btree_map::Entry::Vacant(e) = map.entry(default) {
+        let mut new_map = IntMap::new();
+        new_map.insert(UNKNOWN_EPOCH, removed_bytes);
+        e.insert(new_map);
+    } else {
+        let mut old_section_map = map.remove(&default).unwrap_or_default();
+        if let Some(old_value) = old_section_map.remove(UNKNOWN_EPOCH) {
+            old_section_map.insert(UNKNOWN_EPOCH, old_value.saturating_add(removed_bytes));
+        } else {
+            old_section_map.insert(UNKNOWN_EPOCH, removed_bytes);
+        }
+    }
+}
+
+fn add_basic_removal_to_sectioned_map_for_current_version(
+    map: &mut StorageRemovalPerEpochByIdentifier,
+    removed_bytes: u32,
+) {
+    if basic_sectioned_removal_addition_version() >= 1 {
+        add_basic_removal_to_sectioned_map(map, removed_bytes);
+    } else {
+        legacy_add_basic_removal_to_sectioned_map(map, removed_bytes);
+    }
+}
+
 impl Add for StorageRemovedBytes {
     type Output = Self;
 
@@ -74,38 +152,14 @@ impl Add for StorageRemovedBytes {
                 NoStorageRemoval => BasicStorageRemoval(s),
                 BasicStorageRemoval(r) => BasicStorageRemoval(s.saturating_add(r)),
                 SectionedStorageRemoval(mut map) => {
-                    let default = Identifier::default();
-                    if let std::collections::btree_map::Entry::Vacant(e) = map.entry(default) {
-                        let mut new_map = IntMap::new();
-                        new_map.insert(UNKNOWN_EPOCH, s);
-                        e.insert(new_map);
-                    } else {
-                        let mut old_section_map = map.remove(&default).unwrap_or_default();
-                        if let Some(old_value) = old_section_map.remove(UNKNOWN_EPOCH) {
-                            old_section_map.insert(UNKNOWN_EPOCH, old_value.saturating_add(s));
-                        } else {
-                            old_section_map.insert(UNKNOWN_EPOCH, s);
-                        }
-                    }
+                    add_basic_removal_to_sectioned_map_for_current_version(&mut map, s);
                     SectionedStorageRemoval(map)
                 }
             },
             SectionedStorageRemoval(mut smap) => match rhs {
                 NoStorageRemoval => SectionedStorageRemoval(smap),
                 BasicStorageRemoval(r) => {
-                    let default = Identifier::default();
-                    if let std::collections::btree_map::Entry::Vacant(e) = smap.entry(default) {
-                        let mut new_map = IntMap::new();
-                        new_map.insert(UNKNOWN_EPOCH, r);
-                        e.insert(new_map);
-                    } else {
-                        let mut old_section_map = smap.remove(&default).unwrap_or_default();
-                        if let Some(old_value) = old_section_map.remove(UNKNOWN_EPOCH) {
-                            old_section_map.insert(UNKNOWN_EPOCH, old_value.saturating_add(r));
-                        } else {
-                            old_section_map.insert(UNKNOWN_EPOCH, r);
-                        }
-                    }
+                    add_basic_removal_to_sectioned_map_for_current_version(&mut smap, r);
                     SectionedStorageRemoval(smap)
                 }
                 SectionedStorageRemoval(rmap) => {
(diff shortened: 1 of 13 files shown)


QuantumExplorer:
This is Claude (AI-assisted). Pushed a rework (commit 5d827ed4) addressing the consensus-safety problem in the previous version.

**The bug being fixed:** the default-section-drop only ever affected three of the four basic-into-sectioned arms — `Add: Basic+Sectioned`, `Add: Sectioned+Basic`, and `AddAssign: Basic+=Sectioned`. The fourth, `AddAssign: Sectioned+=Basic`, already reinserted the default section correctly in every shipped version.

The prior version routed **all four** arms through the version selector, so the legacy/v0 path dropped the default section for the fourth arm too — i.e. it *changed shipped v1/v2 output* (its own regression test even asserted the wrong `0` for that case). Since storage-removal bytes feed fee/state accounting, that's a v1/v2 replay divergence.

**Change:**
- `Sectioned += Basic` now calls the default-section-preserving helper unconditionally (it was never broken, so it must not be gated). The three genuinely-buggy arms remain gated to v3.
- Fixed the regression test to assert the preserved total (`13`) on the legacy path, with a comment explaining why that arm is version-independent.
- Documented why the thread-local selector exists (the version-sensitive combination happens inside `Add`/`AddAssign`, reached through ~hundreds of version-less `StorageCost`/`OperationCost` aggregation sites; the costs crate has no `grovedb-version` dep) and that its default (`0`/legacy) is the safe one.

On the thread-local: I looked at threading the version explicitly instead, but the combination flows through `StorageCost`/`OperationCost` operator overloads at ~951 version-less call sites, so explicit threading would require a large, risky refactor of the cost system. The thread-local is the pragmatic localization; it's now documented and the legacy default is safe.

**Verified:** `grovedb-costs` + `grovedb-version` tests pass (incl. `v3_uses_fixed_basic_to_sectioned_storage_removal_addition` and the corrected regression test); `grovedb` deletion-cost tests pass (incl. `latest_delete_preserves_basic_plus_default_sectioned_removal_cost`); clippy clean across all three.

QuantumExplorer:
This is Claude (AI-assisted). Assessment of how needed this fix is, and why the PR is being marked **deferred** rather than merged now.

**TL;DR:** the fix is correct, V4-gated, CI-green and the legacy path is pinned byte-exact — but the bug has **no effect on any fee, refund, pool credit, or root hash Platform computes today**. It is hygiene / future-proofing, not a live defect, so it is parked rather than carrying a thread-local version selector into `grovedb-costs` for a no-impact bug.

### What the bug can actually lose
- Only the **default identifier (`[0; 32]`) section** is dropped by the three buggy arms (`Add: Basic + Sectioned`, `Add: Sectioned + Basic`, `AddAssign: Basic += Sectioned`). Identity-owned sections (owner_id ≠ default) are never touched, so **identity fee refunds are unaffected in every shipped version**.
- What is lost is "system" bytes: the `Basic` removal being folded in, plus the default section it was folded into.

### How Drive reaches the buggy arms
- **Within a single Drive op: never.** `StorageFlags::split_removal_bytes` (`grovedb-epoch-based-storage-flags/src/split_removal_bytes.rs`) returns `Basic/Basic` for unflagged elements and `Sectioned/Sectioned` for flagged ones, so a deletion's key + value removals always combine through a same-kind (correct) arm.
- **Across ops in one batch / cost accumulation:** only when an unflagged removal (`Basic`) is aggregated with a **no-owner-flagged** removal (default-identifier section), and only in the `Basic += Sectioned` order (`Sectioned += Basic` was always correct). The only production no-owner flags in rs-drive are the identity-tree elements (`add_new_identity`: "no reason to store the owner id"); documents, contracts, etc. all carry an owner. Identities are never deleted and their values rarely shrink, so the trigger is rare.

### What happens when it does fire
- Drive routes both `BasicStorageRemoval` amounts and the default-identifier section to `FeeResult.removed_bytes_from_system` (`rs-drive/src/fees/op.rs`, the `BasicStorageRemoval(amount) => (FeeRefunds::default(), amount)` / `remove(&Identifier::default())` arms).
- Nothing in rs-drive-abci consumes that field: it is summed in `FeeResult::checked_add_assign`, zeroed in estimates, and not part of `validate_fees_of_event` (which compares `total_base_fee` / `storage_fee` / `processing_fee`). Storage fees are `added_bytes × credit_per_byte`; refunds come only from owner sections.
- **So even when triggered: no fee, refund, pool credit, or committed state changes.**

### Why defer rather than merge
- The Codex "medium" rating describes the arithmetic in isolation; the Drive-level impact is nil today.
- The fix is cheap and safe, but its design cost is real: a thread-local `Cell<u16>` selector in `grovedb-costs` installed via RAII guards at the Merk/GroveDB entry points (the operator impls can't carry a `grove_version` and `grovedb-costs` has no `grovedb-version` dependency). It also leaves a gap — removal aggregation outside a GroveDB call (Drive's `combine_cost_operations`, `FeeResult` sums) stays on the legacy arithmetic under V4 unless Drive installs the guard itself.
- Both the mechanism and the gap only matter if something starts consuming `removed_bytes_from_system` (e.g. system storage accounting). Until then there is nothing to fix in production.

### State of the branch (so it can be picked back up quickly)
- Rebased on develop, V4-gated (`v1..v3 = 0`, `v4 = 1`), CI green, CodeRabbit clean.
- Regressions pin the exact shipped figures: under GROVE_V3 the combined removal is `SectionedStorageRemoval({})` → **0** of 155 added bytes; under GROVE_V4 `{default: {UNKNOWN_EPOCH: 155}}` → **155**.
- **Re-open / merge when:** Platform starts consuming `removed_bytes_from_system`, or a GroveDB consumer needs `Basic` + default-section removals to aggregate correctly, or the next time a V4 cost-gate sweep is being merged anyway (zero marginal risk then). Otherwise this can be closed with an entry in `docs/audit-non-issues.md` ("default-section loss only affects `removed_bytes_from_system`, which has no consumer").

QuantumExplorer:
This is Claude (AI-assisted). Brought the branch up to date and validated it against the September 2026 audit addendum on #683 (group C008). Draft status and the `deferred` label are untouched — whether to lift them stays with the maintainer.

**Rebase.** Rebased the four existing commits onto `develop` at `6ec8feec` (31 commits behind). It applied cleanly: the `storage_costs` slot in `GroveDBVersions` and its `v1..v4` initialisers slotted in next to the newer V4 gates without conflicts. No production code changed in this push; the fix itself is as described in the PR body. CI green on the new head (`c00356e5`): all three test shards, linting, formatting, security audit, codecov patch/project. CodeRabbit skips reviews on drafts and there are no open inline comments.

**Validation matrix added** (the addendum's three items):

1. *Pin legacy and corrected mixed-order / default-epoch totals.* `costs/tests/coverage_regression.rs` now pins the **exact resulting map** for all four arms — `Basic + Sectioned`, `Sectioned + Basic`, `Basic += Sectioned`, `Sectioned += Basic` — over three default-section shapes (present with `UNKNOWN_EPOCH`, present without it, absent), under version `0` (V1..V3) and `1` (V4). The input default owner already carries `{3: 20, 5: 30, UNKNOWN: 7}` next to an identity-owned `{2: 40}`, with 11 incoming basic bytes, so the legacy result `{identity: {2: 40}}` (40 of 108) shows both losses the audit names: the owner's existing epoch attribution and the incoming basic bytes. The no-`UNKNOWN_EPOCH` shape pins the legacy result at an empty map (0 of 31) — that is the branch codecov flagged as uncovered before.
2. *`Sectioned += Basic` unchanged on legacy.* Asserted byte-identical unguarded, under version `0`, and under version `1` for all three shapes; it produces the corrected map in every case.
3. *Entry-point guards and external aggregation.* Every in-repo caller of the mixed arms (`KeyValueStorageCost::combined_removed_bytes` at the storage-batch commit, plus `StorageCost`/`OperationCost` `+=`) runs inside the guarded scopes; `delete_up_tree_while_empty_with_sectional_storage` funnels into the guarded batch apply. New GroveDB test: two epoch-flagged items deleted in one `apply_batch_with_element_flags_update` batch — V4 yields `{default: {3: v₃, 5: v₅, UNKNOWN: k₃ + k₅}}`, 302 of 302 added bytes; V3 pinned to the empty legacy map, 0 of 302. The guard's default-to-legacy and nest/restore behaviour is pinned too. The requirement that aggregation *outside* a GroveDB call (Drive's cost sums) install the guard to get the corrected arithmetic is now documented in `docs/crates/costs.md` as well as in the PR body.

Local runs: `cargo test -p grovedb-costs`, `-p grovedb-version`, `-p grovedb-merk`, `-p grovedb --all-features` (3182 passed; the `full,estimated_costs` subset does not compile on current `develop` without `unsafe-dump-load`, unrelated to this PR), `cargo fmt --all`, `cargo clippy --workspace --all-features -- -D warnings`.

**Why it was deferred (unchanged).** The three buggy arms only drop the default-identifier (`[0; 32]`) section; identity-owned sections, and therefore identity fee refunds, are never touched in any shipped version. Drive reaches the buggy arms only when an unflagged `Basic` removal aggregates with a no-owner-flagged removal, and both amounts land in `FeeResult.removed_bytes_from_system`, which nothing in Platform consumes for fees, refunds, pool credits, or state. So the defect changes no consensus-relevant figure today; the fix is hygiene that carries a thread-local version selector into `grovedb-costs` and leaves the Drive-side aggregation gap noted above. The PR is now rebased, fully pinned and CI-green, so it can be merged at zero marginal cost alongside the next V4 cost-gate sweep, or when Platform starts consuming `removed_bytes_from_system`; otherwise it can be closed with an entry in `docs/audit-non-issues.md`.

PR #728, opened 2026-05-20
fix(merk): reject overlong keys before encoding

### LINKED ISSUE #681: [audit][low] Raw Merk accepts overlong keys that encode incorrectly

## Summary
The public `grovedb_merk::Merk` API accepts arbitrary key lengths while link/proof encoding stores key length as `u8` after debug-only checks.

## Impact / failure scenario
Direct Merk insertion with a 256+ byte key can truncate/wrap encoded key lengths in release builds, corrupting storage/proofs.

## References
- `merk/src/merk/apply.rs:241`
- `merk/src/tree/mod.rs:141`
- `merk/src/tree/link.rs:410`
- `grovedb-query/src/proofs/encoding.rs:35`

## Suggested fix
Enforce `key.len() <= u8::MAX` at Merk insertion/tree construction and return an encode error rather than relying on `debug_assert!`.

## Suggested tests
Add direct Merk tests inserting 256-byte and 300-byte keys, then commit/reopen/prove.

QuantumExplorer:
## Summary
- reject raw Merk apply batches containing keys longer than 255 bytes before mutating the tree
- make Merk link encoding return an error for overlong keys instead of relying on debug assertions
- make grovedb-query proof op encoding use checked key-length conversion for every key-bearing variant

Fixes #681.

## Verification
- cargo test -p grovedb-merk apply_rejects_overlong_keys --lib
- cargo test -p grovedb-merk encode_link_long_key --lib
- cargo test -p grovedb-query long_key --lib
- cargo test -p grovedb-merk apply_rejects_overlong_keys --lib --release
- cargo test -p grovedb-merk encode_link_long_key --lib --release
- cargo test -p grovedb-query long_key --lib --release
- cargo test -p grovedb-merk -p grovedb-query

### DIFF
--- a/grovedb-query/src/proofs/encoding.rs
+++ b/grovedb-query/src/proofs/encoding.rs
@@ -15,12 +15,18 @@ use crate::{
 /// allocations.
 const MAX_VALUE_LEN: u32 = 64 * 1024 * 1024;
 
+fn invalid_key_length_error() -> EdError {
+    EdError::IOError(std::io::Error::new(
+        std::io::ErrorKind::InvalidData,
+        "key length must be at most 255 bytes",
+    ))
+}
+
+fn key_len_u8(key: &[u8]) -> ed::Result<u8> {
+    u8::try_from(key.len()).map_err(|_| invalid_key_length_error())
+}
+
 impl Encode for Op {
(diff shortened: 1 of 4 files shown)


QuantumExplorer:
Reviewed

PR #729, opened 2026-05-20
fix(bulk-append): reject short completed chunk proofs

### LINKED ISSUE #688: [audit][high] Bulk chunk proofs can verify incomplete chunk results

## Summary
`BulkAppendTreeProof::verify_against_query` checks that a required chunk index is proved but does not require the authenticated chunk blob to contain exactly `chunk_item_count` entries.

## Impact / failure scenario
A state root over a short chunk blob can verify while a query for later positions in that chunk returns no value instead of an error.

## References
- `grovedb-bulk-append-tree/src/proof/mod.rs:477`
- `grovedb-bulk-append-tree/src/proof/mod.rs:518`
- `grovedb-bulk-append-tree/src/proof/mod.rs:526`
- `grovedb-bulk-append-tree/src/proof/mod.rs:604`

## Suggested fix
Reject completed chunks whose decoded length is not exactly `chunk_item_count`.

## Suggested tests
Forge a one-chunk proof for `height=2,total_count=4` where chunk 0 contains one serialized entry, query position 3, and assert verification fails.

QuantumExplorer:
## Summary
- validate every verified completed chunk blob has exactly the expected chunk item count
- reject out-of-range chunk indices before exposing verified proof results
- add a forged one-leaf proof regression where a short authenticated chunk would otherwise verify and omit queried positions

Fixes #688.

## Verification
- cargo test -p grovedb-bulk-append-tree test_verify_against_query_rejects_short_completed_chunk_blob
- cargo test -p grovedb-bulk-append-tree verify_against_query
- cargo test -p grovedb-bulk-append-tree

### DIFF
--- a/grovedb-bulk-append-tree/src/proof/mod.rs
+++ b/grovedb-bulk-append-tree/src/proof/mod.rs
@@ -405,25 +405,13 @@ impl BulkAppendTreeProof {
 
         if validate_completed_chunk_lengths {
             for (chunk_idx, blob) in &chunk_blobs {
-                if *chunk_idx >= completed_chunks {
-                    return Err(BulkAppendError::InvalidProof(format!(
-                        "chunk {} is out of range for {} completed chunks",
-                        chunk_idx, completed_chunks
-                    )));
-                }
-
                 let entries = deserialize_chunk_blob(blob).map_err(|e| {
                     BulkAppendError::CorruptedData(format!(
                         "failed to deserialize chunk blob {}: {}",
                         chunk_idx, e
                     ))
                 })?;
-                let entry_count = u64::try_from(entries.len()).map_err(|_| {
-                    BulkAppendError::InvalidProof(format!(
-                        "chunk {} entry count does not fit in u64",
-                        chunk_idx
-                    ))
-                })?;
+                let entry_count = entries.len() as u64;
                 if entry_count != chunk_item_count {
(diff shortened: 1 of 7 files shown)


QuantumExplorer:
This is Claude (AI-assisted review).

Closing in favor of #740, which adds a code comment instead of the runtime check.

Rationale: the completed-chunk-length validation here is cryptographically redundant. A completed chunk's contents are already bound to the state root (the MMR recomputes `leaf_hash(blob)` from the blob bytes → `mmr_root` → `compute_state_root`), so a short/truncated/tampered blob produces a different computed root. Every caller binds that root to a trusted value — `verify`/`verify_against_query` compare against `expected_state_root`, and the GroveDB path folds it into the parent Merk hash up to the consensus root and also runs an independent completeness check. So a bad blob is rejected by the root comparison, never reaching a length check (the existing `test_verify_and_compute_root_rejects_corrupted_chunk_leaf` already shows internally-consistent-but-wrong data is caught by the root comparison).

It only differs from the genuinely-needed `dense_count == 0` empty-buffer check because that one guards *unbound* data (unused buffer bytes aren't hashed into the root); completed chunk bytes are bound. The only case the length check would help is a caller that trusts `total_count` but verifies against an attacker-chosen root, which is outside the trust model. BulkAppendTree/CommitmentTree are also unreleased, so there's no live exposure regardless.

#740 documents this at the verification site so future audits don't re-flag it.

PR #731, opened 2026-05-20
fix(batch): reject append conflicts with keyed ops

### LINKED ISSUE #685: [audit][high] Batch append ops can conflict with keyed ops for the same tree

## Summary
A batch can include append-only ops such as `MmrTreeAppend` plus keyed insert/replace/patch operations for the same tree key.

## Impact / failure scenario
Preprocessing can stage non-Merk data and emit a root replacement, then later continuation can overwrite the same `(path, key)`, leaving orphaned staged data or a root that ignores the caller’s keyed op.

## References
- `grovedb/src/batch/mod.rs:1185`
- `grovedb/src/batch/batch_structure.rs:194`
- `grovedb/src/operations/mmr_tree.rs:439`

## Suggested fix
Reject any keyless append op that targets the same `(parent_path, tree_key)` as any keyed op, for MMR, bulk append, dense, and commitment trees.

## Suggested tests
Add both-order batch tests asserting an error and unchanged root/storage.

QuantumExplorer:
## Summary
- broaden append-tree preflight conflicts from delete-only to any keyed op targeting the same tree element
- cover MMR, bulk append, dense, and commitment tree append-style operations in both operation orders
- add an apply_batch regression that confirms the root is unchanged when the conflicting batch is rejected

Fixes #685.

## Verification
- cargo test -p grovedb append_keyed_conflict --lib
- cargo test -p grovedb batch_unit_tests --lib

### DIFF
--- a/grovedb-version/src/version/grovedb_versions.rs
+++ b/grovedb-version/src/version/grovedb_versions.rs
@@ -50,7 +50,6 @@ pub struct GroveDBApplyBatchVersions {
     pub apply_batch_with_element_flags_update: FeatureVersion,
     pub apply_partial_batch_with_element_flags_update: FeatureVersion,
     pub estimated_case_operations_for_batch: FeatureVersion,
-    pub validate_append_conflicts_with_keyed_operations: FeatureVersion,
 }
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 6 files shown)


QuantumExplorer (commented):
Reviewed

PR #733, opened 2026-05-20
fix(batch): enforce reference hop limits for existing refs

### LINKED ISSUE #687: [audit][high] Batch reference fast path can persist an intermediate reference hash

## Summary
The batch reference fast path can return a target reference’s stored value hash without decoding it when the new reference has `max_hop = 1`.

## Impact / failure scenario
A new reference can commit with a combined hash for an intermediate reference rather than the terminal element.

## References
- `grovedb/src/batch/mod.rs:1404`
- `grovedb/src/batch/mod.rs:1410`
- `grovedb/src/batch/mod.rs:1421`
- `grovedb/src/batch/mod.rs:1448`

## Suggested fix
Decode the target even on the one-hop path, or return `ReferenceLimit` if the final allowed hop lands on another reference.

## Suggested tests
Create an on-disk reference chain, insert a new `max_hop=1` reference in a batch, and assert it errors or hashes the terminal element.

QuantumExplorer:
## Summary
- remove the hop-one batch reference shortcut that borrowed the target value hash without decoding it
- resolve hop-one on-disk targets through the existing element-aware path so terminal references hit ReferenceLimit
- add a regression for a one-hop batch reference pointing at an existing reference

Fixes #687.

## Verification
- cargo test -p grovedb test_batch_ref_hop_one_to_existing --lib
- git diff --check

### DIFF
--- a/grovedb-version/src/version/grovedb_versions.rs
+++ b/grovedb-version/src/version/grovedb_versions.rs
@@ -50,6 +50,7 @@ pub struct GroveDBApplyBatchVersions {
     pub apply_batch_with_element_flags_update: FeatureVersion,
     pub apply_partial_batch_with_element_flags_update: FeatureVersion,
     pub estimated_case_operations_for_batch: FeatureVersion,
+    pub validate_reference_hop_limits: FeatureVersion,
 }
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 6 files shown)


QuantumExplorer:
I think we don't want this one.

QuantumExplorer:
Closing: the hop=1 fast path this changes is documented intentional behavior, not an oversight. The 18-line comment added in #667 (merged 2026-05-17, before issue #687 and this PR were filed) explicitly declares `max_hop=1` pointing at another Reference to be ill-formed input that is out of scope for the fast path, caught by `verify_grovedb`, with the decode cost deliberately skipped. The path is also unreachable through Drive: hop=1 references (document index refs, contested doc refs) always target Item primary storage, and keep-history uses hop=2 through the latest-version reference, which takes the decode path before the final hop. As written the PR would additionally flip cost behavior inside the already-in-use GROVE_V3 rather than introducing a new version. If we ever want to revisit the documented contract, that should happen as a design discussion on #687, not via this patch.

Issue #686, opened 2026-05-20
[audit][medium] Non-empty tree delete can reopen parent with child tree type


QuantumExplorer:
Created from a Codex audit of grovedb. No code changes were made as part of the audit.

## Summary
Direct delete of a non-empty child tree can reopen the parent path using the child tree’s `tree_type`.

## Impact / failure scenario
Deleting under `CountTree`/sum/count-sum parents with `allow_deleting_non_empty_trees=true` can propagate aggregate/root metadata incorrectly.

## References
- `grovedb/src/operations/delete/mod.rs:740`
- `grovedb/src/operations/delete/mod.rs:750`
- `grovedb/src/operations/delete/mod.rs:847`
- `grovedb/src/operations/delete/mod.rs:880`

## Suggested fix
Reopen with the parent’s saved tree type or reuse the already-open parent subtree handle.

## Suggested tests
Delete a non-empty child under an aggregate parent, then verify aggregate count/sum and `verify_grovedb`.

---

## September 2026 audit addendum — M014

Audit group: **M014**. Classification: **correctness**. Provisional severity: **medium**.

The reviewed snapshot's populated-child deletion reopens the parent as a layered Merk using the deleted child's tree type. Besides the known link-hash/type problem, this can lose base-root pointer maintenance. Empty-child deletion retains the original parent and avoids this route. This is an addendum to the existing #686 root cause. GitHub metadata marks #732 merged; its description reuses the parent under V4 and deliberately preserves V1-V3 behavior. The audit does not establish that the described V4 fix is incomplete, nor that current develop remains affected.

Expected behavior: Deleting a populated child must retain the parent's actual Merk mode and tree type, including base-root pointer maintenance.

Fix direction: Reuse the already opened parent or reopen it through the canonical parent-path helper preserving both MerkType and TreeType. Cover surviving siblings at the base root and nested parent/child type differences. Version persistent root changes where replay compatibility requires it.

### Validation to complete

- Suggested: verify the described V4 parent-reuse fix on its actual target before proposing further remediation.
- Cover base-root deletion with surviving siblings and nested differing parent/child types; preserve legacy replay expectations.

### Limits and existing work

- PR #732 is marked merged and explicitly keeps legacy V1-V3 behavior; no new current-develop or V4 regression claim is made.

Related tracking: [issue #686](https://github.com/dashpay/grovedb/issues/686) (closed), [PR #732](https://github.com/dashpay/grovedb/pull/732) (merged).

[...]

PR #736, opened 2026-05-20
fix(merk): validate fetched links during lazy load

### LINKED ISSUE #682: [audit][low] Lazy-loaded Merk references do not validate fetched node metadata

## Summary
When a `Link::Reference` is loaded, the fetched node is trusted. The key check is debug-only and hash/aggregate/height metadata are not validated against the reference.

## Impact / failure scenario
Corrupt storage or a custom `Fetch` implementation can return a stale or wrong node; release builds can continue with inconsistent link metadata and loaded tree contents.

## References
- `merk/src/tree/mod.rs:1497`
- `merk/src/tree/walk/mod.rs:90`
- `merk/src/merk/source.rs:51`

## Suggested fix
Validate fetched key, `hash_for_link(tree_type)`, aggregate data, and child heights before converting a reference link into a loaded node.

## Suggested tests
Corrupt a referenced child record after commit and assert `get`/`prove` fails at load time.

QuantumExplorer:
## Summary
- expose the tree type through Fetch so lazy loads can validate fetched children against pruned link metadata
- validate fetched key, child heights, aggregate data, and tree-type-aware hash in TreeNode::load and Walker::detach
- add Merk walk regressions for wrong-key and wrong-hash fetched children, including the read-only RefWalker path

Fixes #682.

## Verification
- cargo test -p grovedb-merk walk_pruned --lib
- cargo test -p grovedb-merk tree::walk::test --lib
- git diff --check

### DIFF
--- a/merk/src/merk/restore.rs
+++ b/merk/src/merk/restore.rs
@@ -599,7 +599,10 @@ impl<'db, S: StorageContext<'db>> Restorer<S> {
                 .ok_or(Error::ChunkRestoringError(ChunkError::InternalError(
                     "tree is None in rewrite_heights",
                 )))?;
-        let walker = RefWalker::new(&mut tree, self.merk.source());
+        let walker = RefWalker::new(
+            &mut tree,
+            self.merk.source_without_child_height_validation(),
+        );
 
         rewrite_child_heights(walker, &mut batch, grove_version)?;
 
--- a/merk/src/merk/source.rs
+++ b/merk/src/merk/source.rs
@@ -16,6 +16,15 @@ where
         MerkSource {
             storage: &self.storage,
             tree_type: self.tree_type,
+            validate_child_heights: true,
+        }
+    }
+
+    pub(in crate::merk) fn source_without_child_height_validation(&self) -> MerkSource<'_, S> {
+        MerkSource {
+            storage: &self.storage,
+            tree_type: self.tree_type,
+            validate_child_heights: false,
         }
     }
 }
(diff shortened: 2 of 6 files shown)


QuantumExplorer:
This is Claude, closing per maintainer decision after re-review against current develop (2026-08-22).

Rationale: the validation is optional hardening with no attacker vector — the fetch reads the node's own RocksDB snapshot, and clients are already protected because proof verification recomputes every hash against the root. What it would buy (fail-loud on local corruption, catching mislabeled-tree-type bugs at the fetch site) turned out to carry real costs on today's develop:

- It changes shipped consensus behavior, so it can no longer land ungated: the legacy non-empty-child delete (#686, `delete_internal_on_transaction` v0, GROVE_V1..V3) commits mismatched link hashes that replay must reproduce — with this validation active those v3 paths error instead (the v3 regression tests catch exactly that). It would need a `merk_versions.tree.validate_fetched_links` V4 gate.
- Coverage is partial: `Merk::get` on a pruned node goes through the direct by-key lookup and bypasses the walker validation entirely, so the most common trusted-read path isn't covered — only apply/prove/walk.
- Its per-tree-type compatibility match is a standing maintenance point: it predates the indexed-tree primaries (#657) and `PrivateDocumentStore` (#787), and as written would hard-error every lazy load inside those Merks.
- If any #686-corrupted links exist in historical state, validation turns them from latent into fatal-on-touch once active.

Issue #682 stays open as a known low-severity finding. `verify_grovedb` remains the tool for storage-integrity sweeps. If this is ever revisited, the cheaper shape is debug-only validation (extending the existing `debug_assert` on the key) — that catches #686-class internal bugs in every test run with zero production behavior change and no version gate.

PR #737, opened 2026-05-21
fix(costs): check size arithmetic for overflow

### LINKED ISSUE #684: [audit][high] Unchecked size arithmetic can undercount storage costs

## Summary
Several element sizing and cost-accounting paths use unchecked `u32` arithmetic or unchecked casts.

## Impact / failure scenario
Near-`u32::MAX` lengths can wrap in release or panic in debug, causing undercharged storage costs or verifier bypasses.

## References
- `grovedb-element/src/element/helpers.rs:622`
- `grovedb-element/src/element/helpers.rs:647`
- `grovedb-element/src/element/helpers.rs:681`
- `costs/src/lib.rs:194`
- `costs/src/storage_cost/key_value_cost.rs:64`
- `grovedb/src/batch/mod.rs:2945`

## Suggested fix
Use `checked_add`, `checked_sub`, and `u32::try_from` with explicit overflow errors.

## Suggested tests
Add boundary tests around `u32::MAX`, `u32::MAX - n`, and oversized serialized values.

QuantumExplorer:
## Summary
- add checked arithmetic to element required-space helpers and cost verification paths
- return overflow errors from root key storage-cost sizing instead of wrapping
- check batch JIT value-defined cost sums before returning Merk costs
- add overflow regressions for element sizing and cost storage arithmetic

Fixes #684.

## Verification
- cargo test -p grovedb-element --test element_constructors_helpers
- cargo test -p grovedb-costs --test coverage_regression
- cargo test -p grovedb test_batch_replace_item_with_sum_item_flags_update --lib
- git diff --check

## Release Notes

* **Bug Fixes**
  * Storage cost calculations now properly detect and reject arithmetic overflow conditions instead of silently wrapping or saturating values, providing clearer error reporting.

* **Tests**
  * Added comprehensive overflow detection tests to verify the system correctly handles edge cases in cost computations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/grovedb/pull/737?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/costs/src/error.rs
+++ b/costs/src/error.rs
@@ -31,6 +31,10 @@ use crate::StorageCost;
 /// An Error coming from costs
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    /// Arithmetic overflow
+    #[error("overflow error: {0}")]
+    Overflow(&'static str),
+
     /// Storage Cost Value mismatch
     #[error("storage_cost cost mismatch added: {0} replaced: {1} actual:{actual_total_bytes}",
         expected.added_bytes, expected.replaced_bytes)]
--- a/costs/src/lib.rs
+++ b/costs/src/lib.rs
@@ -117,6 +117,19 @@ pub struct OperationCost {
 }
 
 impl OperationCost {
+    fn checked_len_with_required_space(len: u32, context: &'static str) -> Result<u32, Error> {
+        len.checked_add(len.required_space() as u32)
+            .ok_or(Error::Overflow(context))
+    }
+
+    fn checked_add(left: u32, right: u32, context: &'static str) -> Result<u32, Error> {
+        left.checked_add(right).ok_or(Error::Overflow(context))
+    }
+
+    fn checked_sub(left: u32, right: u32, context: &'static str) -> Result<u32, Error> {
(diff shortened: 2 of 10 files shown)


QuantumExplorer:
This is Claude. Closing this PR after review, alongside its issue #684 (closed earlier as not a reachable bug — the additions can only overflow at ~4 GiB inputs, which cannot exist under the platform's key/value size caps).

Beyond fixing a non-issue, this PR is not the pure hygiene it appears to be: it converts several `saturating_sub` calls to erroring `checked_sub` in the `paid_value_len` element-cost computation (child options, left/right child lengths, sum lengths, sum tree length). Saturate-and-continue and error-out are only equivalent if those subtractions can never underflow — an unverified structural claim across ~6 sites in fee-critical arithmetic. With GROVE_V3 live in production, any case where that claim is wrong would change fees/op outcomes between patched and unpatched nodes on the same protocol version.

Since the benefit is zero (the overflow is unreachable) and the risk is consensus-grade, closing is the right trade. If checked-arithmetic hygiene is wanted later, it should be narrowed to the addition sites only (behavior-identical below 4 GiB by construction), keep saturating semantics on the subtractions, and land at the start of a protocol-version window — not in a patch.

PR #738, opened 2026-05-21
fix(element): version trailing-byte deserialization rejection

QuantumExplorer:
## Summary
- gate element trailing-byte deserialization rejection behind element.deserialize version 1
- keep legacy version 0 behavior for old protocol versions
- add regression coverage for the legacy and latest behaviors

## Tests
- cargo test -p grovedb-element deserialize_ --lib
- cargo check -p grovedb-element --lib
- git diff --check

## Bug Fixes
* Improved legacy element deserialization to properly handle version-specific validation rules, allowing trailing bytes in legacy data formats.
* Enhanced protocol v3 version definitions to enable element deserialization support.

## Tests
* Added regression test for legacy element deserialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/grovedb/pull/738?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/grovedb-element/src/element/serialize.rs
+++ b/grovedb-element/src/element/serialize.rs
@@ -2,7 +2,9 @@
 //! Implements serialization functions in Element
 
 use bincode::config;
-use grovedb_version::{check_grovedb_v0, version::GroveVersion};
+use grovedb_version::{
+    check_grovedb_v0, check_grovedb_v0_or_v1, error::GroveVersionError, version::GroveVersion,
+};
 
 use crate::{
     element::Element,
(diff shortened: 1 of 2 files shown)


QuantumExplorer:
Closing: the divergence this version-gates is unreachable on canonical data. Every GroveDB write path serializes elements via canonical bincode in `Element::serialize`, and replication/chunk restore is root-hash-verified against canonically built state, so no stored element value can ever carry trailing bytes. That means historical replay under GROVE_V1/V2 can never encounter an input where lenient-vs-strict decoding differs, and the unconditional rejection introduced in #724 cannot cause node divergence. Gating it would add a permanent legacy branch guarding a state no honest node can reach. The strict behavior from #724 stands as-is.

PR #739, opened 2026-05-21
fix(proofs): version trailing-byte proof decoding

QuantumExplorer:
## Summary
- route GroveDB proof-envelope decoding through a proof-version trailing-byte policy
- set v3 to reject trailing proof bytes while preserving v1/v2 legacy decoding
- add regression coverage for latest rejection and legacy acceptance

## Tests
- cargo test -p grovedb verify_query_rejects_proof_with_trailing_bytes --lib
- cargo test -p grovedb verify_query_legacy_version_accepts_proof_with_trailing_bytes --lib
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Implemented version-aware proof validation that conditionally enforces strict rejection of trailing bytes in proofs based on the configured GroveDB version. Legacy versions accept proofs with trailing bytes, while newer versions enforce stricter validation rules.

* **Tests**
  * Added regression test to verify backward compatibility for legacy GroveDB versions when validating proofs with trailing bytes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/grovedb/pull/739?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/grovedb-version/src/version/grovedb_versions.rs
+++ b/grovedb-version/src/version/grovedb_versions.rs
@@ -105,6 +105,7 @@ pub struct GroveDBOperationsProofVersions {
     pub verify_subset_query_with_absence_proof: FeatureVersion,
     pub verify_query_with_chained_path_queries: FeatureVersion,
     pub verify_query_get_parent_tree_info_with_options: FeatureVersion,
+    pub decode_rejects_trailing_bytes: FeatureVersion,
 }
 
 #[derive(Clone, Debug, Default)]
--- a/grovedb-version/src/version/v1.rs
+++ b/grovedb-version/src/version/v1.rs
@@ -164,6 +164,7 @@ pub const GROVE_V1: GroveVersion = GroveVersion {
                 verify_subset_query_with_absence_proof: 0,
                 verify_query_with_chained_path_queries: 0,
                 verify_query_get_parent_tree_info_with_options: 0,
+                decode_rejects_trailing_bytes: 0,
             },
             average_case: GroveDBOperationsAverageCaseVersions {
                 add_average_case_get_merk_at_path: 0,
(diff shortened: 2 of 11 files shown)


QuantumExplorer:
This is Claude. Closing this PR after re-triage for the GROVE_V4 boundary — both halves of it are now moot, and the gating half would be actively harmful to land.

**The strictness is already merged and shipped.** PR #661 (May 2026) introduced `decode_grovedb_proof_canonical`, which unconditionally rejects trailing bytes after the encoded envelope; every envelope entry point routes through it (`verify_query`, `verify_query_raw`, `verify_query_with_options`, trunk-chunk, `verify_path_query`, all aggregate verifiers). Confirmed on current develop under every grove version v1–v4, with 10 `trailing_bytes` regression tests passing.

**There was never a lenient GROVE_V3 verifier in production**, so the legacy-acceptance side of this PR's gate has no era to preserve:
- grovedb v4.1.0 — the last release that accepted trailing bytes — does not contain GROVE_V3 at all, so v3 could not have activated on it.
- Every Dash Platform v4.0.0 build from beta.1 (2026-06-02) onward pins grovedb revs that include the strict decode; platform v4.0.0 final pins grovedb v5.0.0, and current v4.1.1 pins v5.0.1 — all strict.
- GROVE_V3 activated ~mid-June 2026, necessarily on one of those strict builds, so strict rejection has been the live v3 behavior since the first v3 block.

Landing this PR (or a rebase of it re-gated to v4) would therefore *loosen* live verifiers for v1/v2/v3 and create exactly the intra-version divergence a version gate exists to prevent — a node on this code would accept malleated proofs every platform ≥4.0.0 node rejects.

This determination is recorded in `docs/audit-non-issues.md` via #820. It should not be re-proposed unless evidence emerges that some production verifier ran a lenient grovedb (< v5.0.0) under GROVE_V3.

PR #752, opened 2026-05-25
feat: subtree dump/restore primitives

shumkov:
## Summary

Three escape-hatch APIs for caller-driven subtree dump/restore, all gated behind a new opt-in Cargo feature `unsafe-dump-load`. Off by default so production builds carry none of this code.

| API | Where | What it does |
|---|---|---|
| `GroveDb::raw_storage()` | `grovedb/src/lib.rs` | Borrow the underlying `RocksDbStorage` for direct use of the `Storage` trait (raw iter, low-level reads). |
| `RocksDbStorage::ingest_subtree_sst(cf, sst_path)` | `storage/src/rocksdb_storage/storage.rs` | Bulk-load a precomputed SST into a column family via RocksDB `IngestExternalFile`. Pinned `allow_global_seqno=false` + `snapshot_consistency=false`. |
| `GroveDb::replace_subtree_root(path, key, element, combined_root, …)` | `grovedb/src/operations/replace_subtree_root.rs` | Replace a subtree leaf's child hash + element data in its parent Merk with caller-provided values, **without** re-reading the subtree to recompute the hash. |

## What changed since the earlier "[DON'T MERGE]" shape

- Merged with `develop` (PR #751's `append_many_raw` now flows in via develop; this branch's net delta is just the bootstrap surface).
- Renamed `replace_commitment_tree_subtree_root` → `replace_subtree_root` and made it generic — takes an `Element` instead of `(total_count, chunk_power, flags)`. The parent-Merk plumbing is identical across all non-Merk tree types (CommitmentTree / Mmr / BulkAppend / Dense), so a single helper covers all of them. Callers build whichever Element variant they're restoring.
- Vestigial `BulkAppendTree::append_many` helper dropped (PR #751 inlined the loop into `append_many_raw` directly; no callers).
- All three APIs `#[cfg(feature = "unsafe-dump-load")]`-gated. `storage` crate exposes `unsafe-dump-load = ["rocksdb_storage"]`; `grovedb` crate exposes `unsafe-dump-load = ["grovedb-storage/unsafe-dump-load"]`.

## Safety contract (also in module-level doc)

`replace_subtree_root` is documented as **caller-responsible**:
- Caller must ensure `new_combined_root` matches the underlying subtree state (typical pattern: ingest SST, reopen typed tree, recompute the post-ingest root, then call this).
- Mismatches produce a silently-inconsistent Merk tree — the parent's recorded child hash diverges from actual subtree state.
- Runtime check is a cheap "is this a tree variant at all" guard; reject Item/Reference. Hash-vs-state correctness is the caller's problem.

`ingest_subtree_sst` bypasses any open transaction. Caller must arrange transaction semantics at a higher layer (in dashpay/platform we only call it when the destination subtree is known empty at InitChain time, and rely on InitChain abort = wipe-and-restart for failure recovery).

`raw_storage` is documented as an escape hatch with no stability guarantee.

## Use case

Devnet shielded-pool genesis snapshot bake/apply in dashpay/platform (the consumer):
1. Bake stage runs `drive-abci snapshot-bake` → generates 30k–1M-note commitment-tree subtree → dumps to SST via `raw_storage().raw_iter()` + `SstFileWriter`.
2. Runtime InitChain reads the embedded SST, ingests via `ingest_subtree_sst`, recomputes the combined root via `compute_commitment_tree_state_root`, then `replace_subtree_root` to patch the parent Merk leaf.

Roundtrip is verified on the consumer side: 30 000 notes → 8.6 MB SST → re-applied to a fresh platform yields a byte-equivalent Sinsemilla anchor (`df37726e…4c2300`).

## Verification

- `cargo check -p grovedb --no-default-features --features minimal` — default (feature off), clean.
- `cargo check -p grovedb --no-default-features --features "minimal,unsafe-dump-load"` — feature on, clean.
- All 1852 grovedb library tests pass.
- CI on this branch is fully green except `codecov/patch` (0% coverage on the 109 new bootstrap-API lines — real gap; coverage is intentionally not raised inside grovedb because the realistic exercise of these APIs is dump+restore at the consumer level, not in-grovedb).

## Stability note

These three APIs are explicitly opt-in escape hatches, not stable public surface. Misuse is silent corruption. The `unsafe-dump-load` feature name is the signal; the docstrings reinforce it.

## Summary by CodeRabbit

* **New Features**
  * Opt-in "unsafe-dump-load" feature to enable out-of-band dump/restore tooling.
  * Read-only access to underlying storage for low-level snapshot/replication workflows.
  * Bulk SST ingest to efficiently load subtree data into the storage engine.
  * Caller-driven subtree-root replacement to support snapshot-based bootstrap scenarios.

* **Tests**
  * Roundtrip test verifying dump→ingest restores identical subtree root hash.
  * Validation test ensuring non-tree replacements are rejected.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,7 @@
+[advisories]
+ignore = [
+    # Reviewed in https://github.com/dashpay/grovedb/issues/722.
+    # The bincode 2.0.1 unmaintained advisory is intentionally accepted for
+    # this workspace; decoding canonicality is tracked separately.
+    "RUSTSEC-2025-0141",
+]
--- a/grovedb-bulk-append-tree/src/proof/mod.rs
+++ b/grovedb-bulk-append-tree/src/proof/mod.rs
@@ -392,6 +392,36 @@ impl BulkAppendTreeProof {
             (root, verified_leaves)
         };
 
+        // NOTE — intentionally NO "completed chunk length" check here.
+        //
+        // An audit (human or AI) may expect us to deserialize each completed
+        // chunk blob and reject any that does not hold exactly
+        // `chunk_item_count` entries (the concern raised in issue #688). That
+        // check is cryptographically redundant and is deliberately omitted; do
+        // not re-add it.
+        //
+        // A completed chunk's contents are fully bound to the root: the MMR
+        // recomputes `leaf_hash(blob)` from the blob bytes, that leaf feeds
+        // `mmr_root`, and `mmr_root` feeds `compute_state_root` below. So a
+        // short, truncated, or otherwise tampered chunk blob yields a
+        // *different* `computed_state_root`. Every caller binds that root to a
+        // trusted value — `verify` / `verify_against_query` compare it against
+        // `expected_state_root`, and the GroveDB lower-layer path folds it into
+        // the parent Merk hash up to the consensus root — so a bad blob is
+        // rejected by the root comparison, never reaching a length check. (The
+        // GroveDB path additionally runs an independent completeness check that
+        // rejects any queried position the proof omits.)
+        //
+        // Contrast with the `dense_count == 0` empty-buffer check below, which
+        // *is* required: unused buffer data is NOT hashed into the root, so it
+        // would be unbound and could be smuggled. Completed chunk bytes are
+        // bound, so they cannot be.
+        //
+        // The only situation a length check would add anything is a caller that
+        // trusts `total_count` yet verifies against an attacker-chosen root —
+        // which is outside the trust model (such a caller can be handed an
+        // entirely different, internally consistent tree). See PR #729 (closed).
+
         // 2. Verify dense tree buffer sub-proof
         let (dense_root, dense_entries) = if dense_count > 0 {
             self.buffer_proof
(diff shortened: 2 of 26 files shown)


shumkov:
Re: codecov 0% patch coverage —

Addressed in 62e1296a via two in-tree tests in `grovedb/src/tests/commitment_tree_tests.rs`:

- `unsafe_dump_load_subtree_roundtrip_preserves_root_hash` — full round-trip exercising all three new APIs: builds a CommitmentTree on GroveDb A, dumps the subtree to an SST via `raw_storage` + `SstFileWriter`, ingests it into a fresh GroveDb B via `ingest_subtree_sst`, patches the parent leaf via `replace_subtree_root`, and asserts the post-restore `root_hash` matches A byte-for-byte.

- `replace_subtree_root_rejects_non_tree_element` — pins the cheap guard rejecting Item / Reference elements (which have no child-hash slot in the parent Merk).

Both are gated on `#[cfg(feature = "unsafe-dump-load")]`; CI's `cargo llvm-cov nextest --all-features` picks them up automatically.

The consumer-level test on platform (`snapshot_dump_apply_preserves_anchor` in dashpay/platform#3774) exercises the same path under realistic scale (30k notes, header + checksum).

QuantumExplorer (approved):

PR #757, opened 2026-06-04
fix(insert): restore v4.1.0 non-batch dispatch for count/sum trees (protocol-v11 consensus)

shumkov:
## What

Restore the grovedb **v4.1.0** non-batch insert dispatch for `CountSumTree`, `ProvableCountTree` and `ProvableCountSumTree` so they are written via the plain-value path (`Op::Put`) instead of the layered-subtree path (`Op::PutLayeredReference`).

## Why — consensus break at protocol v11

`GroveDb::add_element_on_transaction` (the non-batch insert path used by `grove_insert_if_not_exists`) decides whether an element is inserted as a **layered subtree** (`Op::PutLayeredReference`) or a **plain value** (`Op::Put`).

- In **v4.1.0** (`86562f65`), `CountSumTree` / `ProvableCountTree` / `ProvableCountSumTree` fall through to `_ => element.insert()` → **`Op::Put`**.
- **PR #752** broadened the tree arm to include them → **`Op::PutLayeredReference`**.

The two ops compute a different parent-node `value_hash`:

```
Op::Put                : value_hash = value_hash(serialized)
Op::PutLayeredReference: value_hash = combine_hash(value_hash(serialized), child_root_hash)
                                    = combine_hash(value_hash(serialized), NULL_HASH)   // empty tree
```

That changes the parent node hash → the grovedb root. These tree types are created on the **protocol-v11 activation** (Dash Platform `transition_to_version_11` inserts an `empty_provable_count_sum_tree` and an `empty_count_sum_tree`), so the broadened dispatch makes a v11 node compute a **different app-hash than the v4.1.0-era binaries that produced the canonical chain** — a consensus divergence on replay.

### Evidence (testnet block 245,344, the v10→v11 activation block)

Replaying the block from the identical committed state, logging the grovedb root after each transition insert:

| step | op | v4.1.0 (canonical) | a18f7929 (regressed) |
|---|---|---|---|
| 00 | committed v10 state | `76d92e…` | `76d92e…` |
| 01 | `empty_sum_tree` @ `[56]` | `4d08e1c3…` | `4d08e1c3…` (same — SumTree is in the tree arm in both) |
| **02** | **`empty_provable_count_sum_tree` @ `[56,'c']`** | **`e0f9eded…`** | **`9b2b19a9…`** (diverges) |

Step 01 (`SumTree`, unchanged) is identical; step 02 (`ProvableCountSumTree`, which changed arms) diverges. All later steps diverge downstream.

## The fix

In `add_element_on_transaction`, move `CountSumTree` / `ProvableCountTree` / `ProvableCountSumTree` out of the layered-subtree arm back to the `Op::Put` arm, exactly matching v4.1.0. `Tree` / `SumTree` / `BigSumTree` / `CountTree` keep the layered behavior (unchanged in both revs). The **v12-only** `ProvableSumTree` / `ProvableCountProvableSumTree` (and `CommitmentTree` / `MmrTree` / `BulkAppendTree` / `DenseAppendOnlyFixedSizeTree`) were never on consensus and keep their current behavior.

This is an **accidental regression** (PR #752 was "subtree dump/restore primitives"); the broadened behavior never reached consensus, so restoring v4.1.0 unconditionally is correct for all live protocol versions.

## Test

`provable_count_sum_tree_insert_preserves_v11_consensus_root` pins the post-insert root so the dispatch cannot silently regress again (`empty_sum_tree` as the unchanged control).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected consensus-critical behavior during tree insertion operations to ensure all element types are handled according to protocol specifications, preventing potential consensus rule violations.

* **Tests**
  * Added comprehensive regression test for consensus behavior, validating root hash generation and protecting against future deviations from expected protocol compliance.

### DIFF
--- a/grovedb/src/operations/insert/mod.rs
+++ b/grovedb/src/operations/insert/mod.rs
@@ -285,13 +285,23 @@ impl GroveDb {
                     )
                 );
             }
+            // CONSENSUS-CRITICAL (protocol v11 / GROVE_V2). `CountSumTree`,
+            // `ProvableCountTree` and `ProvableCountSumTree` MUST be inserted via the
+            // plain-value path (`Op::Put`, in the `Item` arm below), NOT as a layered
+            // subtree. grovedb <= v4.1.0 routed them through `_ => element.insert()`
+            // (`Op::Put`), and that is the behavior frozen into the live v11 activation
+            // chain — e.g. testnet block 245,344's `transition_to_version_11`, which
+            // inserts an `empty_provable_count_sum_tree` (CLEAR_ADDRESS_POOL) and an
+            // `empty_count_sum_tree` (ADDRESS_BALANCES). Routing them through the
+            // layered-subtree arm changes the parent node's `value_hash` from
+            // `value_hash(serialized)` to `combine_hash(value_hash(serialized), NULL_HASH)`,
+            // which changes the grovedb root and breaks consensus on replay. The
+            // v12-only `ProvableSumTree` / `ProvableCountProvableSumTree` were never on
+            // consensus, so they keep the layered behavior here.
             Element::Tree(value, _)
             | Element::SumTree(value, ..)
             | Element::BigSumTree(value, ..)
             | Element::CountTree(value, ..)
-            | Element::CountSumTree(value, ..)
-            | Element::ProvableCountTree(value, ..)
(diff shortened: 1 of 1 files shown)


shumkov:
Closed in favor of #759

PR #759, opened 2026-06-04
fix: version-gate add_element_on_transaction (v0=Op::Put, v1=layered)

QuantumExplorer:
## What

Splits the non-batch insert helper `GroveDb::add_element_on_transaction` into a **versioned dispatch** (Platform-style: dispatcher + frozen `v0`/`v1` snapshots, each in its own file under `grovedb/src/operations/insert/add_element_on_transaction/`):

| GroveVersion | `add_element_on_transaction` slot | behaviour for `CountSumTree` / `ProvableCountTree` / `ProvableCountSumTree` |
|---|---|---|
| `GROVE_V1`, `GROVE_V2` | `0` → **v0** | plain value (`Op::Put`) — matches grovedb v4.1.0 / the live protocol-v11 chain |
| `GROVE_V3` | `1` → **v1** | layered subtree (`Op::PutLayeredReference`) — current behaviour, consistent with the batch path |

Only `GROVE_V3`'s slot is bumped `0 → 1`; `v1`/`v2` stay `0`.

## Why — protocol-v11 consensus

`add_element_on_transaction` decides whether an element is written as a layered subtree (`Op::PutLayeredReference`) or a plain value (`Op::Put`). The two ops compute a different parent-node `value_hash`:

```
Op::Put                : value_hash = value_hash(serialized)
Op::PutLayeredReference: value_hash = combine_hash(value_hash(serialized), child_root_hash)
                                    = combine_hash(value_hash(serialized), NULL_HASH)   // empty tree
```

In grovedb **v4.1.0**, `CountSumTree` / `ProvableCountTree` / `ProvableCountSumTree` fell through `_ => element.insert()` → **`Op::Put`**. **#752** (subtree dump/restore primitives) broadened the tree arm to include them → **`Op::PutLayeredReference`**, changing the parent node hash and therefore the grovedb root.

These types are created on the **protocol-v11 activation** (`transition_to_version_11` inserts an `empty_provable_count_sum_tree` and an `empty_count_sum_tree` via this non-batch path — testnet block 245,344), so the broadened dispatch makes a v11 node compute a **different app-hash than the v4.1.0-era binaries that produced the canonical chain** — a consensus divergence on replay.

## Why versioned instead of an unconditional revert (cf. #757)

#757 reverts the three types to `Op::Put` for all versions. This PR instead **preserves both behaviours** and gates them, so:

- `GROVE_V1` / `GROVE_V2` replay the protocol-v11 chain correctly (`Op::Put`), and
- `GROVE_V3`+ keep the layered behaviour, which makes the **non-batch path agree with the batch path** on both root hash and fee (the batch path has always written these three types as layered, via `insert_subtree_into_batch_operations` + `LayeredValueDefinedCost`). That removes the long-standing batch/non-batch divergence going forward instead of freezing it.

`v0`/`v1` are full frozen snapshots; they differ **only** in which match arm those three element types fall into. `v0` is intentionally "current body with those three arms moved to `Op::Put`" rather than a literal v4.1.0 copy, so it still handles the newer element types (`ProvableSumTree`, `CommitmentTree`, the append-trees, `NonCounted` wrappers) that v4.1.0 predates, while reproducing the v4.1.0 consensus root exactly for every type that existed then.

## Test

`add_element_on_transaction_version_gate_provable_count_sum_tree_root` replays the `transition_to_version_11` shape (`empty_sum_tree` control at `[56]`, then `empty_provable_count_sum_tree` at `[56,'c']`) under both versions and pins both roots:

- **v0 root is byte-identical to PR #757's pinned protocol-v11 golden** — i.e. `GROVE_V1 → v0 → Op::Put` reproduces the exact canonical root.
- **v1 (layered, `GROVE_V3`) root differs**, and the gate is asserted to switch behaviour so it cannot silently collapse.

## Verification

- `cargo build -p grovedb` — clean, no warnings
- `cargo clippy -p grovedb --lib` — clean
- `cargo test -p grovedb --lib` — **1853 passed, 0 failed, 1 ignored**

## Reviewer notes

- `latest() = GROVE_V3 = v1 = layered`, which is today's behaviour, so all `latest()`-based tests are unaffected. The behaviour change is limited to `GROVE_V1` / `GROVE_V2`, which now correctly use `Op::Put` for these three types (the regressed layered behaviour was never on the canonical chain for those versions).
- The v12-only `ProvableSumTree` / `ProvableCountProvableSumTree` keep the layered behaviour in both v0 and v1 — they were never created via this path on the v11 chain.
- Branch is rebased on #758 (the `orchard` 0.14.0 pin) so it builds.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced versioning infrastructure for database insert operations with improved protocol-specific handling to ensure consistent behavior across all supported grove database versions.

* **Tests**
  * Added comprehensive regression tests validating version-specific protocol behaviors for insert operations, including consensus root hash verification to prevent regressions across grove versions.

### DIFF
--- a/grovedb-version/src/version/v3.rs
+++ b/grovedb-version/src/version/v3.rs
@@ -104,7 +104,11 @@ pub const GROVE_V3: GroveVersion = GroveVersion {
                 insert: 0,
                 insert_on_transaction: 0,
                 insert_without_transaction: 0,
-                add_element_on_transaction: 0,
+                // v1: non-batch insert writes CountSumTree / ProvableCountTree /
+                // ProvableCountSumTree as layered subtrees, consistent with the
+                // batch path. GROVE_V1 / GROVE_V2 keep v0 (Op::Put) to preserve
+                // the protocol-v11 consensus root (testnet block 245,344).
+                add_element_on_transaction: 1,
                 add_element_without_transaction: 0,
                 insert_if_not_exists: 0,
                 insert_if_not_exists_return_existing_element: 0,
--- a/grovedb/src/operations/insert/add_element_on_transaction/mod.rs
+++ b/grovedb/src/operations/insert/add_element_on_transaction/mod.rs
@@ -1,0 +1,91 @@
+//! `add_element_on_transaction` — versioned dispatch.
+//!
+//! Adds a single element (an empty subtree or a value) into an already-open
+//! parent merk on a transaction. This is the **non-batch** insert path
+//! (`GroveDb::insert` / `insert_if_not_exists`).
+//!
+//! The choice between writing certain tree types as a layered subtree
+//! (`Op::PutLayeredReference`) versus a plain value (`Op::Put`) changes the
+//! parent node's `value_hash` and therefore the grovedb root, so it is
+//! **consensus-critical** and version-gated:
+//!
+//! * **[v0]** — grovedb v4.1.0 behaviour. `CountSumTree` / `ProvableCountTree` /
+//!   `ProvableCountSumTree` are written as a plain value (`Op::Put`). This is the
+//!   behaviour frozen into the live protocol-v11 activation chain (testnet block
+//!   245,344, `transition_to_version_11`). Selected by `GROVE_V1` / `GROVE_V2`.
+//! * **[v1]** — current behaviour. Those three types are written as layered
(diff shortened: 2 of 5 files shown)


QuantumExplorer (commented):
Self Reviewed

shumkov (approved):

PR #769, opened 2026-07-06
docs: record adversarially-rejected audit findings (audit-non-issues.md)

QuantumExplorer:
## What

Adds `docs/audit-non-issues.md`: a durable record of audit findings that were adversarially verified and **rejected**, so future audits (human, Codex, or Claude) check the list before re-filing. CLAUDE.md gains an "Auditing This Codebase" section pointing at it.

## Why

Seven open issues from the 2026-05-20 Codex audit batch were just closed after verification (#693, #694, #697, #699, #711, #715, #716), joining the earlier #684/#685. Without a checked-in record, the next automated audit will rediscover the same unchecked arithmetic and by-design behavior and re-file them.

The file is organized to reject re-reports **without suppressing new findings**: each entry states the claim, the invariant that makes it unreachable (255-byte key cap, platform size caps, hash-verified counts, transactional batch rollback), and an explicit *"becomes real if"* condition — if that condition holds, filing is correct and expected.

## Contents

- **Unreachable arithmetic overflow class** — #684, #715, #716, #693 (inputs are real stored bytes or hash-committed counts, capped far below wrap points)
- **Capacity-boundary behavior** — #694 (2^32 leaves + transactional rollback)
- **By-design behavior** — #697 (replicate from checkpoints), #699 (fast-checkpoint WAL tradeoff), #711 (deterministic undercount on rejected ops only), plus the long-standing `clear_subtree` and `Link::Modified` decisions
- **Verified-unreachable proof/encoding claims** — key-length truncation (#506 cap), `feature_type` forgery (verifier discards it; canonical data is hash-verified), `saturating_sub` on net sum budgets

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for codebase audits, including where to check before filing security or correctness findings.
  * Added a reference page listing previously reviewed edge cases and when they should be reconsidered.

PR #750, opened 2026-05-24
fix(merk): reject dangling multi-chunk ids

thepastaclaw:
# PR Body

## Summary

- Validate multi-chunk `ChunkOp` ordering before processing any chunks.
- Reject malformed streams that end with a dangling `ChunkId`.
- Add regressions for both dangling-id forms.
- Assert restore state stays unchanged when rejecting a trailing id after a
  valid chunk.

Fixes dashpay/grovedb#705.

## Validation

- `cargo fmt --all`
- `git diff --check`
- `cargo test -p grovedb-merk test_process_multi_chunk --lib`
  - 5 passed, 0 failed; existing unused-import warnings only.
- Code review gate:
  - Repo: `dashpay/grovedb`
  - Base: `upstream/develop`
  - Head: `review-fix-multi-chunk-dangling-id`
  - Intent: reject malformed multi-chunk ordering before state mutation.
  - Recommendation: ship.

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for multi-chunk operations validation, including edge cases for invalid orderings.

* **Refactor**
  * Improved validation logic for multi-chunk operation ordering to better detect and reject invalid sequences.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/grovedb/pull/750?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #742, opened 2026-05-23
test: cover element type discriminant 20

thepastaclaw:
# PR Body

## Summary

- Add `Element::ProvableCountProvableSumTree` to the serialization
  discriminant drift test
- Update the base-discriminant comments and assertion count from `17` to `18`
- Cover discriminant `20` so future `Element` enum reorderings are caught

Fixes dashpay/grovedb#718.

## Validation

- `cargo test -p grovedb-element test_element_serialization_discriminants_match_element_type`
- Code review gate: Recommendation: ship

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected element type discriminant handling to properly encompass all expected variant types.

* **Documentation**
  * Updated element type documentation to accurately reflect the complete range of supported discriminants.

* **Tests**
  * Enhanced test coverage for element type serialization to validate all expected variants.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/grovedb/pull/742?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #741, opened 2026-05-22
chore: remove duplicate workspace members

thepastaclaw:
# Summary

- Remove duplicate `grovedb-bulk-append-tree` and `grovedb-commitment-tree`
  entries from the root workspace members list.
- Fixes dashpay/grovedb#719.

## Validation

- `cargo metadata --no-deps --format-version 1`
- `code-review dashpay/grovedb upstream/develop origin/fix-duplicate-workspace-members
  "Remove duplicate workspace member entries from root Cargo.toml for
  dashpay/grovedb#719"` → Recommendation: ship

## Summary by CodeRabbit

* **Chores**
  * Reorganized workspace crate configuration to streamline project structure.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/grovedb/pull/741?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #772, opened 2026-07-13
chore: bump workspace crate versions to 5.0.1

QuantumExplorer:
Bumps all 15 workspace crates from 5.0.0 to 5.0.1 (version fields and intra-workspace dependency specs), following the same pattern as the 5.0.0 bump (#764).

## Why a patch bump

Everything on develop since v5.0.0 is a fix, hardening, docs, CI, or a patch-level dependency bump — no new features or API additions:

- #728 — Reject overlong Merk keys before encoding
- #750 — fix(merk): reject dangling multi-chunk ids
- #755 — fix(storage): flush all column families in `RocksDbStorage::flush()`
- #767 — fix(commitment-tree): iterate checkpoints in ascending order in SqliteShardStore
- #768 — chore(commitment-tree): bump orchard to dashified-0.14.1
- #741 — chore: remove duplicate workspace members
- #742 — test: cover element type discriminant 20
- #769 / #770 — docs: audit-non-issues records
- #771 — ci: only deploy book from develop

`cargo check --workspace` passes. Cargo.lock is gitignored in this repo, so no lockfile change.

### DIFF
--- a/costs/Cargo.toml
+++ b/costs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grovedb-costs"
-version = "5.0.0"
+version = "5.0.1"
 edition = "2024"
 license = "MIT"
 description = "Costs extension crate for GroveDB"
--- a/grovedb-bulk-append-tree/Cargo.toml
+++ b/grovedb-bulk-append-tree/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grovedb-bulk-append-tree"
-version = "5.0.0"
+version = "5.0.1"
 authors = ["Samuel Westrich <sam@dash.org>"]
 edition = "2024"
 license = "MIT"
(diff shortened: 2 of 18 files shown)


QuantumExplorer:
This is Claude. The Linting failure was not caused by the version bump — CI's stable toolchain moved to Rust 1.97.0 (2026-07-09), which added three new clippy lints that now fail `-D warnings` on pre-existing code. Fixed in 6d0ad9cc:

- `question_mark` in `grovedb-dense-fixed-sized-merkle-tree/src/verify.rs` — manual match replaced with `?`
- `useless_borrows_in_formatting` at two `proof_debug` printlns in `grovedb/src/operations/proof/verify.rs` — **note: these are inside the locked V0 `verify_layer_proof`**, but the change only removes a redundant `&` in a debug println (identical output, no behavior or wire-format change)
- `for_kv_map` in `grovedb/src/replication/state_sync_session.rs` — iterate `.values()`

Verified locally with `cargo +1.97.0 clippy --workspace --all-features -- -D warnings` (clean).

PR #781, opened 2026-08-02
fix: verify-only access to indexed-axis proofs + bind subset lower layers

QuantumExplorer:
## Summary

Two follow-ups to the indexed-tree family (#657), found while integrating it into Dash Platform:

### 1. Expose indexed-axis proof verification to verify-only builds

The verify-side entry points of the indexed-axis proof envelope (`verify_indexed_axis_top_k` and friends) were `minimal`-gated at the module level, so a consumer compiling with `--no-default-features --features verify` (exactly what Dash Platform's `drive` crate does for its proof-verification layer) could not reach them. The prove-side items are now gated individually and the module is open to both feature sets, mirroring how the rest of the proof code splits prover from verifier. Verified with a reachability probe: without the change, `GroveDb::verify_indexed_count_top_k` and the other verify entry points fail to resolve under a verify-only build.

### 2. Bind, don't reject, a lower layer with no query below it

#657 added a hard rejection in `verify_layer_proof_v1`: a V1 proof supplying a lower layer for a tree the query doesn't descend into errored with *"the element bytes would be unbound"*. The underlying concern is real — a `KVValueHash`-family node hashes only `(key, value_hash)`, so reporting the element without binding it would let a prover attach a dummy layer and swap in forged bytes under a genuine root hash.

But the check conflates that attack with the **normal shape of a subset verification**: `verify_subset_query` exists precisely to run a narrow query against a proof generated for a wider one, and any such proof legitimately descends below the narrow query. The rejection broke previously-passing verifications downstream (Dash Platform extracts a commitment tree's `total_count` from the same proof bytes that prove a note fetch — a single-key, no-subquery subset query that stops at the tree element).

The fix binds instead of rejecting: the lower layer is consumed for its root hash and the existing `combine_hash(H(value), child_root)` chain check does the binding; only the reporting differs. Succinct mode still rejects outright with the original message. The result is strictly stronger than v5.0.1, which bound nothing on this path — pinned by two tamper tests (a dummy layer, and a sibling subtree's real-but-wrong layer, both rejected in subset mode).

## Tests

- 3 new tests (plain-Tree repro, two-tamper binding test, and the exact commitment-tree subset shape that broke downstream), all failing before the fix with the reported error verbatim
- `cargo test -p grovedb --lib`: 2535 passed, 0 failed
- `cargo test --workspace`, `cargo clippy --workspace --all-features -- -D warnings`, `cargo check -p grovedb --no-default-features --features verify`: all clean

## Known related gap (not addressed here)

Terminally-reported `CommitmentTree` / `MmrTree` / `BulkAppendTree` / `DenseAppendOnlyFixedSizeTree` elements (no lower layer at all) are bound by nothing in V1 proofs — pre-existing at v5.0.1, independent of this change, and being worked separately since the fix touches the proof format.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proof verification for parent-level queries and lower-layer root validation.
  * Ensured subset verification correctly reports requested elements while rejecting mismatched or unrequested proof layers.
  * Preserved authentication for empty targets through the verification chain.

* **Compatibility**
  * Made verification-only builds available without enabling proof generation.

* **Tests**
  * Added regression coverage for commitment-tree queries, nested proofs, strict verification, and invalid lower-layer proofs.

### DIFF
--- a/grovedb/src/operations/proof/indexed_axis/axis_api.rs
+++ b/grovedb/src/operations/proof/indexed_axis/axis_api.rs
@@ -3,13 +3,18 @@
 //!
 //! Each wrapper pins the [`IndexAxis`] and forwards; no logic lives here.
 
+#[cfg(feature = "minimal")]
 use grovedb_costs::CostResult;
 use grovedb_element::indexed::IndexAxis;
 use grovedb_merk::proofs::Query as MerkQuery;
+#[cfg(feature = "minimal")]
 use grovedb_path::SubtreePath;
+#[cfg(feature = "minimal")]
 use grovedb_version::version::GroveVersion;
 
-use crate::{Error, GroveDb, TransactionArg};
+#[cfg(feature = "minimal")]
+use crate::TransactionArg;
+use crate::{Error, GroveDb};
 
 use super::{IndexedAxisAggregateResult, IndexedAxisPaginatedResult, IndexedAxisQueryResult};
 
@@ -18,6 +23,7 @@ impl GroveDb {
 
     /// Prove the top-`k` entries of the count axis. Thin wrapper over
     /// [`Self::prove_indexed_axis_top_k`] with `axis = Count`.
+    #[cfg(feature = "minimal")]
     pub fn prove_indexed_count_top_k<'b, B, P>(
         &self,
         path: P,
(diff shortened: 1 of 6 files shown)


QuantumExplorer (commented):
Approved

PR #782, opened 2026-08-02
fix(grovedb): bind terminal non-Merk tree element bytes to the parent value_hash

QuantumExplorer:

A V1 proof that reported an `Element::CommitmentTree` / `MmrTree` / `BulkAppendTree` / `DenseAppendOnlyFixedSizeTree` as a **terminal** result — the query targets the tree element itself and the prover emits no lower layer — left the serialized element bytes unbound to the `value_hash` its parent Merk commits to.

Both existing binding mechanisms skipped these four types:

- the empty-tree `combine_hash(H(value), NULL_HASH)` check is gated on `!is_non_empty_tree()`, but [`is_non_empty_tree()`](https://github.com/dashpay/grovedb/blob/develop/grovedb-element/src/element/helpers.rs) returns `true` unconditionally for them;
- the `child_hash_verified` requirement was gated on `is_non_empty_merk_tree()`, which by construction excludes them.

Meanwhile the prover only rewrote regular non-empty Merk trees to `KVValueHashFeatureTypeWithChildHash`, leaving these emitting a bare `KVValueHash` — a node that hashes only `(key, value_hash)`, so the value bytes never enter the node hash.

**Impact.** A prover could serve forged element bytes (an inflated or deflated `CommitmentTree` `total_count`, a different MMR size) alongside the genuine `value_hash` and still reconstruct the correct root hash. Dash Platform's `GetShieldedNotesCount` verifier reads `total_count` from exactly such a terminal element, so a malicious node could misreport a wallet's shielded sync denominator.

This is **pre-existing, not a regression** — it reproduces identically at tag `v5.0.1`. It is also **not** what the recent subset-verification fix (e2168c1) addressed: that path has a lower layer available and does bind via `combine_hash(H(value), child_root)`. This gap is only on the no-lower-layer terminal path.

## What was done?

Worth stating up front, because it rules out the obvious shortcut: this could **not** be fixed verifier-side by asserting `hash == value_hash(value_bytes)`. These types are written through `insert_subtree` → `KV::new_with_layered_value_hash`, so the parent commits `combine_hash(H(value), state_root)`, not plain `H(value)` — and the state root is not derivable from the element bytes (a `CommitmentTree`'s state root depends on the actual notes, not on `total_count`). The child hash has to travel in the proof.

No new proof format is needed. `KVValueHashFeatureTypeWithChildHash` (tag `0x1c`) already verifies `combine_hash(H(value), child_hash) == value_hash`, which is exactly the composition these types commit — the prover simply was not using it here.

**Prover** (`operations/proof/generate.rs` + `operations/proof/bind_terminal_non_merk_tree/`): the terminal arm now covers all four types and rewrites the node to carry the tree's state root, computed by the new `non_merk_tree_child_hash`. `CommitmentTree` moved out of the empty-trees arm, since it is bound whether or not it holds notes. Each arm mirrors its write path exactly:

- `MmrTree` / `BulkAppendTree` / `DenseAppendOnlyFixedSizeTree` are inserted with `NULL_HASH` while still empty. Note an empty `BulkAppendTree`'s `compute_current_state_root()` is **not** `NULL_HASH`, so the zero-count case short-circuits.
- `CommitmentTree` needs no special case: the sinsemilla/bulk composition already yields `EMPTY_COMMITMENT_TREE_STATE_ROOT` at count 0.

A self-check fails loudly if a recomputed root does not reproduce the committed value hash, so any future convention drift surfaces as a clear prover error instead of a proof that cannot verify.

**Verifier** (`operations/proof/verify.rs`): the child-hash requirement widens from `is_non_empty_merk_tree()` to `is_non_empty_tree()`, which adds exactly these four types.

**V0 envelopes are deliberately untouched.** The gap is broader there (a regular non-empty `CountTree`'s count is forgeable the same way) and V0 documents the child-hash check as V1-only. It is a frozen wire format and Platform no longer accepts V0 proofs.

## How Has This Been Tested?

Five new tests in `grovedb/src/tests/proof_coverage_tests.rs`. Each forgery is tried in two shapes: tampering the value bytes inside the honest child-hash node, and downgrading the node back to bare `KVValueHash` — the latter is what actually exercises the verifier's new requirement, since it is the shape the gap allowed.

- `terminal_commitment_tree_count_forgery_is_detected`
- `terminal_mmr_tree_size_forgery_is_detected`
- `terminal_bulk_append_and_dense_tree_forgeries_are_detected`
- `empty_non_merk_trees_still_prove_and_verify` — empty instances of all four must still prove and verify
- `terminal_non_merk_tree_child_hash_version_gate` — pins **both** sides of the gate: `GROVE_V3` still emits a bare `KVValueHash` and still accepts the forged `total_count`; `GROVE_V4` emits the child-hash node and rejects it. The V3 assertion is deliberately an assertion about a hole — if it starts failing, the fix has leaked into a released version.

Before the fix, the CommitmentTree test failed with `verification accepted it: Ok([Ok(commitment_tree: count: 999 chunk_power: 10)])` against a real count of 3.

- `cargo test -p grovedb --lib` — 2540 passed, 0 failed
- `cargo test -p grovedb-version --lib` — 47 passed, including the `*_unchanged_fields_remain_zero` guards
- `cargo clippy --workspace --all-features -- -D warnings` — clean
- `cargo fmt --all` — applied, `--check` clean
- `cargo check -p grovedb --no-default-features --features verify` — the verifier change compiles in verify-only builds
- The subset-verification tests from e2168c1 (`test_commitment_tree_element_count_subset_query_against_note_fetch_proof`, `test_subset_mode_still_binds_element_bytes_to_lower_layer`) still pass

## Breaking Changes

None on released versions — the fix is gated behind **`GROVE_V4`**.

`GROVE_V3` is live, and per `v4.rs` a fix that changes an accepted/rejected outcome or a tracked cost cannot land on a released version without diverging nodes. This change does both: an upgraded verifier rejects proofs a released one accepts, and deriving the state root costs the prover extra storage reads and hash calls, which feeds fees. So it lands as a new slot, `proof.terminal_non_merk_tree_child_hash` — `0` in V1..V3, `1` in V4 — with both prover and verifier branching on it.

The prover side follows the versioning system's code structure — a module directory with one file per version and a dispatching `mod.rs`, modelled on `operations/insert/add_element_on_transaction/`:

```
operations/proof/bind_terminal_non_merk_tree/
  mod.rs   dispatch on the slot (+ UnknownVersionMismatch), docs on what differs
  v0.rs    released no-op; documents the unbound bytes as a known, preserved gap
  v1.rs    state-root derivation, self-check, node rewrite
```

The gated unit is the binding step rather than the whole enclosing function, so the per-version files stay small — duplicating the ~1000-line `prove_subqueries_v1` would not have been reasonable. `generate.rs` just calls the dispatcher, and the module is `minimal`-gated so verify-only builds are unaffected.

- **Prover, v0:** the node is left exactly as always emitted (bare `KVValueHash`), only the limit moves. Released byte and cost shape untouched.
- **Verifier, v0:** unchanged. Non-empty **Merk** trees keep requiring the child hash at every version — released behaviour since V3 — so only the four non-Merk types are added, and only from V4.

A V4 verifier therefore never rejects an honest V3 proof, and a V3 verifier never demands a node a V3 prover does not emit.

**Worth being explicit:** the forgery stays exploitable on V1..V3 and closes when protocol v4 activates. That is inherent to gating — fixing it in place is precisely the divergence `v4.rs` exists to prevent — and it matches how the other fixes parked on V4 are being handled.

## Note for reviewers

This branch was cut from local work that is not yet on `develop`, so the PR carries two pre-existing commits that are not part of this change:

- `82726a0` feat(grovedb): expose indexed-axis proof verification to verify-only builds
- `e2168c1` fix(grovedb): bind, don't reject, a lower layer with no query below it

**Only the third commit (`ff77fc0`) is this change.** If those two are landing through their own PR, this one should be rebased once they do.

## Checklist:

### DIFF
--- a/grovedb-version/src/version/grovedb_versions.rs
+++ b/grovedb-version/src/version/grovedb_versions.rs
@@ -139,6 +139,33 @@ pub struct GroveDBOperationsProofVersions {
     pub verify_subset_query_with_absence_proof: FeatureVersion,
     pub verify_query_with_chained_path_queries: FeatureVersion,
     pub verify_query_get_parent_tree_info_with_options: FeatureVersion,
+    /// Whether a V1 proof binds the element bytes of a **terminally-reported
+    /// non-Merk tree** — `CommitmentTree`, `MmrTree`, `BulkAppendTree`,
+    /// `DenseAppendOnlyFixedSizeTree` — to the `value_hash` its parent Merk
+    /// commits to. "Terminal" means the query targets the tree element itself
+    /// and the prover emits no lower layer.
+    ///
+    /// - `0` (V1..V3): the prover emits a bare `KVValueHash` node and the
+    ///   verifier does not require a child hash. That node hashes only
+    ///   `(key, value_hash)`, so the serialized element bytes are unbound: a
+    ///   prover can serve a forged entry count (an inflated or deflated
+    ///   `CommitmentTree` `total_count`, a different MMR size) alongside the
+    ///   genuine `value_hash` and still reconstruct the correct root hash.
+    /// - `1` (V4+): the prover emits
+    ///   `KVValueHashFeatureTypeWithChildHash` carrying the tree's own state
+    ///   root, and the verifier requires it, so the merk-level
+    ///   `combine_hash(H(value), child_hash) == value_hash` check closes the
+    ///   loop. This is exactly the composition the parent commits, since these
+    ///   types are written through `insert_subtree`.
+    ///
+    /// Gated rather than applied unconditionally on two counts. It flips an
+    /// accepted/rejected outcome — an upgraded verifier rejects proofs a
+    /// released one accepts — and computing the state root costs the prover
+    /// extra storage reads and hash calls on a released path. The
+    /// non-Merk tree types this covers are the only elements affected;
+    /// non-empty **Merk** trees have required the child hash since V3 and
+    /// stay bound at every version.
+    pub terminal_non_merk_tree_child_hash: FeatureVersion,
 }
 
 #[derive(Clone, Debug, Default)]
--- a/grovedb-version/src/version/v1.rs
+++ b/grovedb-version/src/version/v1.rs
@@ -166,6 +166,7 @@ pub const GROVE_V1: GroveVersion = GroveVersion {
                 verify_subset_query_with_absence_proof: 0,
                 verify_query_with_chained_path_queries: 0,
                 verify_query_get_parent_tree_info_with_options: 0,
+                terminal_non_merk_tree_child_hash: 0,
             },
             average_case: GroveDBOperationsAverageCaseVersions {
                 add_average_case_get_merk_at_path: 0,
--- a/grovedb-version/src/version/v2.rs
+++ b/grovedb-version/src/version/v2.rs
@@ -166,6 +166,7 @@ pub const GROVE_V2: GroveVersion = GroveVersion {
                 verify_subset_query_with_absence_proof: 0,
                 verify_query_with_chained_path_queries: 0,
                 verify_query_get_parent_tree_info_with_options: 0,
+                terminal_non_merk_tree_child_hash: 0,
             },
             average_case: GroveDBOperationsAverageCaseVersions {
                 add_average_case_get_merk_at_path: 0,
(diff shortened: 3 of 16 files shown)


QuantumExplorer (commented):
This should be fine, as proof generation is not in consensus.

Issue #784, opened 2026-08-02
PrivateDocumentStore: append-only element type for fixed-size opaque entries with committed config and provable range reads


QuantumExplorer:
## Motivation

Platform is adding **private document types**: a data contract can declare a document type `private` — no public index structure; each document is stored as a hiding `commitment (32B)` plus a `ciphertext` of uniform, contract-declared size; recipients discover documents by scanning and trial-decrypting rather than querying indices. Uniqueness is enforced by a sibling tag set (existing `ProvableCountTree`, nullifier-style), and the state transitions are funded from the shielded pool — all Platform-side concerns.

Phase one covers **non-modifiable** private documents only: entries are write-once, nothing is ever updated, deleted, or proven-against later. That means the store needs **no Sinsemilla frontier and no anchor** — it is a pure append-only log with provable reads. (Modifiable private documents are a later phase built on the anchored `DataCommitmentTree` of #783 plus a nullifier set; explicitly out of scope here.)

This issue adds the element type for that store. Behaviorally it is fully generic — an append-only store of fixed-size opaque entries; GroveDB never interprets a "document" — the name simply describes its Platform use.

## Design

- **`ElementType::PrivateDocumentStore = 15`** with counted semantics, plus **`NonCountedPrivateDocumentStore = 143`** following the existing +128 convention (12–14 and 140–142 are taken by Mmr/BulkAppend/DenseAppendOnlyFixedSize trees; 15/143 are the next free pair). Matching `TreeType` variant carrying the chunk power.
- **Thin wrapper over `BulkAppendTree`** — the same relationship `CommitmentTree` has to it, minus the Sinsemilla frontier.
- **Committed config**: the element value carries `{entry_size, chunk_power}`, and the config is bound into the state root, e.g. `blake3("pds_state" || config_hash || bulk_state_root)` — so the declared entry size is consensus-visible and a proof can never be reinterpreted under a different config. Precomputed empty-root constant with a runtime-equivalence test (mirroring `EMPTY_COMMITMENT_TREE_STATE_ROOT`).
- **Insert** validates exact entry size; there is no per-entry delete or update — immutability is enforced by the type, matching what the Platform grammar's default promises.
[...]

Issue #783, opened 2026-08-02
DataCommitmentTree: generic commitment tree for data (configurable hash domain, entry format, payload size)


QuantumExplorer:
## Motivation

Platform is designing **anonymous DashPay contact requests**, generalizing into a platform capability: **private document types** that any data contract can declare — no public index structure, entries stored as hiding commitments plus ciphertexts, funded from the shielded pool, with contract-declared constraints enforced by a Halo 2 circuit. This needs a new kind of tree in GroveDB, with two consumer shapes on the Platform side:

1. **Hecate — the key membership tree** (a single global instance): an append-only Sinsemilla Merkle tree over identity keys, leaf = `CRH(key_type || key_bytes)`. Recipient-existence constraints need only *membership* (no knowledge of the key's secret), so they work for every existing key type; sender-authorization constraints prove *knowledge* of the leaf key's secret, which is only affordable in-circuit for a Pallas-curve key type. The tree's anchor is a public input to the circuit.
2. **Modifiable private document stores** (later phase): updating hidden state means appending the new version and nullifying the old — which requires membership proofs against the store, i.e. an anchored tree, composed with a sibling nullifier set in the same pattern the shielded pool uses. *Non-modifiable* private document stores (phase one) never have their anchor consumed by anything and are therefore **not** built on this type — they get the simpler frontier-less `PrivateDocumentStore` element (#784). An anchor consumer is the dividing line: entries that someone later proves membership of need this type; write-once data does not.

The conceptual split from the existing type 11 is **value vs. data**. Type 11's entries are Orchard notes — money — with `rho`/`cv_net`/binding-signature semantics baked into the format because notes carry value that must balance. The new type's entries are pure existence assertions with an opaque fixed-size payload; nothing is ever spent from them. To GroveDB both are just `leaf + payload`, which is why one generic type covers all of it — the value semantics were never GroveDB's business, and type 11 leaking them (rho, cv_net, memo-derived payload sizes) is the layering smell this issue also fixes.

**Naming**: the new GroveDB type is `DataCommitmentTree` (pairs with the existing note commitment tree as data-vs-value). Platform-side instance codenames (Hecate, and the shielded pool tree "medusa") stay out of GroveDB — this library stays application-agnostic.

## Current state (updated 2026-08-25)

The type-11 coupling that motivates this issue is unchanged:

[...]

PR #786, opened 2026-08-03
feat: paginated position-range reads with proofs at the BulkAppendTree layer

QuantumExplorer:
Implements work-plan item 4 of #784: **paginated position-range reads with proofs** at the `BulkAppendTree` layer — the scanning hot path where clients of append-only stores walk "all entries since my cursor" in pages.

### Crate layer (`grovedb-bulk-append-tree`) — the shared seam
- `BulkAppendTree::get_range(start, limit)` → `RangePage { entries, total_count }`. Chunk-aligned: each completed chunk overlapping the page is read and deserialized exactly once, so a page costs O(chunks touched) blob reads plus one read per buffer entry — not O(entries) random reads.
- `position_range_query(start, limit)`: the canonical query shape (8-byte big-endian position keys), derived identically by prover and verifier from `(start, limit)`.
- `BulkAppendTreeProof::generate_for_range` / `verify_range`: paginated proof round-trip over the existing chunk-MMR + dense-buffer proof, completeness enforced.
- `CommitmentTree::get_range` pass-through for the shielded-pool scanning path.

### GroveDB layer
- Ops: `GroveDb::bulk_get_range` and `GroveDb::commitment_tree_get_range`.
- `PathQuery::new_bulk_position_range(path, key, start, limit)`: canonical page query (element key + position-range subquery).
- `GroveDb::prove_bulk_position_range`: proves one page via the existing V1 proof layering (`ProofBytes::BulkAppendTree` / `ProofBytes::CommitmentTree`).
- `GroveDb::verify_bulk_position_range_proof`: verifies the page entries (ascending, contiguous, complete) **and** extracts the authenticated `total_count` from the same proof bytes by subset-verifying the element itself. Absence beyond the end falls out of the provable count — `position >= total_count` does not exist — so ranges past the end need no per-position absence proofs; a page shorter than `limit` means the scan caught up with the tip.

The machinery lives at the shared `BulkAppendTree` layer and dispatches on element type, so the upcoming `PrivateDocumentStore` (#784) and `DataCommitmentTree` (#783) inherit it by adding their element variants to the prove/verify match arms.

### Fix uncovered along the way
A V1 proof descending into an **empty** `BulkAppendTree` could never verify: insert commits the element with a `NULL_HASH` child hash (mirrored by `verify_grovedb`'s integrity walk), but `verify_bulk_append_lower_layer` returned the domain-tagged `blake3("bulk_state" || 0*32 || 0*32)`, so the `combine_hash` chain check always failed. The verifier now returns `NULL_HASH` for a zero-count `BulkAppendTree` element (sound: `verify_and_compute_root` already rejects any proof carrying data for a zero-count tree). `CommitmentTree` is untouched — `EMPTY_COMMITMENT_TREE_STATE_ROOT` deliberately folds in the *computed* empty bulk root.

## Versioning / compatibility
- **No serialized proof-format change** — the proofs reuse the existing V1 layering, so no new `GROVE_V*` flag is introduced.
- **No write-path changes**: no state root moves by a byte. The full pre-existing CommitmentTree suites pass unchanged (the canary).
- Under GROVE_V1/V2 the V0 prover already rejects descending into these (unreleased) tree types; nothing widens there.

## Tests & benches
- Crate-level: `get_range` and proof round-trips across chunk boundaries, empty ranges (limit 0), ranges past the end, single-entry pages, large multi-chunk pages, empty trees, saturating-overflow starts, tampered-root and missing-chunk rejection, and a cursor-walk covering the whole tree.
- GroveDB-level: the same matrix for both `BulkAppendTree` and `CommitmentTree` through `prove_bulk_position_range` / `verify_bulk_position_range_proof`, plus nested paths, wrong-element errors, and narrower-proof rejection — every round-trip asserts the proof root equals the live DB root.
- New bench `bulk_range_scan_benchmark`: paged read / per-page prove / per-page verify at page sizes 16/256/1024, and a full proved cursor-walk (4096 entries, chunk size 64, 96-byte entries).

## Validation
- `cargo nextest run --workspace --all-features`: **4536 passed, 0 failed, 4 skipped**
- `cargo clippy --workspace --all-features -- -D warnings`: clean
- `cargo check --workspace --all-features --all-targets`: clean (bench compiles)
- `cargo fmt --all --check`: clean
- Bench smoke-run in `--test` mode: all sections pass

## Summary by CodeRabbit

- **New Features**
  - Added paginated range reads for BulkAppendTree and CommitmentTree data.
  - Added APIs to retrieve page entries alongside the tree’s total count.
  - Added generation and verification of authenticated position-range proofs.
  - Added query support for bounded position ranges.

- **Bug Fixes**
  - Improved handling of empty, out-of-range, incomplete, and invalid range proofs.

- **Tests**
  - Added extensive coverage for pagination, chunk boundaries, buffering, empty trees, limits, and proof validation.


QuantumExplorer:
This is Claude. Addressed all four review comments in ad8c8a75:

1. **MMR hoisted out of the chunk loop** (`fetch.rs`): `get_range` now builds the `MmrStore` + `MMR` (and its overlay clone) once per call and reads each chunk leaf from that single instance, matching the reuse pattern in `BulkAppendTreeProof::generate`. Missing-chunk corruption errors and deserialization/filtering behavior are preserved.

2. **Shared-chunk disjoint-range case pinned** (`bulk_append_tree_tests.rs`): extended `test_bulk_position_range_proof_wrong_range_rejected` — a proof generated for `[0, 2)` *does* verify a request for `[2, 4)` inside the same chunk, and the test asserts the returned entries and root. This is intended: chunk-aligned proofs carry the whole authenticated blob, so every entry in it is bound to the root; the wider-range rejection (needing an unproved chunk) is unchanged.

3. **CommitmentTree wrong-range rejection test added** (`commitment_tree_tests.rs`): `test_commitment_tree_position_range_proof_wrong_range_rejected`, exercising the `ProofBytes::CommitmentTree` envelope (sinsemilla prefix + different child-hash derivation) as suggested.

4. **Version gates added**: `prove_bulk_position_range` and `verify_bulk_position_range_proof` now start with the standard `check_grovedb_v0*` gate and are represented in the version table (`0` across GROVE_V1..V4 — no behavior change; the delegated `prove_query`/`verify_query_with_options` gates still apply transitively, and under V1/V2 the V0 prover rejects bulk descent as before).

Validation after the changes: `cargo nextest run --workspace --all-features` — 4557 passed, 0 failed; clippy (`--workspace --all-features -- -D warnings`), `--all-targets` check, verify-only build, and `fmt --check` all clean.

QuantumExplorer:
This is Claude. Follow-up on the coverage gate: codecov/patch flagged the new code at 80.55% against the repo's 90% bar, now green at ~90.1% after two commits:

- **9326581f** — tests for every error arm reachable without contrived proofs: `get_range` corruption paths (storage claiming chunks/buffer entries it doesn't hold), version-gate rejections for both new entry points under an unknown feature version, and adversarial `verify_bulk_position_range_proof` inputs (a canonical range proof over a plain `Tree` with 8-byte item keys must be rejected when extracting `total_count`; a proof for a nonexistent key must not invent one).
- **eac3480b** — the MMR storage-error path: `MemStorageContext` (codecov-ignored test util) grew a `fail_gets` switch so a broken backing store is exercised end-to-end and surfaces as `MmrError`, not a panic or a silent empty page.

The remaining uncovered lines are deliberately left as-is: they are defense-in-depth `InvalidProof`/`CorruptedData` arms that cannot be reached by any proof that survives the lower layer's own completeness checks (root-mismatch between sub-proofs of the same bytes, non-8-byte / non-item / non-contiguous rows under a genuine bulk element, and the buffer `ok_or_else` arm — the dense tree's `get()` errors rather than returning `Ok(None)` for a missing value below its count). Reaching them would mean weakening the checks they back up.

All 12 checks are green: full suite (4559 tests run, 4 skipped), clippy `-D warnings`, all-targets check, verify-only build, formatting, and both codecov statuses.

grovedb-bulk-append-tree/src/tree/fetch.rs
```diff
@@ -63,6 +63,90 @@ impl<'db, S: StorageContext<'db>> BulkAppendTree<S> {
...
+    /// This is the paginated-scan read path: clients walking "all entries
+    /// since my cursor" call it with their cursor as `start` and advance by
+    /// `entries.len()`. The read is chunk-aligned — each completed chunk
+    /// overlapping the range is read and deserialized exactly once, so a
+    /// page costs O(chunks touched) blob reads plus one read per buffer
+    /// entry, not O(entries) random reads.
+    ///
+    /// Absence needs no lookup: positions `>= total_count` do not exist, so
+    /// a page shorter than `limit` means the end of the tree was reached.
+    pub fn get_range(&self, start: u64, limit: u16) -> Result<RangePage, BulkAppendError> {
```
QuantumExplorer: **[P1] Preserve range-read storage costs**

`get_range` returns a plain `Result`, so the MMR `element_at_position(...).unwrap()` and every buffer `get` discard their `CostResult.cost`. Consequently, `bulk_get_range` and `commitment_tree_get_range` report none of the page's seeks or loaded bytes—up to 65,535 uncharged buffer reads. Please return and aggregate a `CostResult` through both GroveDB wrappers so cost limits reflect the actual work.

grovedb-bulk-append-tree/src/tree/fetch.rs
```diff
@@ -63,6 +63,90 @@ impl<'db, S: StorageContext<'db>> BulkAppendTree<S> {
...
+                            chunk_idx, e
+                        ))
+                    })?;
+                let blob = node.and_then(|n| n.into_value()).ok_or_else(|| {
+                    BulkAppendError::CorruptedData(format!(
+                        "missing chunk blob for index {}",
+                        chunk_idx
+                    ))
+                })?;
+                let chunk_entries = deserialize_chunk_blob(&blob)?;
```
QuantumExplorer: **[P2] Validate completed chunk length**

`deserialize_chunk_blob` accepts any valid encoded entry count, but this loop assumes each completed chunk contains exactly `epoch_size` entries. A short chunk silently omits positions; an oversized chunk can overlap positions from the next chunk. The returned page then violates its contiguous-page contract and cursor scans can stall before `total_count`. Please reject the chunk as corrupted unless `chunk_entries.len() == epoch_size as usize`.

QuantumExplorer (commented):
I found two actionable issues in the new range-read path. The focused range/proof tests pass, but neither issue is covered by the current tests.

grovedb-bulk-append-tree/src/tree/fetch.rs
QuantumExplorer: This is Claude. Fixed in acb99bdb: `get_range` now returns a `CostResult` — the chunk-MMR node reads and every dense-buffer read charge their seeks and loaded bytes (buffer reads go through `dense_tree.get` directly so nothing is discarded), and the costs aggregate through `CommitmentTree::get_range`, `bulk_get_range`, and `commitment_tree_get_range`. Added `test_bulk_get_range_reports_storage_costs`, which asserts nonzero seek/loaded-byte costs on a chunk+buffer-spanning page and that an 11-entry page loads more bytes than a 1-entry page.

grovedb-bulk-append-tree/src/tree/fetch.rs
QuantumExplorer: This is Claude. Fixed in acb99bdb: `get_range` now rejects a completed chunk as `CorruptedData` unless it deserializes to exactly `epoch_size` entries, with a test that tampers the MMR overlay with both a short and an oversized blob. One note on scope: the check is deliberately only on this raw read path — in proof verification the chunk bytes are bound to the state root (a wrong-length blob changes the root and fails the comparison), and the audit NOTE in `proof/mod.rs` documents that adding a length check there is redundant by design, so that side is unchanged.

QuantumExplorer:
This is Claude. Fixed up the PR against current `develop` (32 commits had landed since the branch point, incl. the unified-PathQuery stack, #822/#826 append accounting, and the indexed-axis read work), pushed as two commits on top of the existing head:

**`9584bdb6` — Merge `develop` into `feat/bulk-append-range-reads`**
- The only textual conflict was `grovedb-bulk-append-tree/src/test_utils.rs`: develop's `MemStorageContext` now has its own fault injection (`fail_get` / `fail_reads()`), which supersedes the `fail_gets` switch this PR added for the MMR storage-error test. Resolved to develop's file verbatim; the test uses `fail_get`.
- Semantic drift, all mechanical: `BulkAppendTree::append` now takes `&GroveVersion` (#822); the per-module path builders (`build_ct_path` / `build_subtree_path_for_bulk`) were replaced by `crate::util::subtree_path_with_key`; `Query` gained `read_mode` — the canonical position-range query sets it to `None`.

**`7c038fc4` — refactor + docs**
- `get_range`'s buffer tail hand-rolled a cost-charging dense-tree read because `get_buffer_value` discarded cost at the time. develop (#826) has since added `get_buffer_value_with_cost` for exactly that, so the loop now goes through it — same accounting, same `CorruptedData` on a lost value, less duplication. The chunk loop keeps its hoisted single MMR (per-chunk `get_chunk_value_with_cost` would rebuild it, which is what the earlier review asked to avoid).
- Book: `bulk_get_range` / `commitment_tree_get_range` added to the op tables, plus a short section on the paginated position-range proof entry points.

No change to the proof format, the state roots, or the version table (the two new ops remain `0` across GROVE_V1..V4). Effective diff vs develop is the PR's original 23 files + the two book pages; nothing else rode in.

Validation on the merged tree:
- `cargo nextest run --workspace --all-features`: **4992 passed, 0 failed, 4 skipped**
- `cargo clippy --workspace --all-features -- -D warnings`: clean
- `cargo check --workspace --all-features --all-targets`: clean
- `cargo fmt --all --check`: clean

Issue #778, opened 2026-08-02
State sync does not support indexed trees (PCIT/PSIT/PCPSIT)


QuantumExplorer:
This is Claude, filing on behalf of @QuantumExplorer.

## Summary

State sync fails closed on indexed trees. Both sides reject them:

- **Source**: `fetch_chunk` refuses to emit a chunk for an indexed subtree's prefix.
- **Discovery**: `state_sync_session.rs` rejects an indexed element with `NotSupported` ("state sync does not yet support indexed trees").

Pinned by `fetch_chunk_source_side_rejects_indexed_tree_chunk` in `grovedb/src/tests/replication_session_tests.rs`.

## Why it matters

**A node cannot state-sync a database containing indexed trees.** Failing closed is the right default — silently syncing a primary without its secondaries would produce a node whose H1-A bindings are wrong — but it means indexed trees are unusable in any deployment that relies on state sync for node bootstrap or recovery, unless this lands first.

## What makes it non-trivial

The secondaries live at derived prefixes (`Blake3(primary_prefix ‖ axis_tag)`) that are not GroveDB paths, so they are invisible to path-driven chunk discovery. A PCPSIT can carry up to three of them. The chunk protocol needs a way to enumerate, transfer, and verify them alongside the primary, and the restored element's `combine_hash_three` / `axes_digest` binding must be checked against the transferred secondary roots rather than trusted.

Worth checking against the known chunk-restore aggregate bug (#671) while designing this — the Provable\* family already has a restore-side OWN-vs-AGGREGATE mismatch.

## Why it can't be deferred indefinitely

The companion repair API (`reconcile_indexed_tree_secondaries`) covers *content* damage, but a Merk root hash also commits to tree **shape**, which is a function of operation history and cannot be recovered by any content repair. Damage that permanently rotated surviving nodes leaves a self-consistent, `verify_grovedb`-clean index whose root may still differ from an undamaged peer's. State sync is the only path back to consensus for that case.

## Scope

Blocks GROVE_V4 activation if Platform intends to use indexed trees at launch, not the merge of #657 (indexed trees are live on no released version).

QuantumExplorer:
**This is Claude**, closing on behalf of the maintainer.

[...]

PR #788, opened 2026-08-03
feat(replication): state sync support for the append-only tree family (fixes #785)

### LINKED ISSUE #785: State sync cannot transfer the append-only tree family — a single populated CommitmentTree makes snapshots from that node unusable
## Summary

The replication module (`grovedb/src/replication/`) has **zero support for the append-only (non-Merk) tree family** — CommitmentTree, MmrTree, BulkAppendTree, DenseAppendOnlyFixedSizeTree. The verified behavior today:

> **A single populated CommitmentTree anywhere in the grove makes state sync from that node impossible.** The source's `fetch_chunk` fails with an opaque `CorruptedData("... cannot create chunk producer for empty Merk")` when the syncing peer requests the CT subtree's chunk, so no peer can ever complete a snapshot sync from any node that holds one.

Since CommitmentTree (element discriminant 11) is the Dash Platform shielded-pool notes tree and is live on mainnet/testnet, this is a **live operational gap independent of any new feature work**: from the moment the first shielded note lands, every node holding it can no longer serve a usable state-sync snapshot, and new nodes can only bootstrap by replaying blocks. The failure mode is "fail hard, fail safe" — no corruption is produced — but availability of state sync is gone network-wide.

Reproduction tests (3 tests, all passing against develop) are on branch [`claude/eager-blackwell-838b49`](https://github.com/dashpay/grovedb/commit/ede9dc81) in `grovedb/src/tests/replication_session_tests.rs`:
- `state_sync_populated_commitment_tree_fails_on_source_fetch`
- `state_sync_populated_non_merk_trees_all_fail_on_source_fetch` (MmrTree / BulkAppendTree / DenseAppendOnlyFixedSizeTree — same failure)
- `state_sync_empty_commitment_tree_succeeds` (empty CT syncs fine; see "why the naive fix is dangerous")

## Trace

1. **Discovery enqueues the CT like a normal Merk subtree.** `discover_new_subtrees_metadata` (`grovedb/src/replication/state_sync_session.rs`) iterates parent elements and uses `value.is_any_tree()`, which includes all four non-Merk types. Only indexed trees get the up-front `NotSupported` rejection (#778); non-Merk trees fall through and are scheduled for Merk-chunk restore.
2. **Source-side `fetch_chunk` breaks on the payload.** `fetch_chunk` (`grovedb/src/replication.rs`) opens the CT's prefix and calls `merk.is_empty_tree()`, which raw-iterates the namespace (`merk/src/merk/mod.rs:534`). A populated CT's namespace contains its non-Merk payload entries (frontier, buffer, MMR nodes), so the check says "not empty" — but the CT's Merk is rootless by design (`root_key = None`), so `ChunkProducer::new` then fails:
   ```
   CorruptedData("failed to create chunk producer by prefix tx:<prefix> with:
   chunking error chunk from empty tree: cannot create chunk producer for empty Merk")
   ```
3. The syncing peer gets this error for the CT chunk request and the session can never reach `is_sync_completed()`.

## Why the naive fix is dangerous

An **empty** CommitmentTree state-syncs successfully today: no payload entries exist yet, `is_empty_tree()` is true, the source returns an empty chunk, the destination verifies the NULL Merk root and moves on — and the final app-hash check passes, because the parent Merk (which embeds the CT element bytes and the payload binding) restores byte-for-byte.

That passing path demonstrates the trap: **nothing in the restore pipeline ever recomputes a non-Merk state root from payload.** If `fetch_chunk` were "fixed" to skip these subtrees or return empty chunks, a sync of a populated CT would complete, pass the app-hash check at commit, and silently produce a destination whose CT element claims `total_count = N` while its frontier and note data are missing — detected only later by `verify_grovedb` or the first `commitment_tree_anchor` / `commitment_tree_get_value` call. Any fix must actually transfer and re-verify the payload.

## Affected element types

| Discriminant | Element | Status |
|---|---|---|
| 11 (+139 wrapped) | `CommitmentTree` | **Live** — shielded-pool notes tree |
| 12 (+140) | `MmrTree` | Same failure, verified |
| 13 (+141) | `BulkAppendTree` | Same failure, verified |
| 14 (+142) | `DenseAppendOnlyFixedSizeTree` | Same failure, verified |
| 15 (+143) | `PrivateDocumentStore` (#784) | Will inherit the gap — same storage model |
| 16 (+144) | `DataCommitmentTree` (#783) | Will inherit the gap — same storage model |

## What a correct transfer needs (per subtree)

Everything for these tree types lives as **raw KV entries in the subtree's single blake3(path)-prefixed data namespace**; the tree's own Merk is always empty and there are never child subtrees beneath it:

- **CommitmentTree** = BulkAppendTree + Sinsemilla frontier:
  - serialized frontier at `COMMITMENT_TREE_DATA_KEY = b"__ct_data__"` (`grovedb-commitment-tree/src/commitment_tree/mod.rs:28`)
  - dense-buffer entries at 2-byte BE position keys (`grovedb-dense-fixed-sized-merkle-tree`, `position_key(u16)`)
  - chunk-level MMR nodes at 4-byte tagged BE position keys (`MmrKeySize::U32`, MSB set for namespace separation); compacted chunk blobs are the MMR leaves
- **BulkAppendTree**: the same, minus the frontier
- **MmrTree**: MMR nodes at 8-byte BE position keys (`MmrKeySize::U64`)
- **DenseAppendOnlyFixedSizeTree**: entries at 2-byte BE position keys

So the transfer unit is simply "all KV pairs under the prefix", paged.

## Receiving-side verification — no new trust assumptions needed

The parent leaf's `value_hash` is already a commitment to the payload: inserts write `combine_hash(value_hash(element_bytes), state_root)` via `insert_subtree(..., combined_root, ...)` (`grovedb/src/operations/commitment_tree.rs`), where for CT

```
ct_state_root   = blake3("ct_state"   || sinsemilla_root || bulk_state_root)
bulk_state_root = blake3("bulk_state" || mmr_root        || dense_tree_root)
```

(#782 recently strengthened exactly this terminal binding.) The parent Merk restore already hash-verifies that leaf up the chain to the app hash, so the receiver holds a **trusted 32-byte commitment to the payload before fetching a single payload byte**. Verification is: write the raw KVs, recompute the type-specific state root, and require

```
combine_hash(value_hash(element_bytes), recomputed_state_root) == elem_value_hash_from_parent
```

The recompute dispatch **already exists** — `compute_non_merk_child_hash` (`grovedb/src/lib.rs:2379`), used by `verify_grovedb` for precisely this check. The future `pds_state` (#784) and `dct_state` (#783) config-committed roots re-verify identically, with the committed config participating in the element bytes / domain tag.

One caveat vs. Merk restore: raw KV paging is not incrementally verifiable per chunk the way Merk chunk proofs are — verification lands at subtree completion (a malicious peer is caught before the subtree is finalized, same transaction-rollback story as other restore failures). If incremental verification is wanted later, the BulkAppendTree's immutable chunk blobs are individually MMR-authenticated and could be verified per-page against the MMR peaks.

## Proposed approach

**Phase 0 — immediate guard (~20 lines + tests).** Reject non-Merk tree types up-front with a descriptive `Error::NotSupported` in both `discover_new_subtrees_metadata` and `fetch_chunk`, exactly like the indexed-tree guards from #778. Sync fails either way today; this replaces an opaque `CorruptedData` with an actionable error and protects against any future refactor that would turn the hard failure into the silent-skip trap above.

**Phase 1 — real support (one PR, est. 500–800 lines incl. tests).**
1. `fetch_chunk`: branch on `tree_type.uses_non_merk_data_storage()` → stream raw KV pages from the prefixed storage context (`raw_iter`) in bounded-size chunks; the local chunk id encodes the resume key. The global chunk id already carries the `TreeType` discriminant (7–10 cover this family), so the ID format extends naturally.
2. `state_sync_session`: for these subtrees, use a raw-restore path instead of a Merk `Restorer` — write KVs through the immediate storage context; on subtree completion recompute the state root via the `compute_non_merk_child_hash` dispatch and enforce the parent binding; reject the subtree on mismatch; skip child discovery (there are no children).
3. Bump / gate on the state-sync protocol version as appropriate. Replication is node-local wire behavior, not a consensus state transition — no committed hash changes — so this should **not** need `GROVE_V4` gating (consistent with #778's guards landing ungated); a `CURRENT_STATE_SYNC_VERSION` bump is the right lever if wire compatibility matters.
4. Tests: round-trip populated CT/MMR/bulk/dense including a multi-epoch CT (buffer + compacted chunks + frontier); tamper tests (bit-flip one payload byte, drop the frontier key → subtree must be rejected); interaction with `subtrees_batch_size` batching.

**Type 11 first?** Possible — CT is the only live exposure, and each type is an independent dispatch arm. But types 12–14 reuse the identical raw-page + recompute path, so the family costs barely more than CT alone; recommend shipping all four together, with #784/#783 (types 15/16) required to plug into the same path when they land.

## Cross-references

- #778 — state sync does not support indexed trees (PCIT/PSIT/PCPSIT): the sibling gap at the same discovery choke point; its guard pattern is the model for Phase 0 here. Whoever picks up either issue should probably land the shared plumbing once.
- #776 — bare-Reference overwrite of an indexed tree bypasses secondary-index cleanup: related indexed-tree hardening, listed for context.
- #783 — DataCommitmentTree: planned type 16/144, same BulkAppendTree+frontier storage model; its state-sync story is Phase 1 of this issue (`dct_state` re-verifies like `ct_state`).
- #784 — PrivateDocumentStore: planned type 15/143, ditto (`pds_state`).
- #775 / #679 — state-sync partial-commit atomicity: the raw-restore path must follow whatever resolution those land on.

QuantumExplorer:
Implements both phases proposed in #785: state sync previously could not transfer the non-Merk append-only tree family at all — a single populated `CommitmentTree` (the live shielded-pool notes tree), `MmrTree`, `BulkAppendTree`, or `DenseAppendOnlyFixedSizeTree` made every snapshot from the holding node unusable, with the source's `fetch_chunk` dying on an opaque `CorruptedData("... cannot create chunk producer for empty Merk")`.

## Commits

1. **Reproduction tests** — two-instance sync tests proving the failure mode for all four types (and that empty ones synced fine, demonstrating why a naive skip fix would have silently dropped payload).
2. **Phase 0 guard** — descriptive `NotSupported` on both the target-side discovery and source-side `fetch_chunk` paths, mirroring the indexed-tree guards from #778.
3. **Phase 1: entry-replay transfer** — real state-sync support for all four types, superseding the Phase 0 guards.

## Phase 1 design: target-driven entry replay

- **Target drives.** The target already holds the subtree's element (entry counts + parameters) from the hash-verified parent Merk, so it encodes a `(start, state, param)` page cursor (17 bytes) into every local chunk id it requests. The source never reconstructs tree geometry from its raw namespace, and the global chunk id format is unchanged.
- **Only leaf entries cross the wire** (plus the serialized Sinsemilla frontier on a commitment tree's first page — the frontier is an accumulator and cannot be replayed from entries without redoing every Sinsemilla hash). The source serves pages through the same accessors normal reads use (`get_chunk_value`/`get_buffer_value`, MMR `element_at_position`, dense `get`), with a 1 MiB / 8192-entry page budget.
- **The target replays** each entry through the real append primitives (`BulkAppendTree::append`, `MMR::push`, `DenseFixedSizedMerkleTree::insert`), so every internal node, chunk blob, and cached hash on the target is locally derived from the wire entries — there is no raw copy of internal state that verification could miss.
- **Verification**: at subtree completion the target recomputes the type-specific state root from its own storage via the new strict `GroveDb::compute_non_merk_state_root` (same dispatch `verify_grovedb` uses, but error-propagating instead of falling back), and requires `combine_hash(value_hash(element_bytes), state_root)` to equal the element value hash bound into the restored parent Merk (the #782 terminal binding). Any tampering with wire bytes — a flipped entry byte, a stripped or altered frontier, a dropped entry — fails the sync before commit.

## Protocol compatibility

`CURRENT_STATE_SYNC_VERSION` stays 1. Mixed old/new peers fail safe with no silent corruption in either direction: an old target requesting an append-only subtree cursor-lessly gets a descriptive `NotSupported` from a new source; a new target syncing from an old source fails when the old source cannot serve pages. Replication is node-local wire behavior — no committed hashes change — so no `GroveVersion` gating (consistent with the #778 guards).

## Tests

- Round trips for all four types: multi-epoch commitment tree (compacted chunk + buffer + frontier), multi-chunk bulk tree, MMR, dense, and a >1 MiB multi-page MMR transfer
- Post-sync usability: appending the same note on source and destination after sync produces identical root hashes
- Byzantine-source tamper rejection: flipped entry byte, stripped frontier, tampered frontier, dropped entry — each rejected with a specific error
- Subtree-batch-boundary interleaving (`subtrees_batch_size = 1` across CT/MMR/dense)
- Old-peer cursor-less request rejection; wire codec unit tests
- Full `grovedb` suite: 2553 passed, clippy clean

Closes #785. Cross-refs: #778 (indexed trees — same choke point, still guarded), #783 / #784 (future types 16/15 should plug into this same path).

## Summary by CodeRabbit

* **New Features**
  * Added replication support for append-only tree types, including commitment, MMR, bulk-append, and dense trees.
  * Added multi-page synchronization with cursor-based requests.
  * Added state validation to ensure restored data matches expected tree roots.

* **Bug Fixes**
  * Replication now reports errors for missing cursors, malformed pages, unreadable data, and integrity mismatches instead of silently proceeding.

* **Tests**
  * Expanded coverage for round-trip synchronization, empty and multi-page trees, tampered data, and invalid replication requests.


QuantumExplorer:
This is Claude. Addressed the CodeRabbit review in df05f983:

- **Applied** (5): tightened the frontier-tamper assertion to the specific rejection errors; documented the append-only page-serving contract in `fetch_chunk`'s notes; extended the `apply_chunk` SAFETY comment with the `set_new_transaction` invariant (good catch — that ordering constraint was previously unstated); sized the multi-page MMR test payload from `MAX_PAGE_BYTES` so the multi-page path can't silently lose coverage; consolidated the four sync-driver loops in the tests into one `run_sync(source, version, batch_size, mutator)` driver.
- **Skipped** (1): deriving `compute_non_merk_child_hash` from the strict `compute_non_merk_state_root`. The empty-tree branches are intentionally different: the lenient variant returns the *actual* inner Merk root (possibly non-null on a corrupt DB), so `verify_grovedb` still flags a count-0 append-only tree that has stray Merk nodes — `combine(value_hash, actual_root) != combine(value_hash, NULL)` bound at insert. Routing it through the strict variant's `NULL_HASH` would make that exact corruption verify clean. The caveat noted in the suggestion ("confirm those are identical for an empty inner Merk") is precisely the case where they must be allowed to differ.

grovedb/src/replication/non_merk_sync.rs
```diff
@@ -0,0 +1,675 @@
...
+    let more = match flag {
+        0 => false,
+        1 => true,
+        other => {
+            return Err(Error::CorruptedData(format!(
+                "non-merk page has invalid more-flag {other}"
+            )));
+        }
+    };
+    let mut sections = unpack_nested_bytes(packed)?;
```
QuantumExplorer, line 240: **[P1] Enforce the page-entry cap on receipt**

`MAX_PAGE_ENTRIES` is enforced only by the honest sender loops. This decoder trusts the packed section count and allocates before applying any protocol limit; `apply_page` then replays every decoded entry. A Byzantine snapshot source can send a few megabytes containing roughly a million zero-length entries, causing large metadata allocations, hashes, and transactional writes before the final root mismatch rejects the subtree. Reject counts above `MAX_PAGE_ENTRIES + 1` before `unpack_nested_bytes`, and enforce a receiver-side byte/work budget.

grovedb/src/replication/non_merk_sync.rs
```diff
@@ -0,0 +1,675 @@
...
+
+                let mut entries = Vec::new();
+                let mut bytes = 0usize;
+                let mut leaf = id.start;
+                while leaf < leaf_count
+                    && entries.len() < MAX_PAGE_ENTRIES
+                    && bytes < MAX_PAGE_BYTES
+                {
+                    let node = store_ref
+                        .element_at_position(leaf_to_pos(leaf))
```
QuantumExplorer, line 409: **[P2] Validate peer-controlled MMR cursor arithmetic**

Both `id.state` and `id.start` come from the requesting peer. With `state = u64::MAX`, `mmr_size_to_leaf_count` returns `2^63`; `start = 2^63 - 1` enters this loop and `leaf_to_pos` panics in debug builds with integer overflow. Release builds wrap to an unrelated position. Validate that the MMR size is canonical and that the leaf index is within the helper's arithmetic range—or derive the metadata from the source's authenticated element—before converting it.

QuantumExplorer (commented):
Found two resource-safety issues in the new non-Merk state-sync path. The functional round-trip and tamper tests pass, but both peer-facing boundaries need additional validation before this is safe against malformed requests/responses.

grovedb/src/replication/non_merk_sync.rs
QuantumExplorer: This is Claude. Fixed in ab4e07a0. `decode_non_merk_page` now peeks the packed section count and rejects anything above `MAX_PAGE_ENTRIES + 1` *before* `unpack_nested_bytes` allocates, and additionally enforces the sender loop's byte invariant on receipt: every entry but the last must fit cumulatively under `MAX_PAGE_BYTES` (only the final entry may overhang, which the transport cap bounds). Unit tests: `non_merk_page_enforces_entry_cap_on_receipt` (incl. a lying `u32::MAX` count with a truncated body) and `non_merk_page_enforces_byte_budget_on_receipt`.

grovedb/src/replication/non_merk_sync.rs
QuantumExplorer: This is Claude. Fixed in ab4e07a0. Added `validate_mmr_size`, which requires the cursor `state` to be a canonical MMR size (`2L - popcount(L)` recomputed with checked arithmetic) and returns the leaf count; the source calls it before any `leaf_to_pos`, and the target calls it on the element in `NonMerkRestorer::new` for symmetry. `state = u64::MAX, start = 2^63 - 1` is now a bounded `CorruptedData("... is not a valid MMR size")`. Note `2^63` *is* canonical (2^62 + 1 leaves) and remains a bounded "missing MMR leaf" read. Tests: `validate_mmr_size_accepts_canonical_and_rejects_others` (edge table) and `fetch_chunk_rejects_non_canonical_mmr_cursor` (source side, crafted cursors).

QuantumExplorer:
This is Claude. Re-reviewed #788 after merging `develop` (clean merge, `4e10d959`) and pushed `ab4e07a0` addressing the two open review findings plus three issues the re-review surfaced:

**Review findings**
- **P1 (page-entry cap on receipt)** — fixed; see thread.
- **P2 (MMR cursor arithmetic)** — fixed; see thread.

**New findings, fixed in the same commit**
1. **Frontier canonicality (Byzantine source → cost divergence on the synced node).** `CommitmentFrontier::deserialize` tolerates trailing bytes, and the target stored the wire frontier verbatim — so a padded frontier passed the state-root check (same Sinsemilla root) while changing `persisted_frontier_len`, which is what the GROVE_V4 frontier-save accounting bills against. The target now requires the frontier to round-trip byte-for-byte through the codec and to declare the element's tree size, and an **empty** commitment tree must not carry a frontier at all (its state root is a constant that never reads the payload, yet the next append would load the planted bytes). The legacy "trailing anchors" encoding predates the crate's first commit, so the strict check cannot reject honest data. Tests: `state_sync_commitment_tree_rejects_non_canonical_frontier`, `state_sync_empty_commitment_tree_rejects_planted_frontier` (both fail against the previous head).
2. **`PrivateDocumentStore` routing regression after the merge.** `develop` added PDS to `TreeType::uses_non_merk_data_storage()`, which this PR used as its routing predicate — so even an *empty* PDS would have broken state sync with an `InternalError` from `NonMerkRestorer::new`. Entry replay is now gated on a precise `supports_entry_replay` (the four append-only types); a populated PDS gets a descriptive `NotSupported` on both sides (discovery and `fetch_chunk`), an empty one keeps syncing via the Merk path exactly as on `develop`. Tests: `state_sync_empty_private_document_store_round_trip`, `state_sync_rejects_populated_private_document_store_up_front`.
3. **`grove_version` threading** for the #822/#825 `append` / `commit_mmr` / `push` / `get_root` signatures (bytes written are version-independent; sync discards costs).

Verification: `cargo test -p grovedb` 2864 passed / 0 failed; `cargo clippy -p grovedb --all-features -- -D warnings` and `cargo fmt --check` clean.

One pre-existing property left as-is (not introduced here): `fetch_chunk` serves every local chunk id packed into a global id, so a requester can ask for many pages per request; each page is budget-bounded and the Merk path has the same shape, so the absolute cap belongs at the transport.

PR #789, opened 2026-08-03
fix: model V4 gated stored-element reads in batch estimated costs

QuantumExplorer:

`GROVE_V4` gates `apply_batch.delete_tree_cleanup_type_source` and `apply_batch.overwrite_indexed_cleanup_inspection`, each charging one extra stored-element read per `DeleteTree` / overwrite-capable op in the **applied** batch path. The batch **estimated**-cost paths (`grovedb/src/batch/estimated_costs/{average,worst}_case_costs.rs`) did not model this read, so estimated costs undershot applied costs by exactly that read on affected ops at grove version >= 4. Surfaced by Dash Platform's identity-op fee tests at protocol v14 (grove v4): applied processing fees came out exactly one read above the v3-era values while estimated-mode paths still produced v3-era numbers.

## What was done?

Both estimators now charge the gated read, keyed off the same two version slots so V1..V3 estimates are byte-for-byte unchanged:

- **Overwrite gate** (`InsertOrReplace` / `Replace` / `Patch`): one average-/worst-case Merk node read of the layer's estimated element. Skipped when tree-override validation is on, for bare-reference elements, and for layers known to be empty — the same conditions under which the applied path skips the read or gets it for free (a get on an empty Merk costs nothing), which keeps the existing fresh-insert estimated==applied tests exact.
- **Delete-tree gate**: the parent storage-context prefix hash plus one seek loading the layered element cost — the exact `storage_loaded_bytes` shape `Element::get_from_storage` reports for tree elements. This read is deterministic (raw storage get, no Merk caching), so the estimated V4−V3 delta equals the applied V4−V3 delta field for field.

New tests (per gate, average and worst case):
- `test_average_case_delete_tree_type_source_read_estimated_vs_applied_parity` — runs the same batch applied and estimated on V3 and V4 and asserts the two deltas match exactly (seek, loaded bytes, storage cost, context-prefix hash calls).
- `test_average_case_overwrite_inspection_read_gated_v4_vs_v3` / `test_worst_case_overwrite_inspection_read_gated_v4_vs_v3` — pin the estimated delta to exactly one modeled node read, and pin the suppression conditions (validation on, bare reference) to a zero delta.
- `test_worst_case_delete_tree_type_source_read_gated_v4_vs_v3` — pins the worst-case delta to context hash + one worst-case layered load.

Existing exact-cost expectations that run at `latest()` (= V4) were updated by the modeled read, with comments documenting the delta; the `estimated == applied` seek assertion in `test_batch_root_one_tree_insert_op_under_element_average_case_costs` became `applied + 1 == estimated` because the applied path gets that particular read for free from the cached root node.

Also made `grovedb/src/estimated_costs/worst_case_costs` `pub(crate)` so the batch estimator can reuse `WORST_CASE_FLAGS_LEN`.

## How Has This Been Tested?

- `cargo test -p grovedb` — full crate suite green (2544 tests), including the 6 new/updated estimator tests.
- `cargo check -p grovedb --no-default-features --features minimal` and `--features verify` — feature-gated builds unaffected.
- `cargo clippy -p grovedb --all-targets` — no new lints.

## Breaking Changes

None on released versions: both charges are gated on the V4 feature slots, so V1..V3 estimated outputs are unchanged. V4 estimated outputs grow by exactly the read the V4 applied path already charges.

## Checklist:

### DIFF
--- a/grovedb/src/batch/estimated_costs/average_case_costs.rs
+++ b/grovedb/src/batch/estimated_costs/average_case_costs.rs
@@ -12,17 +12,22 @@ use grovedb_costs::{
 };
 #[cfg(feature = "minimal")]
 use grovedb_merk::estimated_costs::average_case_costs::{
-    add_average_case_merk_has_value, average_case_merk_propagate, EstimatedLayerInformation,
+    add_average_case_get_merk_node, add_average_case_merk_has_value, average_case_merk_propagate,
+    EstimatedLayerInformation,
 };
 use grovedb_merk::{
     element::tree_type::ElementTreeTypeExtensions, tree::AggregateData, tree_type::TreeType,
     RootHashKeyAndAggregateData,
 };
 #[cfg(feature = "minimal")]
+use grovedb_merk::{tree::kv::KV, tree_type::CostSize};
+#[cfg(feature = "minimal")]
 use grovedb_storage::rocksdb_storage::RocksDbStorage;
 #[cfg(feature = "minimal")]
-use grovedb_storage::worst_case_costs::WorstKeyLength;
+use grovedb_storage::{worst_case_costs::WorstKeyLength, Storage};
 use grovedb_version::version::GroveVersion;
+#[cfg(feature = "minimal")]
+use integer_encoding::VarInt;
 #[cfg(feature = "minimal")]
 use itertools::Itertools;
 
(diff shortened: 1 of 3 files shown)


QuantumExplorer (commented):
This will work

PR #793, opened 2026-08-13
feat: branched indexed-axis proofs — one envelope over N sibling prefix branches

QuantumExplorer:
## What
New proof shape for the indexed-axis family: `prove/verify_indexed_axis_query_branched` (range) and `prove/verify_indexed_axis_top_k_paginated_branched` (paginated top-k) attest one query over **N sibling prefix branches** — indexed trees whose paths differ at exactly one segment — in a **single envelope with a single reconstructed root hash**.

Motivation: Platform's compound ranked indexes lay one indexed tree per prefix value; an `IN` over the leading property reads N of them. Per-branch envelopes duplicate every ancestor layer N times and force callers to cross-check N root hashes (dashpay/platform#4401 currently ships a length-prefixed container of N envelopes as a stopgap — this primitive replaces it).

## How
- Layers **above** the branching level appear once (`shared_layer_proofs` + `shared_ancestor_attestations`).
- The branching level is **one multi-key Merk proof** binding every branch's value-tree element simultaneously.
- Each branch carries only its tail: layers below its key, primary/other-axes attestations, and its secondary proof.
- Verification reuses the audited single-path blocks (`verify_deepest_layer`, `walk_ancestor_chain`) on the tail and shared windows; each branch's reconstructed value-tree root must recompose to the `value_hash` the multi-key proof recorded for that key, so tails cannot be swapped, duplicated, or reordered. Prove-side reuses the existing envelope builders per branch and splits at the branching depth.
- **No storage or hash-composition changes.** New envelope types only, per the envelope module's "new shapes get new envelope types" rule.

## Proof shape

One envelope, drawn over the tree it attests (two branches shown). Solid arrows are the tree; dashed arrows are the verifier's upward reconstruction — note everything above the branching level is walked **once**:

```mermaid
flowchart TD
    ROOT(["**GroveDB root hash** — reconstructed once for the whole envelope"])
    SL0["shared layer 0 — single-key proof<br/><i>shared_ancestor_attestations[0]</i>"]
    SL1["shared layer 1 — single-key proof<br/><i>shared_ancestor_attestations[1]</i>"]
    BRK{{"branching Merk (prefix property-name tree)<br/><b>branching_layer_proof</b>: ONE multi-key proof<br/>binding keys {east, west} + their value_hashes"}}

    ROOT === SL0 === SL1 === BRK

    subgraph EAST["branch 0 — key <b>east</b> (BranchedProofBranch)"]
[...]

PR #794, opened 2026-08-13
feat: AxisPathQuery — query vocabulary for axis-ordered reads

QuantumExplorer:
## Why

An ordinary `PathQuery` **selects keys**: its items name keys or key ranges in the Merk a path points at. It cannot describe the other thing an indexed tree can answer — *"the best `k` groups by aggregate"* — because that ordering lives in the tree's per-axis **secondary**, which is keyed by `sort_key ‖ original_key` and is **not a path-addressable subtree**. The secondary is an internal structure of the element with its own storage prefix, so no path names it, and therefore no `PathQuery` (merged or not) can reach it.

The practical consequence: a caller wanting an axis-ordered answer left the query language entirely and called one of a dozen bespoke `indexed_{count,sum,avg}_{top_k,top_k_paginated,range,range_aggregate}` methods, each with its own argument list — and hand-built the secondary's Merk query when it wanted bounds. That last part is the real hazard: the bounds→Merk-query lowering is **prover/verifier agreement material**, and it was living in each caller (including downstream in Dash Platform) rather than in grovedb.

## What

`AxisPathQuery` gives axis-ordered reads the same shape everything else has — a path plus a description of what to read:

```rust
pub struct AxisPathQuery { path: Vec<Vec<u8>>, query: AxisQuery }

pub struct AxisQuery {
    axis: IndexAxis,             // Count | Sum | Avg
    traversal: AxisTraversal,    // TopK { k, offset } | Bounded { lo, hi, limit }
    descending: bool,
}
```

with three entry points:

```rust
db.query_axis_path_query(&q, tx, gv)   // read
db.prove_axis_path_query(&q, tx, gv)   // prove
GroveDb::verify_axis_path_query(proof, &q)  // verify (verifier-only builds included)
```

- **The bounds lowering moves in**: `AxisQuery::merk_query()` turns inclusive `[lo, hi]` into the secondary's byte range (inclusive at `lo`, exclusive at the *successor* of `hi`, open-ended when `hi` is the axis maximum). Both the prove and verify paths call it, so they cannot drift.
- **Validation is shared**: `k`/`limit` of zero, inverted bounds, bounds wholly outside the axis domain, and empty paths are rejected — so a caller error surfaces as an error rather than an empty page that looks like real absence.
- **`i128` bounds for every axis**, matching the convention the existing aggregate-range entry points already use, clamped per-axis in the lowering.

## This is dispatch, not a new proof shape

Each entry point routes to the existing indexed-axis primitive and produces the existing envelope. A test pins that literally:

```rust
assert_eq!(via_vocabulary, via_primitive,
    "the top-k vocabulary must emit the primitive's exact bytes");
```

for both traversals. No proof bytes change, no verification rule changes, and every existing caller is untouched — this is purely additive.

## Explicitly out of scope

Both are stated in the module docs so the boundary is on record:

1. **Merging sibling axis queries.** N axis path queries whose paths differ at one segment are exactly the branched-proof case (#793); merging belongs with that shape.
2. **Embedding axis queries inside a `PathQuery`'s subquery branches**, so one proof could mix key-selected and axis-ordered layers. That needs the general proof generator to carry a secondary *proof* where it currently carries only a secondary root *attestation* (see the `ProofBytes::CountIndexedTree` comment in `generate.rs`, which redirects secondary-ordered callers elsewhere — this PR is the front door it should redirect to).

## Testing

`axis_path_query_tests`: read/prove/verify round trips for both traversals; all three axes through one vocabulary; bounded inclusivity on both ends; the lowering compared against a hand-built secondary query (including the open-ended maximum case); byte-identity with the primitives; the degenerate-query rejection matrix across read and prove; a proof failing to verify under a different `k` / offset / direction / traversal; and bincode round trips including rejection of an unknown axis tag.

Full `grovedb` suite: **2573 passed**. Verifier-only build (`--no-default-features --features verify`) compiles. Clippy clean.

## Summary by CodeRabbit

* **New Features**
  * Added unified axis-based queries for count, sum, and average indexes.
  * Supports top-k pagination and bounded range queries in ascending or descending order.
  * Added read, proof generation, and proof verification workflows.
  * Query results now include entries, the reconstructed root hash, and skipped-item counts where applicable.

* **Bug Fixes**
  * Added validation for invalid paths, limits, bounds, and unsupported query values.

### DIFF
--- a/grovedb/src/operations/axis_path_query.rs
+++ b/grovedb/src/operations/axis_path_query.rs
@@ -1,0 +1,250 @@
+//! The three entry points for [`AxisPathQuery`]: read it, prove it,
+//! verify it.
+//!
+//! This is dispatch, not new proof machinery. Each entry point
+//! validates the query once, then routes to the indexed-axis primitive
+//! that already serves that shape — so an axis path query produces the
+//! same envelope, byte for byte, as the hand-rolled call it replaces.
+//! What the caller gains is one vocabulary instead of a dozen bespoke
+//! argument lists, and one place where the bounds-to-Merk-query
+//! lowering lives (see [`AxisQuery::merk_query`]) rather than a copy on
+//! each side of the prover/verifier boundary.
+
+#[cfg(feature = "minimal")]
+use grovedb_costs::{cost_return_on_error, cost_return_on_error_no_add, CostResult, CostsExt};
+use grovedb_element::indexed::IndexAxis;
+#[cfg(feature = "minimal")]
+use grovedb_version::version::GroveVersion;
+
+#[cfg(feature = "minimal")]
+use crate::TransactionArg;
+use grovedb_merk::tree::CryptoHash;
+
+use crate::{
+    operations::proof::indexed_axis::AxisEntries,
+    query::{AxisPathQuery, AxisTraversal},
+    Error, GroveDb,
+};
+
+/// The verified answer to an [`AxisPathQuery`].
+#[derive(Debug)]
+pub struct VerifiedAxisPathQuery {
+    /// GroveDB root hash the proof reconstructs. Compare it against the
+    /// root you trust — verification alone proves internal consistency,
+    /// not that the proof is about the state you meant.
+    pub root_hash: CryptoHash,
+    /// The entries, in the query's walk direction.
+    pub entries: AxisEntries,
+    /// Entries the walk attested as skipped before the returned page.
+    ///
+    /// Equals the requested `offset` on a full [`AxisTraversal::TopK`]
+    /// page, and is smaller when the walk ran out during the skip — in
+    /// which case `entries` is empty and the pair proves the secondary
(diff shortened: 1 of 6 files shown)


QuantumExplorer:
Closing alongside #793 in favour of the `SubqueryBranch` unification.

This PR's vocabulary is sound and the `AxisQuery` / `AxisTraversal` types plus the `merk_query()` bounds lowering should be lifted verbatim into that work — the lowering in particular belongs in grovedb rather than in Dash Platform, where it currently lives on the wrong side of the prover/verifier boundary. What gets subsumed is only the standalone `AxisPathQuery` pairing and its three entry points: the useful shape is not "axis query at path P" but "descend to /…/region/, then take top-k by count per region", which is a subquery branch, not a top-level path query.

Branch `feat/path-query-axis-vocabulary` is left in place (head d3248cfb).

_🤖 Closed via [Claude Code](https://claude.com/claude-code)_

PR #795, opened 2026-08-13
refactor: PathQuery::classify() — one shared shape decision for query dispatch

QuantumExplorer:
## Context

This is PR 1 of the **unified PathQuery** effort: making `PathQuery` express and serve every query shape (key selection, aggregates, count-offset pagination, and — in upcoming PRs — indexed-axis reads and sum-budget reads) through one classification shared by reader, prover, and verifier.

Today each entry point re-discovers the shape of the query it was handed by calling the `has_*` / `validate_*` helpers in its own order. Since prover and verifier must agree on what a query *means* (the proof does not carry the query), that discovery is consensus-relevant — and it currently exists in several copies.

## What this does

- **`PathQuery::classify() -> Result<PathQueryShape, Error>`** (new `grovedb/src/query/shape.rs`): pure (no DB access — a verifier holding only the query classifies identically to the prover), total (every query maps to exactly one shape or a typed `InvalidQuery`), and mirrors `prove_query_non_serialized`'s exact gate order (aggregate-count → aggregate-sum → combined → count-offset → key selection) so future call-site migrations keep their error surface byte-for-byte.
- **Shapes**: `KeySelection`, `CountOffsetPaginated`, `AggregateLeaf`/`AggregateCarrier` × `AggregateKind::{Count, Sum, CountAndSum}`. Axis and sum-budget shapes arrive with the vocabulary PR.
- **Behavior-preserving**: implemented entirely over the existing validators; nothing dispatches through it yet. The LOCKED V0/V1 prover internals are untouched.
- **grovedbg drift fix**: the debugger's mirrored `Query` lacked `add_parent_tree_on_subquery` and the converter hardcoded `false`, so debugger-issued queries could never exercise the flag. Added with `serde(default)` so older frontend payloads keep decoding.

## Tests

- Error-surface equality tests: for malformed aggregate / count-offset queries, `classify()` must return the exact error the corresponding validator returns.
- Shape tests for every family (leaf + carrier × 3 kinds), the limit/offset size-constraint split, `offset == Some(0)` treated as no offset, and mixed-aggregate rejection.
- Totality grid: item sets × subqueries × limits × offsets × path emptiness — classify never panics and only fails with `InvalidQuery`.
- Full `grovedb` + `grovedb-query` suites green; `--no-default-features --features verify` build green.

## Noted in passing (not fixed here)

While writing tests: `Query::insert_item` collision-merges an aggregate wrapper with an overlapping plain item and silently degrades it to a plain range (`AggregateCountAndSumOnRange(Range(a..z))` + `insert_key("extra")` → `items == [Range(a..z)]`). Pre-existing; will be filed/fixed separately.

## Summary by CodeRabbit

* **New Features**
  * Added query shape classification for key selection, pagination, and count, sum, or combined aggregate queries.
  * Exposed query shape and aggregate type information for broader query tooling support.
* **Bug Fixes**
  * Subqueries now correctly preserve the option to include parent tree data.
* **Compatibility**
  * Query payloads from older versions remain supported when the new option is absent.

### DIFF
--- a/grovedb/src/debugger.rs
+++ b/grovedb/src/debugger.rs
@@ -746,7 +746,7 @@ fn query_to_grovedb(query: Query) -> crate::Query {
             query.conditional_subquery_branches,
         ),
         left_to_right: query.left_to_right,
-        add_parent_tree_on_subquery: false,
+        add_parent_tree_on_subquery: query.add_parent_tree_on_subquery,
     }
 }
 
--- a/grovedb/src/lib.rs
+++ b/grovedb/src/lib.rs
@@ -243,8 +243,9 @@ use grovedb_version::version::GroveVersion;
 use grovedb_visualize::DebugByteVectors;
 #[cfg(any(feature = "minimal", feature = "verify"))]
 pub use query::{
-    aggregate_sum_path_query::AggregateSumPathQuery, GroveBranchQueryResult, GroveTrunkQueryResult,
-    LeafInfo, PathBranchChunkQuery, PathQuery, PathTrunkChunkQuery, SizedQuery,
+    aggregate_sum_path_query::AggregateSumPathQuery, AggregateKind, GroveBranchQueryResult,
+    GroveTrunkQueryResult, LeafInfo, PathBranchChunkQuery, PathQuery, PathQueryShape,
+    PathTrunkChunkQuery, SizedQuery,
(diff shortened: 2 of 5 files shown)


QuantumExplorer (commented):
Reviewed

PR #799, opened 2026-08-13
feat: axis-ordered reads embedded in the GroveDBProof V1 envelope

QuantumExplorer:
## Context

PR 4 of the **unified PathQuery** effort (stacked on #798 ← #797 ← #795) — the full-integration step the effort was pointed at: a `PathQuery` whose query node carries `ReadMode::Axis` now **proves and verifies through the general V1 proof walk**, via a new appended `ProofBytes::IndexedTreeAxisDescent` variant whose payload carries a proof over the queried per-axis **secondary** in place of the primary descent.

## Design

**Query-as-input, no echoes.** The payload carries no traversal parameters — the verifier resolves axis, bounds, direction, and caps from the query it independently holds, through the new `PathQuery::axis_read_at_path` (the same resolver the prover uses, so prover and verifier cannot disagree about which layers are axis reads). One exception: `RankOfKey` carries the prover-computed rank, required to drive the count-offset verification walk — the count commitments attest it, and the single yielded entry must be the queried key.

**Recomputed attestation.** Where the primary-descent shape (`ProofBytes::CountIndexedTree`) supplies the 32-byte secondary-root attestation raw (safe there because it enters the `combine_hash_three` preimage), the axis descent **recomputes** it from the carried secondary proof, then rebuilds the third combine input — the recomputed root for PCIT/PSIT, or `axes_digest(other_axes + recomputed queried-axis root)` for PCPSIT, family-checked against axis-relabel forgery — and requires the parent-committed `value_hash` to match. `primary_root_hash` / `other_axes_root_hashes` remain prover-supplied but bound: forging either fails the binding.

**Branched = the ordinary walk.** Branch keys are `Key` items at the branching layer (one multi-key merk proof; absence proven natively by merk), each present branch descends the shared suffix to its own axis-descent terminal, and shared-prefix layers are naturally deduplicated by `LayerProof` nesting. The verifier rejects a proof that shows a branch key **present** while omitting its axis layer — hiding entries behind fake absence fails closed.

**Surface.**
- `prove_query` serves axis shapes (V1 envelope only; V0 refusal follows the ACOR template)
- **`verify_path_query`** (new, verify-feature-reachable): the unified verify entry returning a typed `VerifiedPathQuery` for every provable shape — key selection, all six aggregate forms, axis entries / rank / range-aggregate, and branched with authenticated absence. The `verify_query` family stays key-selection-only.
- New `proof.axis_descent_in_v1_envelope` slot (0 in V1–V3, 1 in V4), read by **both** sides: below V4 the prover refuses and the verifier rejects, mirroring the fail-closed version-2 `Query` decode.
- The rank computation and the attestation-binding core are **factored out of** the standalone-envelope machinery (`compute_indexed_axis_rank_of_key`, `recompute_axis_binding_digest`) and shared — one copy of the consensus-relevant logic, not two. The standalone envelopes and their ~26 entry points are untouched and remain first-class.

## Tests

- **Round trips**: top-k, paginated (attested skip), bounded (shared lowering, successor-exclusive upper bound), rank in both directions, range aggregate; PCPSIT three-axis digest reconstruction; branched with a present/absent branch mix.
- **Cross-checks**: verified entries ≡ trusted reads ≡ standalone-envelope results, with matching reconstructed root hashes.
- **Forgeries** (all rejected): forged `primary_root_hash`; relabeled `axis_tag`; tampered secondary-proof bytes; smuggled `other_axes` on a single-secondary target; lying rank echo; stripping a present branch's descent to fake absence.
- **Gates**: V3 prover refuses, V3 verifier rejects a genuine V4 proof, V0 envelope + axis query rejected; `verify_path_query` also round-trips plain key selection.
- Full grovedb suite green (2600+ tests), clippy clean, `--no-default-features --features verify` build green.

## Review focus

The consensus-adjacent surface is `verify_axis_descent_layer` (verify.rs) and the `axis_read_at_path` resolver — an adversarial pass on those two is the highest-value review. A dedicated security-audit pass is being run and its findings will be posted on this PR.

## Summary by CodeRabbit

* **New Features**
  * Added authenticated proofs for indexed-tree axis queries, including ranked pages, bounded reads, rank lookups, and count/sum aggregates.
  * Added a unified path-query proof verification API with typed results and root-hash access.
  * Added support for branched axis reads, including authenticated absent branches.
  * Added version-gated support for axis descents in the latest proof format.

* **Bug Fixes**
  * Corrected count-range documentation and aggregation semantics.
  * Improved handling of missing axis branches and invalid query shapes.
  * Added validation for malformed, forged, duplicate, or incompatible proofs.

### DIFF
--- a/grovedb/src/operations/proof/generate.rs
+++ b/grovedb/src/operations/proof/generate.rs
@@ -1407,6 +1407,43 @@ impl GroveDb {
             )
         );
 
+        // A single-path axis read has exactly one answer — the axis
+        // descent at the queried path. The generic walk cannot produce
+        // one when the target is missing or is not an indexed tree; it
+        // returns `Ok` with an ordinary (or empty) layer instead, and
+        // the verifier then rejects the result as "must verify exactly
+        // one axis layer, got 0". Fail generation here instead, so the
+        // prover never hands out a proof that cannot answer the query
+        // it was asked.
+        //
+        // Branched axis reads are deliberately excluded: an absent
+        // branch key legitimately produces no descent, and its absence
+        // is what the branching-level Merk proof authenticates.
+        if matches!(
+            path_query.classify(),
+            Ok(crate::PathQueryShape::AxisRead { .. })
+        ) {
+            fn count_axis_descents(layer: &LayerProof) -> usize {
+                usize::from(matches!(
+                    layer.merk_proof,
+                    ProofBytes::IndexedTreeAxisDescent(_)
+                )) + layer
+                    .lower_layers
+                    .values()
+                    .map(count_axis_descents)
+                    .sum::<usize>()
+            }
+            let descents = count_axis_descents(&root_layer);
+            if descents != 1 {
+                return Err(Error::InvalidPath(format!(
+                    "a single-path axis read must produce exactly one axis descent at the \
+                     queried path, but the walk produced {descents} — the path does not \
+                     name an indexed tree carrying that axis"
+                )))
+                .wrap_with_cost(cost);
+            }
+        }
+
         Ok(GroveDBProof::V1(GroveDBProofV1 { root_layer })).wrap_with_cost(cost)
     }
 
--- a/grovedb/src/operations/proof/generate.rs
+++ b/grovedb/src/operations/proof/generate.rs
@@ -109,12 +109,35 @@ impl GroveDb {
         prove_options: Option<ProveOptions>,
         grove_version: &GroveVersion,
     ) -> CostResult<GroveDBProof, Error> {
-        // Read-mode gate: axis / sum-budget reads are not served by this
-        // prover. Fail closed rather than misreading such a query as key
-        // selection and returning a proof about the wrong thing.
-        if let Err(e) = path_query.reject_unserved_read_mode() {
-            return Err(e).wrap_with_cost(OperationCost::default());
-        }
+        // Read-mode dispatch. Axis shapes are served by the V1 envelope
+        // — validated here (classify runs the full shape grammar) and
+        // gated below once `prove_version` is known. Sum-budget shapes
+        // have no proof form yet. Anything malformed fails closed
+        // rather than being misread as key selection.
+        let is_axis_shape = if path_query.query.query.has_read_mode_anywhere() {
+            match path_query.classify() {
+                Ok(crate::PathQueryShape::AxisRead { .. })
+                | Ok(crate::PathQueryShape::BranchedAxisRead { .. }) => true,
+                Ok(crate::PathQueryShape::SumBudget { .. }) => {
+                    return Err(Error::NotSupported(
+                        "sum-budget path queries have no proof form yet; prove the underlying \
+                         items with a key-selection query and fold client-side, or use the \
+                         trusted read (run_path_query)"
+                            .to_string(),
+                    ))
+                    .wrap_with_cost(OperationCost::default());
+                }
+                Ok(_) => {
+                    return Err(Error::CorruptedCodeExecution(
+                        "a read-mode-bearing query classified as a non-read-mode shape",
+                    ))
+                    .wrap_with_cost(OperationCost::default());
+                }
+                Err(e) => return Err(e).wrap_with_cost(OperationCost::default()),
+            }
+        } else {
+            false
+        };
         // Aggregate-count gate: validate at entry so malformed ACOR
         // queries (invalid inner range, ACOR-hidden-in-subquery, etc.) are
         // rejected up front instead of being skipped when the recursive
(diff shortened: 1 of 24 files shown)


QuantumExplorer:
This is Claude. A dedicated adversarial security audit ran against this PR's diff (blockchain-security-auditor pass, as promised in the description). Results:

**Two real findings, both now fixed in `c6b2283e` with regression tests:**

1. **Branched absence forgery (Medium)** — the axis-read check sat inside the lower-layer lookup, so stripping the axis-descent layer of a branch whose indexed tree exists but is *empty* (`is_non_empty_tree()` is false for it, so no existing guard fired) let a prover obtain a verified `None` ("proven absent") slot under the genuine root hash. Fixed by hoisting the axis check above the lookup: at an axis-read position a missing layer is now a hard `InvalidProof`, and an existing empty indexed tree reports `Some(empty entries)` — matching the honest prover and the trusted read. The branched absence guard in `verify_path_query` also widened to reject any present trio under the branch subtree, and `run_path_query`'s branched read now resolves the full suffix chain so read and proof agree on `None` at every absence level.

2. **`Bounded` traversal vs empty secondary (Medium, liveness)** — `Merk::prove` refuses empty trees, so a bounded axis proof over a freshly created indexed tree hard-errored while every other traversal (and the trusted read) handled it. Both sides now use the same empty convention as count-offset and aggregate-on-range: empty proof bytes ⇒ `NULL_HASH` secondary root, accepted only when the element genuinely commits an empty secondary.

3. **Duplicate branch keys (Low)** — `classify()` now rejects a branched read naming the same branch key twice (a valid proof could otherwise yield contradictory `Some`/`None` rows for one key).

**Confirmed defended** (with defending lines traced): forged `primary_root_hash`, relabeled `axis_tag` (both the tag check and the element-family check), tampered secondary bytes (root recomputed), smuggled `other_axes` on single-secondary targets, PCPSIT axes reorder/dup (sorted-strict tag check + length-prefixed `axes_digest`), lying/overflowing rank echoes (checked_add + count commitments + yielded-key check), single-path layer stripping and path spoofing, axis payloads at non-axis positions (three arms), nested layers under a descent, prover/verifier resolver divergence (single shared `axis_read_at_path`, conditional branches unreachable by grammar), envelope downgrades (three coordinated gates), decode DoS (16MiB payload cap, ≤256 axes, depth caps), and zero behavior change to the LOCKED V0 path and existing V1 key selection.

One documented non-issue: the bounded secondary executes with the lenient proof version, matching the standalone envelope; safe because the axis decoders consume only proof *keys* (bound into the recomputed root), never values — now stated in a code comment so a future value-consuming change doesn't inherit it silently.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

QuantumExplorer (commented):
Reviewed

QuantumExplorer (commented):
I found two correctness issues in the new axis-proof flow; details inline.

grovedb/src/operations/proof/verify.rs
```diff
@@ -828,6 +1217,50 @@ impl GroveDb {
...
+                    // axis descent — a missing layer is a hard error, not
+                    // a silent skip (which the branched shape would
+                    // endorse as a `None`, i.e. proven-absent, slot).
+                    //
+                    // This also makes an empty indexed tree report an
+                    // empty axis result (verify_axis_descent_layer proves
+                    // it against a NULL-hash secondary), matching the
+                    // honest prover and the trusted read rather than
+                    // reporting the branch absent.
+                    if element.is_indexed_tree() {
```
QuantumExplorer: **[P1] Reject wrong-type axis terminals**

Resolve `axis_read_at_path` before filtering on `element.is_indexed_tree()`. In a branched query, a present normal tree at the axis terminal skips this check and is omitted from the trio results, so `verify_path_query` reports the branch as `None`. I reproduced this with `scores = Element::empty_tree()`: the proof verified as absent while `run_path_query` returned `InvalidPath("Sum axis not indexed at this path")`. Please require that any path governed by an axis read contains a compatible indexed tree before dispatch.

grovedb/src/operations/proof/generate.rs
```diff
@@ -1958,6 +2010,81 @@ impl GroveDb {
...
+                            // query node governing this element carries
+                            // ReadMode::Axis, so instead of descending the
+                            // primary the layer carries an axis-descent
+                            // payload — a proof over the queried per-axis
+                            // secondary. Matched before the primary-descent
+                            // arms below so an axis read can never be
+                            // silently served as a primary descent; matches
+                            // empty primaries too (the payload commits
+                            // NULL_HASH roots naturally).
+                            Ok(Element::ProvableCountIndexedTree(..))
```
QuantumExplorer, line 2060: **[P2] Do not emit unverifiable single-axis proofs**

For a single-path axis query whose target is missing or non-indexed, this arm never runs, but the generic proof walk still returns `Ok` with no axis-descent layer. I reproduced a missing target where `prove_query` succeeded and `verify_path_query` rejected the result with “must verify exactly one axis layer, got 0.” Please preflight the single-axis target or fail generation when no descent is produced.

PR #800, opened 2026-08-13
feat: provable sum-budget reads — SumBudgetWindow in the GroveDBProof V1 envelope

QuantumExplorer:
## Context

PR 5 of the **unified PathQuery** effort (stacked on #799 ← #798 ← #797 ← #795). Sum-budget path queries — `ReadMode::SumBudget`: walk items in key order, stop when the running **net** sum reaches the budget — gain a proof form, completing proof coverage for every read-mode shape.

## Design

**The window is the proof.** A new appended `ProofBytes::SumBudgetWindow` variant carries `{ exhausted, window_len, merk_proof }`: an ordinary Merk proof over exactly the window of elements the budget walk scanned. No new cryptographic machinery — the parent binding falls out of the ordinary `combine_hash(H(value), child_root)` tree-descent check, and merk limit-proof semantics already attest "these are the first N elements in walk order".

**The fold replay is the attestation.** The verifier executes the window proof with the query's *own* items and direction — limited to the claimed window on a stop, **unlimited on claimed exhaustion** (so the proof itself must prove the range end) — then replays the read engine's exact fold arithmetic element by element: saturating net-budget subtraction (negative sum items give budget back), the per-match limit, and the grove-version global scan cap including its counted-but-unprocessed tripping element. A window that continues past a fired stop, stops short of one, or misstates exhaustion is rejected. A plain Merk descent at a sum-budget position is also rejected, so the shape can never be silently served as key selection.

**Provable fold semantics, pinned.** The provable shape skips non-sum elements and skips references — the two behaviors a single-subtree window proof can replay deterministically (a reference's target lives outside the window). `run_path_query`'s trusted read switches to the same options, so read ≡ verified over any state; the legacy `AggregateSumPathQuery` surface keeps its configurable options untouched.

**Engine-truth corrections** made while pinning the replay against `element/aggregate_sum_query`:
- `SumBudgetRead.max_items_checked` → **`match_limit`**: the engine decrements this per **matched result**, not per scanned element — the PR #797 name and doc were wrong (wire layout unchanged; the stack is unmerged so no compat concern)
- `sum_limit > i64::MAX` now rejected at validation (the engine's budget arithmetic is signed and errors there anyway)
- `AggregateSumQueryResult` gains `elements_scanned` — the window size the prover needs, and a useful read-API datum in its own right

**Gating**: new `proof.sum_budget_in_v1_envelope` slot (0 in V1–V3, 1 in V4) read by both sides; V0-envelope refusal follows the ACOR template. `verify_path_query` returns `VerifiedPathQuery::SumBudget { matches, total, stop }` with the replay-attested `SumBudgetStop` (`BudgetReached` / `MatchLimitReached` / `HardScanCapReached` / `Exhausted`).

## Tests

- Round trips for all stop conditions: budget (incl. negative values giving budget back), match limit, exhaustion, and skip-semantics windows containing foreign elements
- Read ≡ verified agreement across budget configurations
- Forgeries: understated window (claiming an earlier stop), exhaustion lies in both directions, plain-descent substitution at the sum-budget position
- V3-refuses / V4-serves gates on both prover and verifier
- Full suites green, clippy clean, verify-only build green

### DIFF
--- a/grovedb-query/src/read_mode.rs
+++ b/grovedb-query/src/read_mode.rs
@@ -38,13 +38,17 @@ use crate::{axis_query::AxisQuery, error::Error, query::Query};
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct SumBudgetRead {
-    /// Stop once the running sum of matched sum-item values reaches
-    /// this. Distinct from a result-count limit: how many entries that
-    /// takes depends on the data.
+    /// Stop once the running **net** sum of matched sum-item values
+    /// reaches this. Distinct from a result-count limit: how many
+    /// entries that takes depends on the data (and negative values give
+    /// budget back). Must fit in `i64` — the budget arithmetic is the
+    /// engine's signed saturating subtraction.
     pub sum_limit: u64,
-    /// Cap on elements scanned (matched or skipped), on top of the
-    /// grove-version global scan cap. `None` = only the global cap.
-    pub max_items_checked: Option<u16>,
+    /// Stop after this many **matched** sum items, regardless of
+    /// budget. `None` = no match cap. (Elements scanned but skipped —
+    /// non-sum elements, references — do not count; the grove-version
+    /// global scan cap bounds those separately.)
+    pub match_limit: Option<u16>,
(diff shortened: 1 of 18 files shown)


QuantumExplorer (commented):
Reviewed

PR #801, opened 2026-08-14
feat: version the indexed-axis family; fallible, direction-aware merges

QuantumExplorer:
## Context

PR 6 of the **unified PathQuery** effort (stacked on #800 ← #799 ← #798 ← #797 ← #795). Two coordinated **breaking changes** for the rs-drive workspace bump — the last code PR of the sequence before the book chapter.

## Indexed-axis versioning

The standalone indexed-axis family shipped entirely outside `grove_version`: no slots, and verifiers that take no `GroveVersion` at all. This PR brings it in:

- New `GroveDBOperationsIndexedAxisVersions { read, prove_single_path, verify_single_path }` — all `0` in V1–V4. Availability isn't the point (indexed trees can't exist in pre-V4 data anyway); the slots exist so the **first future divergence bumps a number instead of forking silently**.
- The five generic `verify_indexed_axis_*` functions and their twelve per-axis wrappers gain a `GroveVersion` parameter + check — the breaking part (~205 call sites migrated in-repo; precedent: the aggregate verifiers already take it under verify-only builds).
- The five provers gain the `prove_single_path` gate; the trusted-read helpers gain the `read` gate at their generic chokepoints.

## Merge semantics

- **`Query::merge_multiple` / `merge_with` become fallible**: both reject inputs carrying a `ReadMode` at any nesting level — axis and sum-budget reads have no defined merge semantics, and silently merging them as key selection would change what a query means. The merge bodies move to private `_unchecked` twins, so the recursive internals (whose inputs are descendants of already-checked queries) stay infallible — one check at the public boundary, no ripple.
- New `Query::merge_multiple_directional`: additionally requires top-level direction agreement.
- **`PathQuery::merge` direction fix behind `path_query_methods.merge = 1` (GROVE_V4)**: every input must agree on `left_to_right` (typed error on conflict) and the shared direction propagates to the merged root. The fix matters because the historic behavior is worse than the known "first wins" description: for inputs at different paths, the merged root is a **synthesized** query whose direction is the default — input directions were dropped entirely. V1–V3 keep that behavior exactly (pinned by test); merged queries feed proofs and the verifier re-runs the same merge with the same grove version, so both sides agree at every version.

## Tests

Direction agreement / conflict / propagation at V4; the pre-V4 quirk pinned as-is; Query-level read-mode merge rejection (direct and nested); unknown-slot rejection via a doctored `GroveVersion`. Full suites green, clippy clean, verify-only build green (one `cfg` fix on the known-fragile verify boundary, caught by the required build).

## Summary by CodeRabbit

* **New Features**
  * Added version-aware indexed-axis reads, proof generation, and proof verification.
  * Added directional query merging with validation that merged queries use matching directions.
  * Added versioned controls for indexed-axis operations and query-merge behavior.

* **Bug Fixes**
  * Query merges now reject unsupported read modes, including nested occurrences.
  * Conflicting merge directions and unknown merge versions now return clear errors.
  * Improved validation prevents unsupported indexed-axis operations from running.

### DIFF
--- a/grovedb-query/src/merge.rs
+++ b/grovedb-query/src/merge.rs
@@ -16,16 +16,49 @@ impl SubqueryBranch {
             (None, Some(subquery)) => Some(subquery),
             (Some(subquery), Some(other_subquery)) => {
                 let mut merged_subquery = subquery.clone();
-                merged_subquery.merge_with(*other_subquery);
+                merged_subquery.merge_with_unchecked(*other_subquery);
                 Some(merged_subquery)
             }
         }
     }
 
+    /// Whether either side of a branch merge carries a read mode
+    /// anywhere in its subquery. Branch merges combine subqueries by
+    /// field, which cannot express "both of these are axis reads" — so
+    /// the public entry points refuse rather than drop the mode.
+    fn branch_merge_carries_read_mode(&self, other: &Self) -> bool {
+        let carries = |branch: &Self| {
+            branch
+                .subquery
+                .as_deref()
+                .is_some_and(|subquery| subquery.has_read_mode_anywhere())
+        };
+        carries(self) || carries(other)
+    }
+
     /// Merges two subquery branches, combining their subquery paths and
     /// subqueries. When paths differ, creates conditional subqueries to
     /// preserve both branches.
-    pub fn merge(&self, other: &Self) -> Self {
+    ///
+    /// Errors if either side carries a [`ReadMode`](crate::ReadMode)
+    /// anywhere: merging combines subqueries field by field, and the
+    /// mode is not one of the combined fields, so a merge would answer
+    /// key selection where the caller asked for an axis or sum-budget
+    /// read.
+    pub fn merge(&self, other: &Self) -> Result<Self, crate::error::Error> {
+        if self.branch_merge_carries_read_mode(other) {
+            return Err(crate::error::Error::NotSupported(
+                "can not merge subquery branches carrying read modes (axis / sum-budget reads)"
+                    .to_string(),
+            ));
+        }
+        Ok(self.merge_unchecked(other))
+    }
+
+    /// The merge body, private: every public path checks read modes
+    /// first, and the recursive internals only ever see branches whose
+    /// descendants passed that check.
+    fn merge_unchecked(&self, other: &Self) -> Self {
         match (&self.subquery_path, &other.subquery_path) {
             (None, None) => {
                 // they both just have subqueries without paths
@@ -81,7 +114,7 @@ impl SubqueryBranch {
                             Some(left_path_leftovers)
                         };
                         merged_query.insert_key(left_top_key.clone());
-                        merged_query.merge_conditional_boxed_subquery(
+                        merged_query.merge_conditional_boxed_subquery_unchecked(
                             QueryItem::Key(left_top_key),
                             SubqueryBranch {
                                 subquery_path: maybe_left_path_leftovers,
@@ -96,7 +129,7 @@ impl SubqueryBranch {
                         };
 
                         merged_query.insert_key(right_top_key.clone());
-                        merged_query.merge_conditional_boxed_subquery(
+                        merged_query.merge_conditional_boxed_subquery_unchecked(
                             QueryItem::Key(right_top_key),
                             SubqueryBranch {
                                 subquery_path: maybe_right_path_leftovers,
@@ -121,7 +154,7 @@ impl SubqueryBranch {
                         merged_query.insert_key(first_key.clone());
                         // our subquery stays the same as we didn't change level
                         // add a conditional subquery for other
-                        merged_query.merge_conditional_boxed_subquery(
+                        merged_query.merge_conditional_boxed_subquery_unchecked(
                             QueryItem::Key(first_key),
                             SubqueryBranch {
                                 subquery_path: maybe_left_path_leftovers,
@@ -146,7 +179,7 @@ impl SubqueryBranch {
                         merged_query.insert_key(other_first.clone());
                         // our subquery stays the same as we didn't change level
                         // add a conditional subquery for other
-                        merged_query.merge_conditional_boxed_subquery(
+                        merged_query.merge_conditional_boxed_subquery_unchecked(
                             QueryItem::Key(other_first),
                             SubqueryBranch {
                                 subquery_path: maybe_right_path_leftovers,
@@ -184,7 +217,7 @@ impl SubqueryBranch {
                 };
                 // our subquery stays the same as we didn't change level
                 // add a conditional subquery for other
-                merged_subquery.merge_conditional_boxed_subquery(
+                merged_subquery.merge_conditional_boxed_subquery_unchecked(
                     // there are no conditional subquery branches yes
                     QueryItem::Key(our_top_key),
                     SubqueryBranch {
@@ -220,7 +253,7 @@ impl SubqueryBranch {
                 };
                 // their subquery stays the same as we didn't change level
                 // add a conditional subquery for other
-                merged_subquery.merge_conditional_boxed_subquery(
+                merged_subquery.merge_conditional_boxed_subquery_unchecked(
                     // there are no conditional subquery branches yes
                     QueryItem::Key(their_top_key),
                     SubqueryBranch {
@@ -245,7 +278,7 @@ impl Query {
     ) {
         if let Some(current_subquery) = self.default_subquery_branch.subquery.as_mut() {
             if let Some(other_subquery) = other_default_branch_subquery {
-                current_subquery.merge_with(*other_subquery);
+                current_subquery.merge_with_unchecked(*other_subquery);
             }
         } else {
             // None existed yet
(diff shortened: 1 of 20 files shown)


QuantumExplorer:
This is Claude. Rebased onto `develop` now that #799 and #807 have landed — the branch was previously stacked on the pre-squash #799/#800 commits.

The commit replayed cleanly; two call-site fixes were needed against the new base, both consequences of this PR's own change meeting work that landed since:

- `axis_descent_proof_tests.rs` (added in #799) called `verify_indexed_axis_rank_of_key` with 6 args — this PR adds the `GroveVersion` parameter, so the new avg-axis rank test needed it threaded through.
- `merge_versioning_tests.rs` constructed `Some(ReadMode::Axis(..))`; develop boxes the field (`Option<Box<ReadMode>>`, the `large_enum_variant` fix), so it is now `Some(Box::new(..))`.

Validation on the rebased head: `cargo test --workspace --all-features` green, `cargo clippy --workspace --all-features -- -D warnings` clean, `cargo build --no-default-features --features verify -p grovedb` clean.

Ready for review. #802 (book chapter) is next in the stack and will be rebased on top once this merges.

QuantumExplorer:
This is Claude. Addressed the codecov failure — patch coverage **59.4% → 98.8%** (170/172 new lines).

The gap was that this PR adds fourteen version gates but tested one of them. `indexed_axis_version_slots_are_wired` covered `verify_indexed_axis_top_k` only; the other thirteen `check_grovedb_v0!` rejection branches never executed. That is exactly the failure mode the slots exist to prevent — one unchecked entry point is a silent fork the day a slot diverges — so the fix is to exercise every one:

- `every_trusted_read_rejects_an_unknown_read_version` — all 5 `read`-slot gates (`indexed_sum_top_k`, `..._paginated`, `indexed_sum_range`, and both `*_range_aggregate` readers), plus a positive call at the real version so a broken fixture can't masquerade as a passing gate.
- `every_prover_rejects_an_unknown_prove_version` — all 5 `prove_single_path` gates.
- `every_verifier_rejects_an_unknown_verify_version` — all 4 `verify_single_path` gates, asserting the *version* error rather than a decode failure (the gate fires before decoding, so garbage bytes are the right input).
- `merge_rejects_an_unknown_merge_version`, `merge_refuses_limits_and_offsets_at_every_merge_version`, `merge_surfaces_read_mode_conflicts_at_both_merge_versions` — `PathQuery::merge`'s own version handling at both merge versions.

Per-file: `indexed_tree.rs` 5→0 uncovered gate lines, `indexed_axis/verify.rs` 91.8%→94.2%, `indexed_axis/generate.rs` 84.8%→87.6%, `query/mod.rs` 95.9%→96.6%.

Two new lines remain uncovered — the `.map_err` closures on the version-gated `Query::merge_multiple`/`merge_multiple_directional` dispatch. They are unreachable through `PathQuery::merge`, which refuses read-mode queries at an earlier gate before that dispatch runs; the Query-level refusal behind them is defense in depth for direct rs-drive callers and is covered separately by `query_level_merges_reject_read_modes`. Writing the initial version of that test I described it as routing through the Query-level merge — it does not, and the comment now says so.

Validation: `cargo test --workspace --all-features` green, `cargo clippy --workspace --all-features -- -D warnings` clean, `cargo build --no-default-features --features verify -p grovedb` clean.

grovedb-query/src/merge.rs
```diff
@@ -16,7 +16,7 @@ impl SubqueryBranch {
             (None, Some(subquery)) => Some(subquery),
             (Some(subquery), Some(other_subquery)) => {
                 let mut merged_subquery = subquery.clone();
-                merged_subquery.merge_with(*other_subquery);
+                merged_subquery.merge_with_unchecked(*other_subquery);
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed — the drop was real. I reproduced it before changing anything: `SubqueryBranch::merge` with a read-mode subquery on the incoming side returned `read_mode: None`, because `merge_with_unchecked` destructures `read_mode: _`.

All four public branch-merge entry points now validate and return `Result`, matching the checked/unchecked split this PR already uses for `merge_multiple`/`merge_with`:

- `SubqueryBranch::merge` — checks both sides (dropping either is equally wrong), delegates to a private `merge_unchecked`.
- `Query::merge_default_subquery_branch` — checks the existing and incoming default branch.
- `Query::merge_conditional_boxed_subquery` and `merge_conditional_subquery_branches_with_new_at_query_item` — you flagged the first two; these are the same class, so they are covered too.

The 18 internal recursive call sites now target the `_unchecked` twins, since the public wrappers validated at the boundary. The contract comment is rewritten to enumerate all seven public entry points and say explicitly that a new wrapper skipping the check reintroduces this bug — the bodies destructure `read_mode` away by design.

Tests added in `merge_coverage.rs`, both `ReadMode::Axis` and `ReadMode::SumBudget` for every entry point (separate variants; an Axis-only check would pass while dropping sum-budget reads), plus a nested-mode case and positive cases proving plain branches still merge.

One caller consequence: `PathQuery::merge` propagates the new error instead of discarding it. It cannot fire today — the read-mode gate at the top of `merge` rejects such inputs first — but discarding a `Result` there would be the same silent drop one level up.

`cargo test --workspace --all-features` green (61 in merge_coverage, +4), clippy `-D warnings` clean, verify-only build clean.

QuantumExplorer (commented):
Reviewed

QuantumExplorer (commented):
I found two correctness gaps in the new merge and indexed-axis version boundaries; details inline.

grovedb-query/src/merge.rs
```diff
@@ -600,16 +739,45 @@ impl Query {
...
+    /// Errors if the incoming branch carries a
+    /// [`ReadMode`](crate::ReadMode) anywhere — see
+    /// [`SubqueryBranch::merge`] for why a branch merge cannot carry
+    /// one through.
     pub fn merge_conditional_boxed_subquery(
         &mut self,
         query_item_merging_in: QueryItem,
         subquery_branch_merging_in: SubqueryBranch,
+    ) -> Result<(), crate::error::Error> {
+        if subquery_branch_merging_in
```
QuantumExplorer: **[P2] Reject existing read modes in conditional merges**

This guard only checks the incoming branch. If `self.conditional_subquery_branches` already contains an overlapping Axis or SumBudget read, the code calls `merge_unchecked` and returns `Ok(())`, despite the documented invariant that every public merge boundary rejects read modes. I reproduced this by adding an Axis conditional and merging a plain branch at the same key. Check `self.has_read_mode_anywhere()` here, and inspect the existing map in `merge_conditional_subquery_branches_with_new_at_query_item`, before entering the unchecked machinery.

grovedb/src/operations/indexed_tree.rs
```diff
@@ -1378,6 +1386,10 @@ impl GroveDb {
     where
         B: AsRef<[u8]> + 'b,
     {
+        grovedb_version::check_grovedb_v0_with_cost!(
```
QuantumExplorer, line 1389: **[P2] Gate inverted ranges before returning**

The public count, sum, and avg range wrappers return `Ok([])` for `lo > hi` before reaching this new version check. With `operations.indexed_axis.read = 9`, I reproduced all three APIs succeeding instead of returning `VersionError`, leaving part of each entry point outside the new version contract. Perform the version check before the wrappers' degenerate-range fast path, and add inverted-bound cases to the unknown-version tests.

grovedb-query/src/merge.rs
QuantumExplorer: Agreed, and fixed. Reproduced first: adding an Axis conditional via `add_conditional_subquery` then merging a plain branch at the same key returned `Ok(())` and walked straight into the unchecked machinery.

The guard was one-sided — I checked both sides on `SubqueryBranch::merge` and `merge_default_subquery_branch` but only the incoming side on the two conditional entry points, which is exactly the inconsistency you spotted. Both now reject when either side carries a mode: `merge_conditional_boxed_subquery` checks `self.has_read_mode_anywhere()`, and `merge_conditional_subquery_branches_with_new_at_query_item` inspects the existing branch map it is handed.

Worth recording why the existing side matters, since the same-key repro happens to preserve the mode rather than drop it: an overlapping item splits both sides' branches and recombines them through `merge_with_unchecked`, so an existing read mode can end up governing a different key range than the caller established. Tests cover same-key, overlapping-range, and the associated-fn form, for both `ReadMode::Axis` and `ReadMode::SumBudget`, plus a positive case proving plain conditional merges still work.

grovedb/src/operations/indexed_tree.rs
QuantumExplorer: Agreed, and fixed. Reproduced first: with `operations.indexed_axis.read = 9`, `indexed_sum_range(path, 100, 0, ..)` returned `Ok([])` instead of `VersionError` — the wrapper's `lo > hi` fast path returns before it ever delegates to the gated generic, so inverted bounds sat outside the version contract every other input to that entry point is held to.

The version check now runs before the degenerate-range fast path in all three wrappers (`indexed_count_range`, `indexed_sum_range`, `indexed_avg_range`). The generic keeps its own check for direct callers; a second check of the same slot is idempotent.

The two `*_range_aggregate` readers already gated before their fast path, so they were correct — I added inverted-bound cases for them too, so a future refactor cannot reintroduce the gap there. The new test also pins that an inverted range at the REAL version still answers empty rather than erroring: the gate moved, the semantics did not.

PR #802, opened 2026-08-14
docs: book chapter for the unified PathQuery

QuantumExplorer:
## Context

PR 7 — the final PR of the **unified PathQuery** effort (stacked on #801 ← #800 ← #799 ← #798 ← #797 ← #795).

Adds **The Unified PathQuery** chapter to the mdBook, placed right after The Query System: read modes behind the `Query` encoding's version byte, the three read-mode shapes and their constructors, `classify()` as the one consensus-relevant shape decision, `run_path_query` / `verify_path_query`, both new V1 envelope layer kinds (axis descents with recomputed attestations, sum-budget windows with the verifier's fold replay), the GROVE_V4 gate table, and the relationship to the kept-first-class specialized surfaces. The Query System chapter gains a short pointer section.

`mdbook build` passes. Translations intentionally not touched (they lag by design; adding a chapter does not require translating it).

### DIFF
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -10,6 +10,7 @@
 - [The Storage Layer](storage-layer.md)
 - [The Proof System](proof-system.md)
 - [The Query System](query-system.md)
+- [The Unified PathQuery](unified-path-query.md)
 - [Aggregate Sum Queries](aggregate-sum-queries.md)
 - [Aggregate Count Queries](aggregate-count-queries.md)
 - [Aggregate Sum on Range Queries](aggregate-sum-on-range-queries.md)
--- a/docs/book/src/query-system.md
+++ b/docs/book/src/query-system.md
@@ -184,4 +184,16 @@ algorithm finds common path prefixes and combines query items:
     Merged:  path=["users"], query=items=[Key("alice"), Key("bob")]
 ```
 
+
+## Beyond key selection
+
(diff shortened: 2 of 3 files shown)


QuantumExplorer:
This is Claude. Rebased onto `develop` (the branch was stacked on the pre-squash #801 head) and brought the chapter up to date with everything that merged since it was written:

- **`TopK` → `RankedPage`** with the directional top-k/bottom-k reading.
- **Tag 3 is `AggregateOverValueRange { lo, hi, fold }`** — the chapter now carries the fold table (counts `[3,1,5]`, band `[2,10]`: Population 2, Total 8), the entries-not-distinct-values note, and the fact that all four (axis, fold) cells are served since #809.
- **All secondaries are dual-aggregate PCPS**, documented as a *security requirement*, not a convenience (per the #809 audit: the single-aggregate node hashes share a preimage layout).
- **Query-as-input, fold edition**: the embedded descent payload carries no fold — dual-aggregate commitments mean one proof serves both questions, each verified against its own hash-bound number; the standalone envelopes echo and authenticate the fold instead. The chapter explains why the two wire paths differ here.
- `read_mode` shown as `Option<Box<ReadMode>>`, constructor list updated (`new_axis_aggregate_over_value_range`, `bottom_k`).

`mdbook build` passes locally; the sum-budget and classify sections needed no changes (`match_limit` was already correct).

PR #805, opened 2026-08-14
feat(query): route carrier sum and count+sum reads through run_path_query

QuantumExplorer:
Follow-up to #803 (merged) and #798 (merged) — the dispatch wiring both PRs left open.

## What this closes

`run_path_query`'s `AggregateCarrier` arm served only the count axis and refused the other two by name:

```rust
AggregateKind::Sum | AggregateKind::CountAndSum => Err(Error::NotSupported(
    "carrier aggregate-sum reads have no trusted per-key read primitive; use \
     prove_query with verify_aggregate_sum_query_per_key / ..."
))
```

That was accurate when #798 landed — the readers didn't exist. #803 added them, so the arm now routes all three axes:

| `AggregateKind` | reader |
|---|---|
| `Count` | `query_aggregate_count_per_key` |
| `Sum` | `query_aggregate_sum_per_key` *(newly reachable)* |
| `CountAndSum` | `query_aggregate_count_and_sum_per_key` *(newly reachable)* |

**Every shape `classify()` can produce now has a reader behind it.** The only `NotSupported` the dispatch still raises is the read-mode version gate, which is intentional and unrelated. Module doc updated to say so.

## The result-type decision

`PathQueryRun` gains two variants rather than collapsing the carrier family:

```rust
AggregateCountPerKey(Vec<(Vec<u8>, u64)>),             // existed
AggregateSumPerKey(Vec<(Vec<u8>, i64)>),               // new
AggregateCountAndSumPerKey(Vec<(Vec<u8>, u64, i64)>),  // new
```

The plan for this follow-up was to mirror a `VerifiedPathQuery::AggregatePerKey { per_key: Vec<(Vec<u8>, Option<u64>, Option<i64>)> }`. **That type does not exist anywhere in the repo** — #798 shipped the read dispatch only, and its own module doc notes that \"the unified proof dispatch arrives separately.\" With no mirror target, three typed variants win on every axis I can find:

- **Consistent with this enum's own leaf family** one screen above: `AggregateCount(u64)` / `AggregateSum(i64)` / `AggregateCountAndSum { count, sum }`. Collapsing only the carrier half would make the two halves disagree.
- **Honest types.** A sum carrier can never produce a count, so `Option<u64>` would be statically always-`None` on that axis — the collapsed shape encodes a value that can't occur.
- **Additive.** `AggregateCountPerKey` is already public on develop; collapsing it is a breaking change to a just-shipped variant.
- **1:1 with the readers**, so the routing is a straight hand-off with no reshaping.

When the unified *verify* dispatch lands, it should mirror these three variants rather than the collapsed shape.

## Tests

`aggregate_carrier_count_matches_per_key_reader_and_others_are_refused` existed to assert the refusal, so it is now `aggregate_carrier_all_kinds_match_their_per_key_readers`: a differential assertion per axis (unified answer == dedicated reader) over `ProvableSumTree` and PCPS carrier fixtures, plus a check of the values themselves so the test would catch both readers being wrong in the same way.

New: `aggregate_carrier_per_key_dispatch_surfaces_reader_errors`. Routing must not paper over a reader's rejection, so a non-tree outer match and a single-axis host under a combined carrier each have to fail through `run_path_query` with the same error text the dedicated reader produces.

## Validation

- `cargo clippy --workspace --all-features -- -D warnings` — clean (exit 0, zero diagnostics)
- `cargo test -p grovedb -p grovedb-query --all-features` — 2633 + 255 pass, 0 fail
- `cargo build --no-default-features --features verify -p grovedb` — clean
- `cargo fmt --check` — clean

No version gating: purely additive routing over readers that are already gated on their own existing slots (`query_aggregate_{sum,count_and_sum}_on_range`), so no new slots and no V4 gate.

### DIFF
--- a/grovedb/src/operations/get/run_path_query.rs
+++ b/grovedb/src/operations/get/run_path_query.rs
@@ -4,10 +4,15 @@
 //! [`GroveDb::run_path_query`] classifies the query once
 //! ([`PathQuery::classify`]) and routes it to the engine that already
 //! serves that shape — the key-selection reader, the aggregate-on-range
-//! readers, the indexed-axis primitives, or the budgeted sum reader.
-//! It returns a [`PathQueryRun`] whose variant mirrors the shape, so a
-//! caller holding an arbitrary `PathQuery` gets a typed answer without
-//! knowing in advance which of the specialized entry points serves it.
+//! readers (leaf and per-key carrier, on all three axes), the
+//! indexed-axis primitives, or the budgeted sum reader. It returns a
+//! [`PathQueryRun`] whose variant mirrors the shape, so a caller holding
+//! an arbitrary `PathQuery` gets a typed answer without knowing in
+//! advance which of the specialized entry points serves it.
+//!
+//! Every classified shape now has a reader behind it: the dispatch
+//! returns no `NotSupported` of its own except the read-mode version
+//! gate below.
 //!
 //! Read-mode shapes (axis and sum-budget reads) are gated on
 //! `path_query_methods.unified_read_mode` — `0` before GROVE_V4 means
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Reviewed

Issue #806, opened 2026-08-14
Count-axis secondary cannot answer 'total of counts in a band' — decide before V4 activation


QuantumExplorer:
This is Claude. Filed as an agreed follow-up from the `RangeAggregate` → `AggregateOverValueRange` naming discussion on #799.

## The gap

`AxisTraversal::RangeAggregate` (soon `AggregateOverValueRange`) answers, per axis, *that axis's own aggregate over the entries whose value falls in `[lo, hi]`*:

- **Sum axis** — the TOTAL of the selected entries' sums. Over sums `[40, -10, 25]`, band `[0, 100]` answers `65`. ✔ what people expect.
- **Count axis** — HOW MANY entries were selected, each contributing 1. Over counts `[3, 1, 5]`, band `[2, 10]` answers `2`, **not `8`**.

The reason: the count-axis secondary is a `ProvableCountTree` (`axis_secondary_tree_type` in `grovedb/src/operations/indexed_tree.rs`), which aggregates by counting nodes — it holds no sum of the count values. So "total of the counts in a band" (e.g. *total messages across all chats with 2–10 messages*) has **no O(log n) answer today**. The documented workaround is `Bounded` + caller-side summing, which is `O(log n + k)` and unprovable as a single scalar.

## The fix, if wanted

Upgrade the count-axis secondary to `ProvableCountProvableSumTree`, storing each entry's `count_value` as its sum item — exactly the shape the sum and avg axes already use:

```rust
// grovedb/src/operations/indexed_tree.rs
IndexAxis::Count => TreeType::ProvableCountTree,          // today
IndexAxis::Count => TreeType::ProvableCountProvableSumTree, // proposed
```

Then the existing sum-aggregate-on-range machinery answers the total in `O(log n)`, provable through both the standalone envelope and the V1-embedded axis descent, and the current population semantics stay available via the count half.

Touches: secondary insert/update paths, `verify_grovedb` indexed checks, average/worst-case cost estimates, and a new traversal (or an axis-side flag) to select *total* vs *population* on the count axis.

## Why decide now

This changes the secondary's tree type and therefore the hash-bound wire shape of every PCIT/PCPSIT count secondary. **All indexed-tree types are GROVE_V4-gated and V4 has not activated** — the change is free today and impossible after activation without a new tree-type migration.

If the answer is "not needed", closing this issue is itself the decision; the per-axis semantics are already documented at the traversal, the constructor, and the verified result (#799, commit 817148ad).

PR #808, opened 2026-08-14
feat: explicit AggregateFold on the value-range aggregate (issue #806, part 1)

QuantumExplorer:
Part 1 of #806, in two commits.

## Commit 1 — the rename (mechanical)

`range_aggregate` → `aggregate_over_value_range`, `RangeAggregate` → `AggregateOverValueRange`, across grovedb-query, grovedb, and the book. Wire tag 3 unchanged. Grep gate holds: zero old-name hits.

## Commit 2 — the explicit fold

Tag 3 becomes `AggregateOverValueRange { lo, hi, fold }` with **`AggregateFold { Population = 0, Total = 1 }`** as a frozen wire byte. The caller now says which scalar they mean, making the count-axis misread *unexpressible*: over counts `[3, 1, 5]`, band `[2, 10]` — `Population` answers 2, `Total` answers 8. Everything is V4-gated and pre-activation, so the tag-3 payload change is free.

The 2×2 (axis × fold) matrix:

| | `Population` | `Total` |
|---|---|---|
| **count** | ✅ existing behavior | 🔒 typed `NotSupported` naming #806, until part 2 makes the count secondary sum-bearing |
| **sum** | ✅ **enabled here** — the sum secondary is already PCPS, so its count aggregate answers "how many entries fall in this sum band" with zero merk changes | ✅ existing behavior |

**Structural anti-drift:** the byte range follows the *axis*, the walker follows the *fold*, and one shared `build_aggregate_secondary_proof` now reuses the verifier's own `count/sum_aggregate_inner_range` reconstructors — collapsing four builders and three duplicated clamp/degenerate/out-of-domain sites into code that cannot disagree across the prover/verifier boundary.

**Fold authentication differs by wire path, deliberately:**
- The **standalone envelope** echoes the fold and the verifier authenticates the echo (`fold mismatch` = `CorruptedData`) — pinned by a forgery test.
- The **embedded (V1) payload** carries no fold **on purpose**: PCPS secondaries commit dual `(count, sum)` aggregates in every node hash, so one proof stream serves both questions, and the query — which the verifier holds independently — is the sole source of the fold. `the_fold_lives_in_the_query_not_the_embedded_proof` pins the actual security property: cross-feeding a proof built for one fold to a query asking the other yields *that question's own correct answer* under the genuine root. (An earlier draft of that test asserted rejection; running it showed the dual commitments make rejection impossible *and unnecessary* — the pinned property is the stronger one.)

Also: `indexed_sum_population_over_value_range` trusted reader (version gate ahead of the degenerate-range fast path, per the #801 review finding, with gate tests for both orderings); `AxisAggregateValue::{Count,Sum}` → `{Population(u64), Total(i64)}`; both folds rejected on the Avg axis at `validate()`; unknown fold bytes fail closed at decode.

Part 2 (next PR): flip the count secondary to PCPS and turn the 🔒 cell on. Closes nothing yet — #806 closes with part 2.

## Verification

`cargo test --workspace --all-features` green, `cargo clippy --workspace --all-features -- -D warnings` clean, verify-only build clean, golden encodings re-pinned, fold round trips + cross-fold + count+Total refusal pins in `axis_descent_proof_tests`.

## Summary by CodeRabbit

* **New Features**
  * Added population and total aggregation modes for indexed value-range queries.
  * Added population counting over sum-index value ranges.
  * Aggregate proofs now authenticate and validate the requested aggregation mode.

* **Changes**
  * Renamed indexed range-aggregation APIs to clarify that ranges apply to values.
  * Count results are now labeled as population, while sum results are labeled as total.

* **Bug Fixes**
  * Unsupported combinations, including count totals and average-axis aggregation, are now rejected consistently.

### DIFF
--- a/grovedb-query/src/axis_query.rs
+++ b/grovedb-query/src/axis_query.rs
@@ -247,6 +247,14 @@ pub enum AxisTraversal {
     /// selects `40` and `25`: `Population` answers `2`, `Total` answers
     /// `65`.
     ///
+    /// **Currently unsupported: `Total` on the count axis.** The query
+    /// validates and serializes (the vocabulary is stable), but every
+    /// execution surface — trusted read, embedded prover/verifier,
+    /// standalone prover/verifier — refuses it with a typed
+    /// `NotSupported` naming issue #806 until the count secondary
+    /// becomes sum-bearing (#806 part 2, which removes this paragraph).
+    /// The other three (axis, fold) cells are served today.
+    ///
     /// **Cost** — `O(log n)` in every case, either fold. The walk
     /// classifies each subtree as fully Contained, Disjoint, or Partial
     /// and folds a Contained subtree's stored aggregate in one step,
--- a/grovedb-query/src/axis_query.rs
+++ b/grovedb-query/src/axis_query.rs
@@ -103,8 +103,11 @@ pub const MAX_RANK_OF_KEY_LEN: usize = 255;
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum AggregateFold {
-    /// How many entries the band selects; each contributes 1. Answered
-    /// by the secondary's count aggregate, so it needs a count-bearing
+    /// How many entries the band selects; each contributes 1.
+    /// Entries, not distinct values: two entries sharing the same axis
+    /// value are two nodes of the secondary (it is keyed
+    /// `sort_key ‖ original_key`) and count as 2. Answered by the
+    /// secondary's count aggregate, so it needs a count-bearing
     /// secondary.
     Population,
     /// The sum of the selected entries' axis values. Answered by the
--- a/grovedb-query/src/axis_query.rs
+++ b/grovedb-query/src/axis_query.rs
@@ -92,7 +92,7 @@ pub const MAX_RANK_OF_KEY_LEN: usize = 255;
 
 /// How an [`AxisQuery`] walks the secondary. Wire tags are explicit and
 /// frozen: `RankedPage = 0`, `Bounded = 1`, `RankOfKey = 2`,
-/// `RangeAggregate = 3`.
+/// `AggregateOverValueRange = 3`.
 ///
 /// # Cost
 ///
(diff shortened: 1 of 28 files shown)


grovedb-query/src/axis_query.rs
```diff
@@ -177,44 +229,42 @@ pub enum AxisTraversal {
...
     ///
-    /// **Cost** — `O(log n)` in every case. The walk classifies each
-    /// subtree as fully Contained, Disjoint, or Partial and folds a
-    /// Contained subtree's stored aggregate in one step, descending
-    /// only along the two range boundaries. **No term in the number of
-    /// matched entries** — aggregating a million in-range entries costs
-    /// what aggregating one costs, which is what makes this preferable
-    /// to [`Self::Bounded`] whenever only the total is wanted.
-    RangeAggregate {
+    /// **Cost** — `O(log n)` in every case, either fold. The walk
```
QuantumExplorer, line 261: **[P2] Do not advertise the unsupported count-total cell**

These public docs say both folds work on both Count and Sum, explicitly claiming Count+Total returns 8 and that every combination is O(log n). However, the trusted reader and both proof paths return `NotSupported` for Count+Total until #806 part 2. Because validation still accepts this query, callers can construct and serialize something guaranteed to fail. Please document Count+Total as currently unsupported, or reject it during validation until part 2 lands.

QuantumExplorer (commented):
One actionable finding: the public rustdoc currently advertises a Count×Total matrix cell that the implementation deliberately rejects.

grovedb-query/src/axis_query.rs
QuantumExplorer: Agreed — the docs were writing checks the code refuses to cash. Fixed in `738a5c19` at all three places a caller meets the claim: the variant doc now carries an explicit **"Currently unsupported: `Total` on the count axis"** paragraph (typed `NotSupported` naming #806 at every execution surface; the other three cells served today), and the constructor doc points at it.

On the validation question — I kept the query **valid and serializable**, deliberately, and here is the reasoning for your review: part 2 is stacked next and turns the cell on by flipping the count secondary, with **no grammar change** — the vocabulary is identical across parts 1 and 2, and part 2's diff deletes one doc paragraph and one refusal arm per surface. A temporary `validate()` rejection would put a rule into the shared classify path (the consensus-relevant surface the reader, prover, and verifier all share) only to remove it days later, and would also make the refusal *untestable at the verifier arms* — the new tests reach the descent verifier's refusal via a count+Population proof against a count+Total query, which requires the query to classify. If part 2 stalls for any reason, promoting the refusal into `validate()` is a two-line follow-up.

Also hardened the refusal itself while here: it is now tested AT each verifier arm, not just upstream — including a relabeled-fold-echo envelope that passes the echo check and must be stopped by the inner dispatch.

grovedb/src/operations/proof/verify_path_query.rs
```diff
@@ -106,7 +106,7 @@ pub enum VerifiedPathQuery {
         /// The attested rank.
         rank: u64,
     },
-    /// `RangeAggregate`: the attested aggregate over the entries the
+    /// `AggregateOverValueRange`: the attested aggregate over the entries the
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in `738a5c19` with your proposed direction — the `value` doc now reads per-FOLD (`Population` = non-negative entry count, `Total` = signed sum of the selected values) and closes with "the query carries the fold." You were right that the old text was exactly the confusion this PR exists to remove; it was a leftover from the pre-fold docs I wrote two days ago.

PR #809, opened 2026-08-14
feat: dual-aggregate count secondary — count+Total served (closes #806)

### LINKED ISSUE #806: Count-axis secondary cannot answer 'total of counts in a band' — decide before V4 activation

QuantumExplorer:
Part 2 of #806 (part 1 was #808). Closes #806.

## The flip

`axis_secondary_tree_type(Count)` → `ProvableCountProvableSumTree`, with each entry's `count_value` mirrored into the sum half. **"Total of the counts in a band" is now one committed scalar, O(log n), provable** through both the embedded V1 descent and the standalone envelope. Every axis's secondary now has the same dual-aggregate shape, and the (axis × fold) matrix is complete:

| | `Population` | `Total` |
|---|---|---|
| **count** | ✅ | ✅ **turned on here** |
| **sum** | ✅ | ✅ |

The issue's own example, verbatim, is the headline test: counts `[3, 1, 5]`, band `[2, 10]` → `Population` 2, `Total` **8** — read three ways (trusted, embedded proof, standalone envelope) and re-checked after a delete.

## Integrity mechanics

- **Four lockstep payload sites** (batch mirror row, mirror closure, reconcile's desired map, `verify_grovedb`'s expected payload) all go through one fail-closed conversion: `count > i64::MAX` is a typed error, **never a clamp** — a clamped total would lie through authenticated state. Unreachable for real trees; unit-tested at the boundary.
- **The mirror closure's signature widens to `|count, sum|`**: the count-axis payload is now a function of the count, which the old `|sum|`-only fast-path equality check could not see — left alone it would have silently skipped real payload updates.
- `verify_grovedb` holds across count-changing mutations and deletes (its expected-payload check doubles as the drift detector); a tampered sum inside the carried secondary proof breaks the `combine_hash_three` binding.
- TopK / RankOfKey / Bounded proofs survive with **zero merk changes** — the count-offset verifier was already dual-flavor (`HashWithCount` / `HashWithCountAndSum`).
- Cost estimator sizes the count row as `SumItem(i64::MAX)`; the count axis now sizes exactly like the sum axis, which the mirror-scaling cost test pins (the old count-vs-avg deltas collapse to the 1-byte payload-shape difference).

Wire-affecting but free: indexed trees are GROVE_V4-gated and V4 has not activated — no migration, no new version slots. The "currently unsupported" doc paragraphs from #808 are deleted, per their own text.

## Verification

`cargo test --workspace --all-features` green, clippy `-D warnings` clean, verify-only build clean. Local llvm-cov patch ~94% (remaining lines are `?`-region artifacts in closures plus the overflow arm, which the unit test covers). New tests: the 2×2 matrix differential across all three read paths, count+Total round trip incl. post-delete, `verify_grovedb` across mutations, secondary-sum tamper rejection, overflow fail-closed, gate + inverted-bounds gate for the new reader.

A blockchain-security-auditor pass over this diff is running and its findings will be posted before merge.

### DIFF
--- a/grovedb-query/src/axis_query.rs
+++ b/grovedb-query/src/axis_query.rs
@@ -249,14 +249,7 @@ pub enum AxisTraversal {
     /// The same pair over sums `[40, -10, 25]` with band `[0, 100]`
     /// selects `40` and `25`: `Population` answers `2`, `Total` answers
     /// `65`.
-    ///
-    /// **Currently unsupported: `Total` on the count axis.** The query
-    /// validates and serializes (the vocabulary is stable), but every
-    /// execution surface — trusted read, embedded prover/verifier,
-    /// standalone prover/verifier — refuses it with a typed
-    /// `NotSupported` naming issue #806 until the count secondary
-    /// becomes sum-bearing (#806 part 2, which removes this paragraph).
-    /// The other three (axis, fold) cells are served today.
+
     ///
     /// **Cost** — `O(log n)` in every case, either fold. The walk
     /// classifies each subtree as fully Contained, Disjoint, or Partial
--- a/grovedb/src/batch/indexed_tree/mirror.rs
+++ b/grovedb/src/batch/indexed_tree/mirror.rs
@@ -23,17 +23,25 @@ use grovedb_storage::StorageContext;
 use grovedb_version::version::GroveVersion;
 
 use super::{read_entry_aggregates, AggregatePair, AggregateTransition};
-use crate::{operations::indexed_tree::make_axis_secondary_key, Element, Error};
+use crate::{
+    operations::indexed_tree::{count_value_as_sum, make_axis_secondary_key},
+    Element, Error,
+};
 
 /// The per-axis payload stored alongside the sort key. The key encodes the
 /// ordering value; the payload carries what the secondary's own aggregate
-/// must sum to, which is why the sum and avg axes cannot store a bare item.
-fn axis_row_payload(axis: IndexAxis, sum: i64) -> Element {
-    match axis {
-        IndexAxis::Count => Element::new_item(Vec::new()),
+/// must sum to, which is why NO axis can store a bare item: every
+/// secondary is a dual-aggregate `ProvableCountProvableSumTree`, and the
+/// count axis mirrors its `count_value` into the sum half so a band
+/// TOTAL is one committed scalar (issue #806).
+///
+/// Fallible only through [`count_value_as_sum`]'s fail-closed guard.
+fn axis_row_payload(axis: IndexAxis, count: u64, sum: i64) -> Result<Element, Error> {
+    Ok(match axis {
+        IndexAxis::Count => Element::new_sum_item(count_value_as_sum(count)?),
         IndexAxis::Sum => Element::new_sum_item(sum),
         IndexAxis::Avg => Element::new_item_with_sum_item(Vec::new(), sum),
-    }
+    })
 }
 
 /// Enforce the precise per-axis item-key bound: avg prepends a 16-byte sort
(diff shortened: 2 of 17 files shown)


QuantumExplorer:
This is Claude. The blockchain-security-auditor pass over this diff completed (after merge — findings below are follow-ups, none block correctness of what landed).

**All five probed surfaces hold**: payload-site byte-identity (it found a fifth site — the cost estimator row — beyond my four), fold confusion (both wire paths authenticated or query-sourced; range-reconstructor sharing is total with no residual duplicated shape), overflow (prover/verifier symmetric i128-accumulate + final i64 gate; band-sum overflow already covered by `docs/audit-non-issues.md`), sort-key-vs-payload divergence (`verify_grovedb` compares payload BYTES at lib.rs:1772, and reconcile repairs stale payloads, not just missing keys), and cost sizing (`SumItem(i64::MAX)` = 11 bytes = exactly `SUM_ITEM_COST_SIZE`; estimate == actual).

**Three findings, none critical:**

- **A (Low, maintainability-with-teeth)**: the payload is hand-copied at five sites and three surrounding comments still describe the OLD empty-`Item` shape — including the doc on `mirror_indexed_axis_to_secondary` itself. A contributor who trusts the comment and "restores" the old payload at any one site gets either a false-positive corruption alarm on healthy state or two entry points committing different root hashes for identical writes. Fix: promote one shared `axis_row_payload(axis, count, sum)` called from all five sites + a (count × sum × axis) grid test, and delete the stale comments. Follow-up PR incoming.
- **B (Low, operational)**: the shape flip is ungated — any PCIT/PCPSIT data written before this commit fails `verify_grovedb` until rebuilt or reconciled. Consistent with the #791 precedent and fine iff no released grove version ever allowed indexed-tree elements — **worth positive confirmation from whoever owns the platform pin** rather than assumption.
- **C (Informational, the invariant to protect)**: `node_hash_with_count` and `node_hash_with_sum` share a preimage layout (no domain tag), so a single-aggregate secondary's count proof could be node-type-rewritten into a "sum" proof with an identical root. Closed today precisely BECAUSE every axis secondary is dual-aggregate (112-byte preimage) — meaning dual-aggregate is a **security requirement, not a convenience**. If anyone later "optimizes" an axis back to a single-aggregate tree, a Total becomes forgeable from a Population proof. The follow-up PR adds a comment at `axis_secondary_tree_type` stating this.

Also verified clean: Population can't be skewed by the payload change; nothing consumes the count secondary's sum aggregate outside the fold; the overflow guard is consistent and unreachable; envelope trailing-byte rejection still applies.

Issue #822, opened 2026-08-22
Append-only tree family charges write churn as new storage: compaction blob, epoch≥2 buffer writes and frontier rewrites are all reported as added_bytes


QuantumExplorer:
## Summary

For `CommitmentTree` / `BulkAppendTree` (and by construction `PrivateDocumentStore`), every data-storage write is reported to the fee layer as **`added_bytes`**, and nothing is ever reported as replaced or removed. Three writes that are physically *replacement churn* are therefore billed as permanent new storage:

1. **Epoch compaction** writes the chunk blob (2048 × ~328 B at `chunk_power 11`) as ~630 KB of `added_bytes` — but the blob *supersedes* the dense-buffer entries it was built from. The buffer's 2-byte position keys are simply reused by the next epoch (`DenseFixedSizedMerkleTree::reset()` only zeroes `count` and the cache), so net growth at the boundary is the blob minus the buffer it replaces ≈ a few KB, not 630 KB.
2. **Dense-buffer writes from epoch 2 onward** overwrite last epoch's stale value at the same position key, yet are charged as new bytes every epoch. Each note's bytes are effectively charged twice over its life (buffer write + blob copy).
3. **The frontier** (`__ct_data__`) is one value rewritten in place on every append, sized `~42 + 32 × popcount(position)` (74 B → 1,066 B); each rewrite is charged in full as `added_bytes`.

All three come from the same mechanism: the data puts are issued with `cost_info: None` — `DenseFixedSizedMerkleTree::put_value` (`tree.rs`), the MMR `storage_adapter.rs:98` put that persists chunk blobs and MMR nodes, and `commitment_tree/mod.rs:563` for the frontier — so the commit path, which has no previous-size information for them, charges key+value as added. The previous size *is* knowable at each site: the frontier and buffer writes are preceded by reads of the same key, and the blob's predecessor bytes are exactly the buffer being compacted.

## Measured (Dash Platform, real applies, 1-action ShieldedTransfer, 216-byte note, from an empty pool)

| write | charged bytes | physical reality |
|---|---|---|
| nullifier insert (Merk) | 214 B | genuinely new (32 B key + node framing) |
| note append | 429 → 745 B within an epoch (312 B entry + ~50 B put overhead + **frontier rewrite**) | entry new once; frontier is in-place replacement |
| **compaction (append #2048)** | **630,166 B** added, `NoStorageRemoval` | blob replaces the ~640 KB buffer it was built from |
[...]

PR #828, opened 2026-08-22
feat(dense): O(height) dense-buffer root maintenance via per-position hash records (GROVE_V4)

QuantumExplorer:
## Summary

The dense fixed-sized Merkle tree — the buffer of `BulkAppendTree`, and through it `CommitmentTree` (the shielded pool, live on mainnet) and `PrivateDocumentStore`, as well as the standalone `DenseAppendOnlyFixedSizeTree` — kept no intermediate hashes: **every insert re-derived the root by walking every filled position out of storage** (one read and two blake3 calls per position). On the live commitment-tree path that walk, not compaction, was the dominant per-append cost: the k-th note of an epoch cost O(k), ≈ 2k reads and ≈ 4k blake3 calls on the last insert of every `chunk_power = 11` epoch, O(C²) per epoch.

This PR adds a GROVE_V4 gate, `dense_tree_versions.root_maintenance`, under which the tree keeps a **per-position hash record** beside each value and an insert updates only its ancestor path: `2 + depth` blake3 calls, ≤ `2·depth` record reads, `depth + 1` record writes — O(height) whatever the fill. The root is the record at position 0. **Root hashes are identical under both versions**; only the work (and so the fee) and the records written beside the values change.

### Design

- Record key `b'h' || position` (3 bytes — collides with nothing: slots are 2-byte, MMR keys 4/8-byte); value `generation (u64 BE) || value_hash (32) || node_hash (32)`.
- `generation` is the epoch tag (the bulk tree's chunk count, advanced by `reset`), so a record left by an earlier epoch over the same slot keys is never trusted.
- A record that is absent (a buffer filled under V1..V3) or stale is recomputed from the values — the version-0 walk over that subtree — and recorded: a one-time catch-up that costs at most one V0 walk per buffer. Every V4 insert path (`insert`, `try_insert*`, including the `_no_root` variants) maintains records, so no current-generation record can ever be stale.
- Record writes are sized from the read `resolve_record` performs anyway (rewrite → `for_in_place_value_rewrite(72, 72)`; absent key → new storage); the new leaf's record is read once only when the owner reports the slot as `SlotWriteAccounting::Overwrite`.
- Billing conventions are unchanged: the `Result`-returning bulk / commitment appends bill the dense tree's hash count (now `2 + depth` instead of `2 · count`) plus `storage_accounting_cost` and drop the record reads, as they always dropped the walk's reads; record writes reach every caller at commit; the `CostResult`-returning `append_deferred_roots` (PrivateDocumentStore) bills everything.
- Versioned dispatch follows the repo pattern: `grovedb-dense-fixed-sized-merkle-tree/src/tree/root_maintenance/{mod,v0,v1}.rs`; v0 is byte-for-byte the shipped behaviour. `insert`/`try_insert*`/`root_hash` and `compute_current_state_root` on the bulk / commitment / store trees now take `grove_version`.

### Estimators (V4 is unreleased, so its models are edited in place)

- Shared `dense_record_maintenance_bound(height, catch_up)` in `batch/estimated_costs/mod.rs`.
- **CommitmentTreeInsert** (`commitment_tree_insert_op_cost`): **keeps** the `2·(epoch − 1)` hash bound — a buffer filled under GROVE_V3 pays one full walk at its first V4 append and that insert is admitted under this bound (tested by seeding under `GROVE_V3` and appending under V4, at chunk_power 4 and 11); adds the records' seeks and storage. Tightening this needs a Platform-coordinated backfill of live buffers (or one epoch after activation) and is left as a follow-up.
- **PrivateDocumentStoreInsert**: new shared `private_document_store_insert_op_cost` used by both arms (average: declared `chunk_power`; worst: the physical ceiling). The store is V4-born, so the record model is the bound; the compaction read-back, the blob's MMR-peak read-back and the element load are included. This replaces the old average-case arm, which was an amortized figure rather than a bound (it did not dominate the compacting append).
- **BulkAppend**: record terms; the average arm gains the preprocessing element-load bytes (a pre-existing under-bound exposed by the new position sweep); the worst-case hash bound fixed to the physical ceiling.
- **DenseTreeInsert**: record terms added, legacy hash figures kept.

### Tests

- `grovedb-dense-fixed-sized-merkle-tree/src/root_maintenance_tests.rs`: v1 roots == v0 roots at every fill level for heights 1..=6, across sessions (cold caches) and within one session; `try_insert_no_root` maintains records; a V0-filled buffer is caught up by the first V1 insert (≤ one V0 walk) and is O(height) afterwards; records from an earlier epoch / unparsable records are never trusted; per-insert work bounded by depth (hashes exactly `2 + d`, ≤ `2d` record reads, `d + 1` writes, no value read-back); session-independent costs; rewrite vs new sizing; root read = one record read; rollback on a record-write fault; unknown version rejected; record encoding round-trip.
- `grovedb/src/tests/append_family_cost_bound_tests.rs`: `estimated >= actual` in every dimension for the V3→V4 commitment-tree catch-up (cp 4 and cp 11, roots equal to a V4-from-birth tree), PrivateDocumentStore sweeps (cp 4 positions + cp 11 epoch boundary), BulkAppend sweep, DenseTreeInsert every position; catch-up paid once.
- Bulk crate: `buffered_append_hash_count_follows_the_root_maintenance_version` (V3 `2·count`, V4 `2 + depth`, equal roots both ways); storage identity across versions apart from the records.
- PDS / bulk / CT crate pins updated to the V4 figures; #822 accounting-delta tests extended with the leaf-record term.

### Docs

Book chapters 14 (bulk), 15 (commitment), 16 (dense) and `docs/crates/costs.md` updated; `v4.rs` documents the gate.

## Summary by CodeRabbit

* **New Features**
  * Grove V4 now maintains dense-tree roots incrementally using persisted hash records.
  * Added version-aware root maintenance, including legacy compatibility and automatic record catch-up.
  * Added public support for hash records and dense-tree maintenance configuration.

* **Bug Fixes**
  * Improved rollback and handling of stale or malformed hash records.
  * Updated state-root and proof calculations to consistently respect the active Grove version.

* **Documentation**
  * Updated tree behavior, storage layout, and operation-cost documentation for Grove V4.

* **Tests**
  * Added comprehensive cross-version, storage-accounting, failure-recovery, and cost-bound coverage.


grovedb/src/lib.rs
```diff
@@ -2987,7 +2991,7 @@ impl GroveDb {
                     .unwrap();
                 use grovedb_dense_fixed_sized_merkle_tree::DenseFixedSizedMerkleTree;
                 match DenseFixedSizedMerkleTree::from_state(*height, *count, storage_ctx) {
-                    Ok(t) => match t.root_hash().unwrap() {
+                    Ok(t) => match t.root_hash(grove_version).unwrap() {
```
QuantumExplorer: [P1] Recompute integrity roots from payload values, not hash records

This is the `verify_grovedb` integrity path, but under V4 `root_hash` returns the position-0 record without reading any dense payload values. I inserted four V4 values, overwrote position 3 on disk with a same-length value while leaving the records intact, and `verify_grovedb(..., GROVE_V4)` returned an empty issue set. The same blind spot reaches the BulkAppendTree, CommitmentTree, and PDS branches above because they use this V4 dense-root path too. Please give integrity/state-validation callers an explicit value-walking root API (or otherwise validate records against values) and add a tamper regression; the record fast path is fine for normal root reads, but it cannot be the independent audit.

grovedb/src/batch/estimated_costs/average_case_costs.rs
```diff
@@ -430,24 +429,33 @@ impl GroveOp {
...
-                        replaced_bytes: 0,
+                        added_bytes: value_size.saturating_add(records.added_bytes),
+                        replaced_bytes: records.replaced_bytes,
                         removed_bytes: StorageRemovedBytes::NoStorageRemoval,
                     },
-                    storage_loaded_bytes: (value_size as u64) * (AVG_COUNT as u64),
-                    hash_node_calls: AVG_HASH_CALLS,
+                    storage_loaded_bytes: (value_size as u64) * (AVG_COUNT as u64)
+                        + records.loaded_bytes,
+                    hash_node_calls: AVG_HASH_CALLS.max(records.hash_calls),
```
QuantumExplorer: [P1] Include the legacy-buffer catch-up in standalone dense estimates

`catch_up = true` only expands the record-write count; `records.hash_calls` remains the normal `height + 1` ancestor-path cost. The first V4 insert into a `DenseAppendOnlyFixedSizeTree` populated under V3 can instead walk and hash the entire existing buffer. Both this arm and the worst-case counterpart still cap legacy work at 8/255 nodes. In a height-10 tree seeded with 600 V3 entries, the first V4 insert cost 644 seeks / 1,208 hashes, while the estimates were 76 / 24 (average) and 324 / 773 (worst). This violates the admission invariant in both models. Please bound the migration walk (and its loaded bytes) over every permitted legacy fill, or carry enough state to make the bound tighter, and add the missing V3-to-V4 standalone dense-tree regression.

QuantumExplorer (commented):
I found two blocking V4 correctness issues, both confirmed with focused regression tests: the integrity walk can accept tampered dense-buffer payloads when records remain intact, and the standalone dense-tree admission estimates do not cover the V3-to-V4 record catch-up. The committed test suites and formatting pass, but these two omitted cases violate the integrity and estimated >= actual contracts.

grovedb-dense-fixed-sized-merkle-tree/src/test_utils.rs
```diff
@@ -30,8 +40,24 @@ impl<'db> StorageContext<'db> for MemStorageContext {
...
+    /// tests can compare a cold read against the tree's deterministic
+    /// cache-hit charge.
     fn get<K: AsRef<[u8]>>(&self, key: K) -> CostResult<Option<Vec<u8>>, grovedb_storage::Error> {
-        Ok(self.data.borrow().get(key.as_ref()).cloned()).wrap_with_cost(OperationCost::default())
+        self.gets.borrow_mut().push(key.as_ref().to_vec());
+        if self.fail_record_gets.get() && key.as_ref().len() == 3 {
+            return Err(grovedb_storage::Error::StorageError(
+                "injected hash record read failure".to_string(),
+            ))
+            .wrap_with_cost(OperationCost::default());
```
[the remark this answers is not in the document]
QuantumExplorer: This is Claude. Done in 0e6a2c17 — the injected record-read fault now returns `seek_count: 1` so the attempted lookup stays charged. _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

grovedb/src/batch/estimated_costs/average_case_costs.rs
```diff
@@ -314,26 +317,60 @@ impl GroveOp {
...
                 };
+                // The buffer's hash records (GROVE_V4 root maintenance):
+                // the ancestor-path rewrites, and the catch-up of a buffer
+                // filled under GROVE_V3. At the declared epoch scale, or the
+                // physical ceiling when undeclared. The bulk append bills
+                // their writes (at commit), not their reads.
+                let records = super::dense_record_maintenance_bound(
+                    append_tree_chunk_power.unwrap_or(super::PHYSICAL_MAX_CHUNK_POWER),
+                    true,
+                );
```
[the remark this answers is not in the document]
QuantumExplorer: This is Claude. Not changed: the BulkAppend / DenseTreeInsert / PrivateDocumentStoreInsert arms are only reached under GROVE_V4 (`apply_batch.keyless_op_cost_dispatch: 1` — V1..V3 skip keyless ops entirely, pinned by `test_keyless_append_ops_still_estimate_as_free_before_v4`), and the CommitmentTreeInsert arms are version-dispatched with the record terms only in the v1 model (`*_commitment_tree_insert_v1`; the v0 legacy constants are untouched and pinned by `test_commitment_tree_insert_*_case_cost_pinned_before_v4`). So no V1..V3 estimate gains record terms; a separate `root_maintenance` branch inside V4-only arms would only ever fire for a hand-doctored version, where over-reserving is the safe direction. _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

grovedb/src/batch/estimated_costs/average_case_costs.rs
```diff
@@ -350,75 +387,37 @@ impl GroveOp {
...
-                        .saturating_add(AMORTIZED_MMR_HASHES),
-                    sinsemilla_hash_calls: 0,
-                })
+                // need no separate declaration. An upper bound at the
+                // declared epoch scale, shared with the worst-case arm.
+                item_cost.add_cost(super::private_document_store_insert_op_cost(
+                    entry.len() as u32,
+                    chunk_power,
+                    element_flags_load_bound,
+                ))
```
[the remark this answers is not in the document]
QuantumExplorer: This is Claude. Already the case: both PDS arms delegate to the shared `private_document_store_insert_op_cost`, which adds `NON_COUNTED_WRAPPER_BYTE` (1) to `added_bytes` unconditionally (`batch/estimated_costs/mod.rs`), so the average-case and worst-case bounds both reserve it; the bound sweeps in `append_family_cost_bound_tests` run against stores that are not wrapped and `test_batch_append_preserves_non_counted_wrapper` covers the wrapped apply. _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

grovedb/src/lib.rs
QuantumExplorer: This is Claude. Fixed in dcfea731: the dense tree now exposes an explicit value-walk audit API (`root_hash_from_values`, and `recorded_root` for the record the fast path would trust), surfaced as `compute_current_state_root_from_values` / `buffer_record_mismatch` on BulkAppendTree, CommitmentTree and PrivateDocumentStore. `compute_non_merk_child_hash` (verify_grovedb) and `compute_non_merk_state_root` (state-sync finalize) use the value walk, and verify_grovedb additionally reports a current position-0 record that disagrees with the walked root under a `__dense_hash_records__` sentinel. Regressions in `non_merk_integrity_audit_tests.rs`: your exact scenario (4 V4 values, position 3 overwritten same-length behind intact records) now yields both the child-hash issue and the record issue, for the dense tree, the store and the bulk tree; plus the forged-record case (values intact → only the record issue) and a V3-filled buffer (no records → clean). _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

grovedb/src/batch/estimated_costs/average_case_costs.rs
QuantumExplorer: This is Claude. Fixed in dcfea731: shared `dense_tree_insert_op_cost(value_size, height)` now bounds the catch-up walk — `2^height - 1` value reads (loaded bytes scaled by the value size) and two hashes each, plus the record reads/writes/storage with catch-up — at the tree's declared height (`TreeType::DenseAppendOnlyFixedSizeTree(height)` on its own layer, plumbed through `append_tree_chunk_power` like the other append trees) for the average case and at the physical ceiling for the worst case / undeclared. Regression `test_dense_tree_v4_estimates_cover_the_catch_up_insert`: height-10 tree seeded with 600 V3 entries, the first V4 insert (≥ 1,200 hashes) and the following O(height) inserts are dominated by both estimators, and the declared average stays below the ceiling's walk. _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

QuantumExplorer (commented):
Reviewed

Issue #721, opened 2026-05-20
[audit][medium] grovedbg build script downloads release artifact at build time


QuantumExplorer:
Created from a Codex audit of grovedb. No code changes were made as part of the audit.

## Summary
The `grovedbg` feature build script downloads a GitHub release zip during build and panics on network/download errors.

## Impact / failure scenario
The SHA-256 check helps integrity, but builds with the feature enabled are network-dependent and less reproducible.

## References
- `grovedb/build.rs:15`
- `grovedb/build.rs:24`
- `grovedb/build.rs:29`
- `grovedb/build.rs:33`

## Suggested fix
Vendor the artifact, require an explicit prefetch step, or make the build script use a local override/cache with clear error handling.

## Suggested tests
Test `--features grovedbg` in an offline or network-disabled environment.

QuantumExplorer:
This is Claude, triaging on behalf of @QuantumExplorer.

Closing as **acceptable supply-chain posture**. The code is unchanged at `grovedb/build.rs:1-37` and does what the issue describes, but the two things that would make it dangerous are both already in place:

- **Gated off by default.** `grovedb/Cargo.toml` sets `default = ["full", "estimated_costs"]`; `grovedbg` is a separate opt-in feature. With it off, `build.rs` compiles to `#[cfg(not(feature = "grovedbg"))] fn main() {}` — a literal no-op with no network dependency, and `reqwest`/`sha2` are optional build-deps that aren't even resolved. Enabling it is a deliberate act by a developer who wants the web visualizer.
- **Integrity enforced.** Both the version tag (`v1.2.0`, `build.rs:10`) and `GROVEDBG_SHA256` (`:8-9`) are pinned, and the hash is asserted at `:33`. A substituted or tampered artifact fails the build. The missing HTTP status check at `:23` means a 404 body gets written to disk — but it then fails the SHA assert, i.e. fail-closed.

**Becomes real if** `grovedbg` is added to the default feature set, or if the SHA-256 pin is dropped in favor of an unpinned or floating artifact.

One real (minor) defect the issue doesn't name, worth a separate ticket if anyone has hit it: the `if !grovedbg_zip_path.exists()` guard at `:15` caches a truncated or 404 download in `OUT_DIR`, so every subsequent rebuild fails the SHA assert without re-fetching until the user manually clears `target/`. Two-line fix — verify before persisting, or delete on mismatch.

[...]

Issue #703, opened 2026-05-20
[audit][medium] Public versioned APIs miss explicit version gates


QuantumExplorer:
Created from a Codex audit of grovedb. No code changes were made as part of the audit.

## Summary
Several public methods accept `GroveVersion` but do not appear to check whether the operation is supported for that version.

## Impact / failure scenario
Future unsupported versions could execute writes or queries instead of returning `VersionError`.

## References
- `grovedb/src/lib.rs:498`
- `grovedb/src/lib.rs:517`
- `grovedb/src/lib.rs:919`
- `grovedb/src/operations/mmr_tree.rs:38`
- `grovedb/src/operations/bulk_append_tree.rs:38`
- `grovedb/src/operations/dense_tree.rs:35`
- `grovedb/src/operations/commitment_tree.rs:62`

## Suggested fix
Add explicit version checks to all public `GroveVersion` entry points, especially non-Merk writes.

## Suggested tests
Add unsupported-version tests that expect `VersionError`.

QuantumExplorer:
This is Claude, triaging on behalf of @QuantumExplorer.

Closing as **not reachable as filed**. The finding splits into two parts, and neither supports the stated impact ("could execute writes instead of returning VersionError").

**Three of the seven cited refs are read-only APIs.** The `grovedb/src/lib.rs` line numbers are stale; resolved against the contemporaneous commit `e98bab5f`, they are `pub fn root_key`, `pub fn root_hash`, and `pub fn verify_grovedb` — two accessors and a verification helper, none with version-dependent behavior and none a write path. Adding `check_grovedb_v0!` to them would be gratuitous and would actively break callers the moment an unrelated slot was bumped.

**The four typed non-Merk append entry points do lack a top-level gate, but they already fail closed** — deeper, and before any mutation:

- `mmr_tree_append` → `push_call_site_hashes` → `grovedb-merkle-mountain-range/src/cost/mod.rs:88-98` returns `UnknownVersionMismatch`, *before* `mmr.push`.
- `bulk_append` → `tree.append` → `grovedb-bulk-append-tree/src/cost/mod.rs:182-201`, same.
- `dense_tree_insert` → `tree.insert` → `grovedb-dense-fixed-sized-merkle-tree/src/tree/root_maintenance/mod.rs:56-71`, same.
- `commitment_tree_insert` → `ct.append_raw`/`ct.save` → `grovedb-commitment-tree/src/commitment_tree/cost/mod.rs:43,73`, same.

[...]

Issue #702, opened 2026-05-20
[audit][medium] GroveVersion::default creates protocol version 0


QuantumExplorer:
Created from a Codex audit of grovedb. No code changes were made as part of the audit.

## Summary
`GroveVersion` derives `Default`, producing protocol version `0`, while active versions start at V1.

## Impact / failure scenario
Public callers can pass a default version that satisfies v0 method gates despite not being an active protocol version.

## References
- `grovedb-version/src/version/mod.rs:14`
- `grovedb-version/src/version/mod.rs:35`
- `grovedb-version/src/tests.rs:163`

## Suggested fix
Remove `Default`, implement it as `latest()`/`first()`, or validate that the protocol version exists in `GROVE_VERSIONS` before use.

## Suggested tests
Update the `grove_version_default` test and add public-entrypoint validation tests.

QuantumExplorer:
This is Claude, triaging on behalf of @QuantumExplorer.

Closing as **inert**. Two independent findings, either of which is sufficient:

1. **`protocol_version` is never read by any logic anywhere in the workspace.** A full-repo grep finds it only in the four `v1.rs`/`v2.rs`/`v3.rs`/`v4.rs` const definitions and in test assertions (e.g. `grovedb/src/tests/query_tests.rs:879`, `grovedb-version/src/tests.rs:64`). Nothing dispatches on it — every gate reads a *feature* slot such as `grove_version.grovedb_versions.operations.insert.*`, never `protocol_version`.

2. **`GroveVersion::default()` is behaviorally identical to `GROVE_V1`.** Every `FeatureVersion` in `GROVE_V1` is `0`; the only non-zero values in `grovedb-version/src/version/v1.rs` are `protocol_version: 1` (never read) and `max_aggregate_sum_query_elements_scanned: 1024` (`v1.rs:242`), which is not a version slot and whose hand-written `impl Default` (`grovedb_versions.rs:19-25`) returns the same 1024.

So the worst outcome of calling `GroveVersion::default()` is that the caller gets V1 behavior — indistinguishable from passing `GROVE_V1` explicitly, which is already legal. The stated impact ("satisfies v0 method gates despite not being an active protocol version") is true only vacuously: v0 gates are exactly what an active protocol version selects. No caller in this repo uses it outside one test, and Platform passes `platform_version.drive.grove_version`.

**Becomes real if** `protocol_version` ever gains a reader that dispatches behavior on it.

[...]

Issue #700, opened 2026-05-20
[audit][medium] delete_checkpoint can delete non-checkpoint GroveDB directories


QuantumExplorer:
Created from a Codex audit of grovedb. No code changes were made as part of the audit.

## Summary
`delete_checkpoint` validates a path by opening it as GroveDB/RocksDB, then removes the directory.

## Impact / failure scenario
A caller can delete any closed GroveDB/RocksDB directory, not just paths created by `create_checkpoint`.

## References
- `grovedb/src/checkpoints.rs:39`
- `grovedb/src/checkpoints.rs:43`
- `grovedb/src/checkpoints.rs:46`

## Suggested fix
Write a checkpoint marker/manifest during checkpoint creation, canonicalize paths, and reject normal DB directories.

## Suggested tests
Assert deleting a normal non-checkpoint GroveDB path is rejected.

QuantumExplorer:
This is Claude, triaging on behalf of @QuantumExplorer.

Closing as **not a security issue** (hygiene at most). The code is unchanged at `grovedb/src/checkpoints.rs:28-48`, and the observation is technically accurate — a RocksDB checkpoint directory is structurally identical to a normal RocksDB directory, so `Self::open_checkpoint(path)?` genuinely cannot distinguish them, and the `path.components().count() < 2` guard (`:32-37`) rejects `/` but admits `/tmp`.

But there is no confused deputy here. `delete_checkpoint` is a `pub fn` whose only input is a path the caller chose, and it deletes exactly what it was told to delete. Nothing in grovedb calls it with externally-sourced paths — the only call sites are `grovedb/src/tests/checkpoint_tests.rs:211` and `grovedb/src/tests/misc_coverage_tests.rs:1821,1832`. A live database cannot be destroyed this way either: RocksDB's LOCK makes `open_checkpoint` fail on an open DB, so only *closed* databases are reachable, and only by a caller who typed their path. The existing open-to-validate step is already stricter than a bare `remove_dir_all` and reads as belt-and-braces rather than a security control.

One genuine side effect the issue doesn't mention, for the record: `open_checkpoint` on a real DB directory *mutates* it (WAL replay / MANIFEST churn) before the delete decision is made.

**Becomes real if** grovedb itself ever calls `delete_checkpoint` with a path derived from configuration or network input rather than from an immediate caller.

Rationale being added to `docs/audit-non-issues.md`.

PR #838, opened 2026-08-26
docs(audit): record six verified-rejected findings; correct the #693 invariant

QuantumExplorer:
Records the rejected findings #691, #700, #702, #703, #682, and #721, with their caller and trust-boundary assumptions, so later audits can distinguish duplicate claims from different failure modes.

Corrects the #693 rationale: non-Merk verification can consume proof-supplied metadata before parent authentication, but the bulk/commitment completed-chunk conversion is bounded independently. The validated height is in `1..=16`, so `completed_chunks = total_count / 2^height <= 2^63 - 1` for every `u64 total_count`. This conversion cannot overflow; its safety does not depend on release wrapping.

Clarifies two other contracts:
- Lower-layer cost-version checks precede child writes, but direct typed appends can leave child writes in a caller-supplied transaction when a later parent update fails. The caller must roll back on error; locally owned transactions commit only on success.
- Disabling `grovedbg` leaves its build script inactive and its optional build dependencies unactivated through that feature. Those dependencies still participate in workspace lockfile resolution.

Rejections are scoped to the documented claim, caller, input trust boundary, and applicable version. Independent MMR arithmetic, pagination, and resource use before authentication require separate analysis.

Documentation only. Validation: checked the corrected statements against the PR-head source and Cargo resolver documentation; verified the patch with `git diff --check`. No new runtime test run was needed for this prose-only update.

### DIFF
--- a/docs/audit-non-issues.md
+++ b/docs/audit-non-issues.md
@@ -31,9 +31,19 @@ counts**, both of which are capped far below the wrap point:
 - Value/element sizes are capped orders of magnitude below 4 GiB by platform
   state-transition and document size limits.
 - Tree element counts are accumulated one append at a time and committed into
-  hash-verified `Element` bytes; a verifier reads the count from bytes that
-  are checked against the trusted root **before** any arithmetic runs, so a
-  forged count is rejected as a bad proof, not fed into the math.
+  `Element` bytes that the proof chain binds to the trusted root, so a forged
+  count is ultimately rejected as a bad proof.
+
+  Be precise about *when* that binding happens, though: for the non-Merk
+  lower layers it is **after** the arithmetic, not before. `verify.rs:1953-2016`
+  dispatches the MMR / bulk-append / dense verifiers using an `Element`
+  decoded from the proof's own `value_bytes`, and only at `verify.rs:2018`
+  does `combine_hash(value_hash(value_bytes), &lower_hash)` check it against
+  the parent-committed hash. A forged count therefore *does* reach the math;
+  what makes that safe is that the release build wraps and the chain check
+  then rejects the result. Do not lean on "verified before the arithmetic"
+  as the invariant — and note that a build with `overflow-checks = true`
+  panics instead of wrapping.
 
 Verified-and-rejected instances:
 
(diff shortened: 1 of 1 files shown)


docs/audit-non-issues.md
```diff
@@ -31,9 +31,19 @@ counts**, both of which are capped far below the wrap point:
...
+  `Element` bytes that the proof chain binds to the trusted root, so a forged
+  count is ultimately rejected as a bad proof.
+
+  Be precise about *when* that binding happens, though: for the non-Merk
+  lower layers it is **after** the arithmetic, not before. `verify.rs:1953-2016`
+  dispatches the MMR / bulk-append / dense verifiers using an `Element`
+  decoded from the proof's own `value_bytes`, and only at `verify.rs:2018`
+  does `combine_hash(value_hash(value_bytes), &lower_hash)` check it against
+  the parent-committed hash. A forged count therefore *does* reach the math;
+  what makes that safe is that the release build wraps and the chain check
```
QuantumExplorer, line 43: [P2] Use the actual verifier bound for #693

`verify_and_compute_root` rejects `height` outside `1..=16` before this calculation. Therefore `chunk_item_count = 2^height` is at least 2, so for every `u64 total_count`, `completed_chunks <= floor(u64::MAX / 2) = 2^63 - 1`. Consequently `2 * completed_chunks - popcount(completed_chunks)` cannot overflow. Neither release wrapping nor an `overflow-checks = true` panic is reachable on this verifier path. Please document this bound and remove the wrap/panic assertions; as written, the audit guide records behavior that cannot occur and the unchanged “Becomes real if” rule immediately matches the newly described unauthenticated input.

docs/audit-non-issues.md
```diff
@@ -94,3 +107,33 @@ Two related claim shapes, both resolved:
...
...licit version gates and could execute writes instead of returning `VersionError` | Three of the seven cited refs are stale line numbers resolving to `root_key`, `root_hash` and `verify_grovedb` — two accessors and a verification helper, none version-dependent, none a write path; gating them would break callers on any unrelated slot bump. The four typed non-Merk append entry points (`mmr_tree_append`, `bulk_append`, `dense_tree_insert`, `commitment_tree_insert`) do lack a *top-level* gate but already fail closed deeper and **before any mutation**, via their cost dispatchers (`grovedb-merkle-mountain-range/src/cost/mod.rs:88-98`, `grovedb-bulk-append-tree/src/cost/mod.rs:182-201`, `grovedb-dense-fixed-sized-merkle-tree/src/tree/root_maintenance/mod.rs:56-71`, `grovedb-commitment-tree/src/commitment_tree/cost/mod.rs:43,73`). Writes also land in a `StorageBatch` committed only on success. The genuine residual is narrower and different: those four types have no `element_creation` gate of the kind `PrivateDocumentStore` uses (`grovedb_versions.rs:180-186`, enabled only in `v4.rs:407`) — a design-consistency question about when they become *creatable*, worth filing separately if wanted. |
```
QuantumExplorer, line 133: [P2] Qualify the StorageBatch atomicity claim

The direct typed append operations commit their child `data_batch` into the transaction before updating the parent Merk; the source comments explicitly say a later parent-update failure leaves orphaned subtree data in the transaction and that a caller-supplied transaction must be rolled back. For example, `mmr_tree_append` can pass the MMR cost gate, stage and commit its child writes, and then fail a later parent `insert_subtree` version gate. The early lower-layer dispatcher supports the narrow argument, but “committed only on success” is not generally true. Please qualify this to locally owned transactions/the caller rollback contract, or remove the sentence and narrow the “Becomes real if” condition.

docs/audit-non-issues.md
```diff
@@ -94,3 +107,33 @@ Two related claim shapes, both resolved:
...
+| [#721](https://github.com/dashpay/grovedb/issues/721) (closed) | `grovedbg` build script downloads a release artifact at build time | Acceptable posture. Gated off by default (`default = ["full", "estimated_costs"]`; with `grovedbg` off, `build.rs` compiles to a literal no-op and `reqwest`/`sha2` are not even resolved), and integrity is enforced by a pinned version tag plus a pinned `GROVEDBG_SHA256` asserted at `build.rs:33`. A substituted, tampered, or 404 artifact fails the build — fail-closed. Real but minor DX defect not named in the issue: the `if !grovedbg_zip_path.exists()` guard at `:15` caches a bad download, so every later rebuild fails the SHA assert without re-fetching until `target/` is cleared. |
```
QuantumExplorer, line 134: [P3] Say optional dependencies are not activated, not unresolved

Cargo still resolves optional dependencies while producing the workspace lockfile even when their features are disabled. A feature-off focused test run here generated a `Cargo.lock` containing both `reqwest` and `sha2`; they are absent from the active feature-off build graph, so the accurate claim is that they are not activated, built, or executed. “Not even resolved” overstates the default-feature isolation.

QuantumExplorer (commented):
Found three documentation-correctness issues. Most importantly, the corrected #693 rationale still identifies the wrong verifier invariant: the validated bulk-tree height bounds completed_chunks below the multiplication overflow threshold, so the asserted release wrap and overflow-check panic cannot occur. The transaction and optional-dependency statements also need narrower wording. Focused bulk-append, dense-tree, and version tests passed (367 tests), and git diff --check passed.

Issue #835, opened 2026-08-26
Pin envelope byte-equality: prove_query on single-path axis queries vs the direct indexed-axis provers


QuantumExplorer:
Companion to the `PathQueryRun` skipped-field issue, for the same goal: platform routing all ranked / having-range queries through the unified `PathQuery` surface and retiring the direct `indexed_*` call sites.

## What

A golden test (or family of tests) asserting that for the same state and arguments:

- `prove_query(PathQuery::new_axis_top_k(path, axis, k, offset, descending))` emits **byte-identical** proof bytes to `prove_indexed_axis_top_k_paginated(path, axis, k, offset, descending, ...)`, and
- `prove_query(PathQuery::new_axis_bounded(path, axis, lo, hi, limit, descending))` emits byte-identical bytes to the direct `prove_indexed_{count,sum,avg}_query(...)` for the equivalent bounds,

across all three axes, empty and populated secondaries, offset 0 / mid / past-the-end, and both directions — plus the cross-check that `verify_path_query` and the direct `verify_indexed_axis_*` entry points accept each other's bytes.

## Why

The unified dispatch was built over the same indexed-axis proof family, so equality is expected — but platform's switch from the direct provers/verifiers to `prove_query`/`verify_path_query` should be a **pure refactor**, proven by a pinned test rather than assumed. If the bytes genuinely differ somewhere, that is not a blocker (the surface is unreleased and platform owns both prover and verifier, so both sides can switch together) — but the difference should be a documented, deliberate fact rather than a surprise in a platform CI run.

## Acceptance

- Byte-equality goldens in place for the top-k-paginated and bounded shapes, or — where equality does not hold — a documented enumeration of the divergence and mutual-acceptance tests between the two verifier entry points.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

QuantumExplorer:
Superseded by #839: the standalone indexed-axis provers/verifiers (and the standalone unproved read entry points) are retired outright — PathQuery is the only public surface, so there are no longer two public entry points to hold byte-equal. The platform switch is validated by its own proof round-trip and attested-offset regression suites instead.

Issue #834, opened 2026-08-26
Carry the attested skip on PathQueryRun's axis variants


QuantumExplorer:
Platform wants to route **all** ranked / having-range document queries — single-prefix included — through the unified `PathQuery` surface (`run_path_query` / `prove_query` / `verify_path_query`), retiring the direct `indexed_*` call sites in its executors. One capability gap blocks that today.

## The gap

Single-prefix ranked queries support `OFFSET`, and their page carries the **count-commitment-attested skip**: the direct read primitives return `IndexedTopKKeysPage { entries, skipped }`, and on the proved side `VerifiedPathQuery::AxisEntries` already carries `skipped: Option<u64>`. But the unproved unified read does not: `PathQueryRun::AxisEntries(AxisEntries)` and `PathQueryRun::AxisKeys(AxisKeys)` return only the page. Moving platform's single-prefix unproved reads onto `run_path_query` would silently drop the attestation — including the "an offset past the end returns an empty page whose `skipped` attests the population" semantics.

## The change

Extend the two single-path axis variants to carry the skip alongside the page:

- `PathQueryRun::AxisEntries { entries, skipped: Option<u64> }`
- `PathQueryRun::AxisKeys { keys, skipped: Option<u64> }`

`Some(n)` for `AxisTraversal::RankedPage` traversals — derived from the counted subtree commitments exactly as `indexed_*_top_k_paginated*` derives it, equal to the requested offset on a full page and smaller when the walk exhausted the secondary — and `None` for `Bounded` (no skip concept), mirroring the proved side's existing contract.

The **branched** variants stay unchanged: a non-zero offset is rejected together with branching upstream, so a branched read's skip is structurally zero, and per-branch skip carries no meaning for a merged union.

Breaking change to `PathQueryRun` (in-tree callers and dashpay/platform adapt at the pin bump; the enum is V4-era API with no released consumers).

## Acceptance

- The unproved unified read of a `RankedPage` traversal returns the same `(entries, skipped)` pair as the direct primitive for the same state and arguments, across offset 0, mid-population offsets, and past-the-end offsets (empty page, `skipped` = population).
- `Bounded` traversals return `skipped: None`.

Part of the platform-side single-read-path unification; companion issue: prove/verify envelope byte-equality for single-path axis queries.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #839, opened 2026-08-26
feat!: retire the standalone indexed-axis provers/verifiers — PathQuery is the only public proof surface

QuantumExplorer:
## What

Removes the standalone indexed-axis proof family from the public API. `PathQuery`'s axis constructors (`new_axis_top_k`, `new_axis_bounded`, `new_axis_rank_of_key`, `new_axis_aggregate_over_value_range`, `new_branched_axis`) with `prove_query` / `verify_path_query` are now the **only** public proof surface for indexed-axis reads.

The retired entry points — `prove_indexed_axis_*`, `verify_indexed_axis_*`, the per-axis `axis_api` wrappers, the standalone envelope builders, and their verification cores — are now compiled `#[cfg(test)]`. They survive solely as in-crate oracles: the test suites (notably `envelope_byte_equality_tests.rs` from #837) cross-check the unified V1-envelope axis proofs against this independent implementation of the same engines, byte for byte.

## Why now

The standalone envelope wire format (`IndexedAxisRangeProof` / `IndexedAxisPaginatedProof` / `IndexedAxisAggregateProof`) has **never been emitted by a released version** — indexed trees are GROVE_V4-gated and V4 is not live. `envelope.rs` declares these types consensus-frozen once a version carrying them activates. Retiring them from the public API before V4 activates means the format never freezes: only the V1 envelope's axis-descent format ships, and GroveDB never has to maintain two frozen wire formats for the same reads.

## What is unchanged

- **The engines**: descent payload builders, secondary-proof builders, target-chain resolution, `axis_lowering` — the unified surface is a thin envelope over exactly this code.
- **Trusted reads**: `indexed_*_top_k*`, `indexed_*_range`, aggregate reads, and the `_keys` projections stay public.
- **Envelope and result types** stay public for the oracles and downstream transition code.
- Zero production-code behavior change: everything gated was reachable only from the retired public entry points (the compiler's dead-code analysis was the audit — every newly gated item was dead in lib builds after the entry points were demoted).

## Docs

- `docs/book/src/unified-path-query.md`: "Relationship to the specialized surfaces" now names `PathQuery` as the only public axis-proof surface and explains the never-freeze rationale.
- `docs/book/src/count-indexed-tree.md`: proof examples rewritten onto `new_axis_top_k` / `new_axis_bounded` / `new_axis_aggregate_over_value_range` (the count-band question now uses the O(log n) `Population`-fold aggregate rather than enumerate-and-count).
- The V0 rejection message for indexed subqueries now points at `PathQuery::new_axis_top_k` instead of the retired entry point (4 tests pinning the message updated).

## Trusted reads (second commit)

The per-axis trusted-read wrappers (`indexed_*_top_k*`, `indexed_*_range*`, the aggregate reads and the `_keys` projections) are now `pub(crate)` as well: they are the engine `run_path_query` routes axis shapes to, not a public API. External callers build the same axis `PathQuery` for reads and proofs alike — one request shape, three consumers (`run_path_query`, `prove_query`, `verify_path_query`). The non-paginated top-k family is `#[cfg(test)]` (nothing routes to it — the paginated walk with `offset = 0` serves that case). `IndexedTopKPage` / `IndexedTopKKeysPage` stay re-exported for downstream transition code.

## Platform

Platform has four remaining direct call-site groups (ranked top-k prove/verify in `drive_document_ranked_query` / `verify_ranked_top_k_proof`, having-range prove/verify in `drive_document_having_query` / `verify_having_range_proof`). A companion platform PR switches them to the unified surface; since `prove_query`/`verify_path_query` already exist at platform's current grovedb pin, the two PRs are independently mergeable — but platform cannot bump past this commit without switching, which is the point.

Relates to #835 (byte-level relationship pinned in #837) and #834.

## Summary by CodeRabbit

- **New Features**
  - Added unified path-query support for indexed-tree top-k, range, aggregate, and population queries.
  - Axis queries support ascending or descending traversal, bounded conditions, pagination, and efficient population counts.
  - Aggregate sum queries and proof verification remain available through the unified query interface.

- **Breaking Changes**
  - Dedicated indexed-tree read and proof APIs are no longer part of the public interface.
  - Indexed subqueries now require V1 proof envelopes; V0 proofs are unsupported.

- **Documentation**
  - Updated indexed-tree and proof guidance with the supported query and verification workflows.

### DIFF
--- a/docs/book/src/count-indexed-tree.md
+++ b/docs/book/src/count-indexed-tree.md
@@ -478,13 +478,32 @@ Merk is not touched. The verifier receives the primary's root hash plus a
 ### Top-k by count
 
 ```rust
-// Shipped API on `GroveDb`:
-let entries: Vec<IndexedAxisEntry<u64>> = db
-    .indexed_count_top_k(path, k, /* descending: */ true, transaction, grove_version)?
-    .expect("top-k");
+// Trusted read — the axis read through the unified PathQuery surface
+// (the only public read surface for indexed-axis queries):
+let path_query = PathQuery::new_axis_top_k(
+    path_vec.clone(),
+    IndexAxis::Count,
+    k,
+    /* offset: */ 0,
+    /* descending: */ true,
+);
+let PathQueryRun::AxisEntries { entries, skipped } = db
+    .run_path_query(
+        &path_query,
+        true,  // allow_cache
+        true,  // decrease_limit_on_range_with_no_sub_elements
+        true,  // error_if_intermediate_path_tree_not_present
+        QueryResultType::QueryPathKeyElementTrioResultType,
+        transaction,
+        grove_version,
+    )?
+    .expect("top-k")
+else {
+    unreachable!("an axis read runs to AxisEntries")
+};
+// entries: AxisEntries::Count(Vec<IndexedAxisEntry<u64>>)
 
-// Verifiable variant — the axis read through the unified PathQuery
-// surface (the only public proof surface for indexed-axis reads):
+// Verifiable variant — the same PathQuery, proved:
 let path_query = PathQuery::new_axis_top_k(
     path_vec,
     IndexAxis::Count,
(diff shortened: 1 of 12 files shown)


QuantumExplorer (commented):
Reviewed

PR #844, opened 2026-08-31
feat: per-instance query limits (Query::limit) — wire format, trusted-read engine, fail-closed gates

QuantumExplorer:
Part 1 of 3 of the per-instance ("nested") query limits feature. This PR adds the field, the wire format, the trusted-read engine, and fail-closed gates everywhere else. Part 2 teaches the V1 prover/verifier the accounting and lifts merge restrictions; part 3 is the book chapter.

## What a per-instance limit is

`SizedQuery::limit` is one global budget shared by the whole traversal — you can't say "return each parent's top 3 children", and merging limited path queries has always been refused because a single counter can't represent "5 from A plus 7 from B".

`Query::limit` is **per execution instance**: a query node runs once for every parent key it is reached under, and each run gets a fresh budget of `limit` rows for everything originating in that instance's subtree. Caps compose by `min` with every enclosing budget (ancestor instances and the global limit). The root node executes exactly once, so its cap ≡ the global limit.

## Wire format

`Query` encoding version 3, selected only when the node actually carries a limit: version byte `3`, then a flags byte (bit 0 = read mode present, bit 1 = per-instance limit — always set, the encoding is canonical), standard fields, then the appended payloads. Every already-expressible query keeps its exact version-1/2 bytes (golden pins extended with v3 pins + canonicality/fail-closed decode tests). Decoders that predate the field reject version 3 — fail closed by construction, same as `read_mode`'s version 2. `Some(0)` is rejected as malformed at every serving entry point.

## Trusted-read engine (GROVE_V4)

New `element.path_query_push` **v2** (V4 was on v1, unreleased), threading a `QueryBudget` per recursion frame:

- global/offset keep the long-standing copy-and-reconcile scheme;
- the instance chain seeds `min(inherited remaining, own Query::limit)` at each descent, so each node's cap is fresh per parent while ancestors keep binding;
- **descents now reconcile by total consumed budget (rows + empty-subtree charges) instead of returned rows.** v0/v1 silently drop empty-subtree charges made below the first nesting level, while the prover threads one shared counter and keeps them — a latent read-vs-proof divergence. v2 aligns the read path with the prover's accounting (pinned by a differential test against GROVE_V3's legacy behavior).

Empty-subtree charges consume only the global budget by default (they exist to bound traversal work); the new `QueryOptions::decrease_instance_limits_on_range_with_no_sub_elements` (default `false`) opts them into instance budgets too.

## Fail-closed everywhere else

New version slot `path_query_methods.per_instance_query_limits` (0 in V1–V3, 1 in V4):

- pre-V4 grove versions reject limit-carrying queries at `query_raw` (and per-frame in the engine as defense in depth);
- **proof generation and verification reject** at every entry until part 2 lands;
- `PathQuery::merge` and the whole `Query`/`SubqueryBranch` merge family refuse limit-carrying nodes (same `_unchecked`-family invariant as read modes);
- read-mode shapes, aggregate leaf/carrier validators, and count-offset-paginated validation reject;
- `query_keys_optional` / `query_raw_keys_optional` reject — their terminal-keys projection would report keys beyond an instance cap as falsely absent (same hazard class that blocks absence proofs, which stay rejected in part 2 as well).

## Tests

- 14 new tests in `per_instance_limit_tests.rs`: top-k-per-parent, global∧instance composition, root-cap ≡ global, conditional-branch caps, ancestor caps binding descendants, reverse direction, offset interplay (skips don't consume instance budgets), empty-subtree charge routing (default vs opted-in), the nested-charge global alignment with a GROVE_V3 legacy differential, and all the fail-closed gates.
- Golden encoding pins for v3 + canonicality (v3 without the limit flag / unknown flag bits / truncated payload all rejected); v1/v2 pins byte-identical.
- Full workspace suite: 5118 passed / 0 failed.

## Summary by CodeRabbit

* **New Features**
  * Added per-instance result limits for nested queries, allowing each query instance to receive its own cap.
  * Limits compose with overall query limits, offsets, nested branches, and directional queries.
  * Added support for encoding, decoding, and displaying per-instance limits.

* **Bug Fixes**
  * Improved nested-query budget accounting, including empty results and offsets.

* **Compatibility**
  * Per-instance limits are supported for trusted reads in GroveDB V4 and later.
  * Unsupported versions, proofs, projections, and query merges now fail clearly and safely.


grovedb-query/src/query.rs
```diff
@@ -67,16 +67,64 @@ pub struct Query {
...
+    /// may select nothing is a malformed query, not an empty result).
+    /// Serving is version-gated (`GROVE_V4`+); older grove versions
+    /// fail closed, and on the wire a query carrying a per-instance
+    /// limit anywhere encodes that node as version 3, which decoders
+    /// that predate the field reject.
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "Option::is_none")
+    )]
+    pub limit: Option<u16>,
```
QuantumExplorer, line 109: [P2] Keep serde field positions stable for positional formats

`Query` derives serde, while both `read_mode` and this new field use `skip_serializing_if = "Option::is_none"`. In a positional/non-self-describing serializer such as serde-bincode, omitted fields have no tags: I reproduced an existing `read_mode: Some(_), limit: None` round trip failing with `UnexpectedEnd`, and a `read_mode: None, limit: Some(3)` payload shifting the limit bytes into `ReadMode` and failing on an invalid variant. This regresses a previously working serde shape even though the custom bincode codec is fine. Please use a custom/versioned serde representation (or otherwise keep positional fields present) and add serde-bincode tests for all four option combinations.

grovedb-query/src/merge.rs
```diff
@@ -312,6 +333,21 @@ impl Query {
                     .to_string(),
             ));
         }
+        let carries_limit = |branch: &SubqueryBranch| {
+            branch
+                .subquery
+                .as_deref()
+                .is_some_and(|subquery| subquery.has_instance_limit_anywhere())
+        };
+        if carries_limit(&self.default_subquery_branch)
```
QuantumExplorer: [P2] Include existing conditional branches in this limit gate

This only scans the two default branches, but the unchecked merge below can promote `self.default_subquery_branch` into a conditional and merge it with an already-existing conditional branch. For example, if the receiver default has path `[k]`, its existing `Key(k)` conditional carries `Query::limit`, and the incoming default is pathless, line 498 reaches `merge_conditional_boxed_subquery_unchecked` and merges into the limited branch after this check passes. That lets a public merge conflate result sets governed by an instance cap. Please reject when `self.has_instance_limit_anywhere()` is true (or explicitly scan the existing conditionals) and add this overlapping-promotion regression case.

grovedb/src/query/mod.rs
```diff
@@ -758,6 +782,61 @@ impl PathQuery {
...
+    /// and a zero cap is rejected outright — a node that may select
+    /// nothing is a malformed query, not an empty result.
+    pub(crate) fn reject_unserved_per_instance_limits(
+        &self,
+        grove_version: &GroveVersion,
+    ) -> Result<(), Error> {
+        if !self.has_instance_limits() {
+            return Ok(());
+        }
+        if grove_version
```
QuantumExplorer: [P2] Validate the capability and engine tuple exactly

This gate rejects only slot `0`, so an unknown future value such as `2` silently runs today's semantics instead of failing with `UnknownVersionMismatch`. It is also independent of `element.path_query_push`: a public cloned/custom V4 table with `per_instance_query_limits = 1` and `path_query_push = 0` admits a limited query, then the v0 descent drops the ancestor instance budget. The canonical V1-V4 tables are coherent, but this still fails open for custom/future tables. Please require the known capability value exactly and ensure it selects the instance-aware engine (currently `1`), or make one selector authoritative; mirror the same check in the per-frame defense.

grovedb/src/operations/proof/verify.rs
```diff
@@ -219,6 +219,12 @@ impl GroveDb {
         // thing.
         query.reject_unserved_read_mode()?;
 
+        // Per-instance limit gate: the verifier's limit accounting is a
+        // single shared counter and does not (yet) express per-instance
+        // caps. Fail closed rather than accepting a proof over a result
+        // set the caps would have shrunk.
+        query.reject_per_instance_limits("proof verification")?;
```
QuantumExplorer, line 239: [P2] Reject unsupported limited queries before decoding the proof

All public verification wrappers decode caller-supplied proof bytes before reaching this gate (`verify_query_with_options` at line 115, the parent-info variant at 159, and `verify_query_raw` at 187). The canonical decoder permits envelopes up to 256 MiB, so an unsupported limited query can still force bounded but avoidable parsing/allocation and returns a proof-decode error before the intended `NotSupported`. Please run the recursive `reject_per_instance_limits` check before `decode_grovedb_proof_canonical` in every public wrapper, while retaining these internal checks as defense in depth. An empty malformed proof plus a limited query is a useful error-precedence test.

grovedb/src/element/path_query_push/v1.rs
```diff
@@ -78,125 +101,119 @@ pub(crate) fn path_query_push_v1(
...
-                };
-
-                if let Some(limit) = limit {
-                    *limit = limit.saturating_sub(1);
+                    // The terminal was absent. v1 charges the limit
+                    // unconditionally here; v1 keeps the global charge
+                    // (it bounds walks across many missing terminals)
+                    // but classes it with the empty-subtree charges for
+                    // the instance chain: no row was produced.
+                    budget.charge_empty_subtree(
```
QuantumExplorer, line 222: [P2] Honor the governing empty-range option for instance charges

`QueryOptions` documents `decrease_instance_limits_on_range_with_no_sub_elements` as having no effect when `decrease_limit_on_range_with_no_sub_elements` is false, but this absent-terminal branch passes only the subordinate instance flag to `charge_empty_subtree`. With global empty charging disabled and instance empty charging enabled, an earlier missing terminal consumes the instance cap and can suppress a later real row, contrary to the public option contract. Please pass the conjunction of both options for the instance charge (while preserving any intentional legacy global charge) and add a missing-first/present-second regression test.

grovedb/src/element/query.rs
```diff
@@ -624,16 +439,351 @@ impl ElementQueryExtensions for Element {
...
+    let mut cost = OperationCost::default();
+
+    // Per-instance limits are serving-gated: grove versions whose
+    // engines don't account for them must fail closed rather than run
+    // the query with its caps silently ignored. The zero cap is
+    // rejected outright — a node that may select nothing is a
+    // malformed query, not an empty result. Checking only this node's
+    // own `limit` keeps the check O(1) per frame; deeper caps are
+    // caught by their own frames before they could take effect.
+    if let Some(instance_limit) = sized_query.query.limit {
```
QuantumExplorer, line 686: [P2] Validate the complete query at the public low-level boundary

This per-frame check only runs for nodes the walk actually reaches. The public `Element::get_path_query` / `get_query_apply_function` wrappers enter the internal walk without a recursive preflight, so an unmatched conditional branch containing `Query::limit` is accepted under GROVE_V3, and an unmatched `Some(0)` is accepted under the latest version. `GroveDb::query_raw` is protected by its whole-query gate, but direct callers of these public trait methods bypass it. Please validate all nested limits once in the public wrapper(s), keeping this O(1) frame check as defense in depth, and cover an unmatched limited conditional.

QuantumExplorer (commented):
Fresh full review of the current PR head. I found the inline issues below after revalidating the complete diff, the latest path_query_push v1 fold, and the targeted per-instance-limit tests.

grovedb-query/tests/instance_limit_gates.rs
```diff
@@ -0,0 +1,202 @@
...
+        Err(Error::NotSupported(message)) => {
+            assert!(message.contains("per-instance"), "got: {message}")
+        }
+        other => panic!("merge_with must refuse a carrying receiver, got {other:?}"),
+    }
+}
+
+#[test]
+fn version_3_round_trips_through_the_owned_decode_path_too() {
+    // `decode_from_slice` exercises `BorrowDecode`; a reader exercises
```
QuantumExplorer: [P3] Exercise V3 through the borrowed decoder

The comment is backwards: `decode_from_slice` exercises owned `Decode`, as does the `decode_from_std_read` call in this test. The only explicit `borrow_decode_from_slice` round trip elsewhere encodes a V1 query, so the separate V3 branch of `Query::borrow_decode_with_depth`—including flags, read mode, and limit parsing—remains untested. Please add a borrowed V3 round trip plus malformed-flags coverage using these V3 bytes.

grovedb-query/src/query.rs
QuantumExplorer, line 109: [P3] Update the read-mode wire-version documentation

The `read_mode` docs above still say `Some(_)` always bumps the query to wire version 2. With this field present as well, the node uses version 3 and carries both flags. Please describe the lowest representable version for the combinations so readers do not infer that a read-mode-plus-limit query remains v2.

grovedb/src/element/query.rs
```diff
@@ -624,16 +439,351 @@ impl ElementQueryExtensions for Element {
...
+/// The body of [`ElementQueryExtensions::get_query_apply_function`] —
+/// one frame of the trusted query walk.
+///
+/// `inherited_instance_limit` is the enclosing frame's remaining
+/// per-instance budget (`None` at the root and everywhere on queries
+/// without per-instance limits); it is `min`-combined with this query
+/// node's own `Query::limit` to seed the frame's instance budget.
+/// Returns `(elements, skipped, consumed)` where `consumed` is
+/// everything this frame's subtree charged against the global budget —
+/// result rows plus empty-subtree charges — which the v2
```
QuantumExplorer: [P3] Rename the remaining v2 engine references after the fold

The latest commit folded the never-shipped `path_query_push` v2 implementation into v1, but this new documentation still calls it the v2 engine. The same stale label remains in `element/query_budget.rs` and the consumed-budget test comment. Please update these references to v1 so the docs match the actual V4 selector and dispatcher.

grovedb-query/src/query.rs
QuantumExplorer: Fixed in c73322a8: removed `skip_serializing_if` from both optional fields (keeping `serde(default)` for deserialization) so positional serde formats keep their presence tags, and added serde-bincode round-trip tests for all four `(read_mode, limit)` option combinations.

grovedb-query/src/merge.rs
QuantumExplorer: Fixed in c73322a8: the gate now scans the whole receiver (`has_instance_limit_anywhere`) plus the incoming branch, and the overlapping-promotion case (receiver default path `[k]`, existing limited `Key(k)` conditional, pathless incomer) is a regression test.

grovedb/src/query/mod.rs
QuantumExplorer: Fixed in c73322a8: the capability slot is matched exactly everywhere — unknown future values return a typed `VersionError` — and serving additionally requires a coherent table (`element.path_query_push >= 1`); a table serving the capability over the v0 engine is refused as `CorruptedCodeExecution`. Both checks are mirrored in the O(1) per-frame defense.

grovedb/src/operations/proof/verify.rs
QuantumExplorer: Fixed in c73322a8: all three public wrappers (`verify_query_with_options`, the parent-tree-info variant, `verify_query_raw`) now run the recursive limited-query gate before `decode_grovedb_proof_canonical`, with the internal checks kept as defense in depth; added the error-precedence test (empty undecodable proof + limited query → `NotSupported`).

grovedb/src/element/path_query_push/v1.rs
QuantumExplorer: Fixed in c73322a8: the absent-terminal charge now passes the conjunction of both flags, and the descent reconciliation does too (child `consumed` can carry the unconditional absent-terminal global charges, so the instance chain must not absorb them either while the governing flag is off); added the missing-first/present-second regression test.

grovedb/src/element/query.rs
QuantumExplorer: Fixed in c73322a8: `Element::get_path_query` and `get_query_apply_function` run the recursive whole-query gate once up front (shared `reject_unserved_instance_limits_in_query`), keeping the per-frame check as defense in depth; covered an unmatched limited conditional failing closed under GROVE_V3 and an unmatched `Some(0)` rejected at latest.

grovedb-query/tests/instance_limit_gates.rs
QuantumExplorer: Fixed in c73322a8: corrected the backwards comment (`decode_from_slice` is the owned path) and added borrowed-decoder V3 round trips — limit-only and both-flags — plus malformed-flags rejections through both decode paths.

grovedb-query/src/query.rs
QuantumExplorer: Fixed in c73322a8: the `read_mode` docs now state the lowest-representable-version rule — version 2 when the read mode is alone, version 3 with both flags when a per-instance limit rides along.

grovedb/src/element/query.rs
QuantumExplorer: Fixed in c73322a8: renamed the remaining v2-engine references (the internal walk doc in `element/query.rs`, `query_budget.rs`, and the consumed-budget test comment) to v1.

QuantumExplorer (commented):
Reviewed

grovedb-query/src/query.rs
```diff
@@ -62,21 +65,75 @@ pub struct Query {
...
+    /// may select nothing is a malformed query, not an empty result).
+    /// Serving is version-gated (`GROVE_V4`+); older grove versions
+    /// fail closed, and on the wire a query carrying a per-instance
+    /// limit anywhere encodes that node as version 3, which decoders
+    /// that predate the field reject.
+    ///
+    /// Serde note: always serialized, absence tolerated only on
+    /// deserialization — see [`read_mode`](Self::read_mode) for why
+    /// `skip_serializing_if` must not be added.
+    #[cfg_attr(feature = "serde", serde(default))]
```
QuantumExplorer, line 108: [P2] Version the Serde representation before adding `limit`

The new same-version round-trip test passes, but this derived struct is still unsafe across versions in both directions. I reproduced a valid base-version serde-bincode payload with `read_mode: Some(_)` failing at this head with `UnexpectedEnd { additional: 1 }`: `serde(default)` does not turn EOF into an omitted positional field. Conversely, serializing `limit: Some(3)` as JSON at this head and deserializing it with the base `Query` layout succeeds because Serde ignores the unknown `limit` key, silently turning the bounded query into an unlimited one. Please use a versioned/custom Serde representation (or otherwise make limit-bearing payloads fail closed for old readers) and add base-writer/head-reader plus head-writer/base-reader tests; current/current round trips cannot catch either regression.

grovedb/src/element/query.rs
```diff
@@ -402,182 +371,48 @@ impl ElementQueryExtensions for Element {
...
+        // This legacy signature threads only the global limit/offset
+        // pair, and a per-instance budget cannot ride it: the budget is
+        // per node *instance*, while this entry point is called once
+        // per item — re-seeding a fresh cap per item would silently
+        // change what the cap means. The engine's own walk seeds the
+        // instance budget in `get_query_apply_function_internal`; a
+        // direct caller whose queried node carries its own cap fails
+        // closed here instead of having the cap ignored. (Caps on
+        // subqueries BELOW this node are served through the descent.)
+        if sized_query.query.limit.is_some() {
```
QuantumExplorer, line 388: [P2] Recursively preflight `query_item`'s entire query

This public low-level entry point checks only the current node's `limit`, unlike the new recursive preflight on the other Element wrappers. I reproduced `Element::query_item` returning `Ok(())` under `GROVE_V3` when the queried node has no limit but an unmatched conditional branch carries `limit: Some(1)`. If stored data later makes that branch match, the same shape fails during descent, so acceptance depends on database contents. Run `reject_unserved_instance_limits_in_query(&sized_query.query, grove_version)` here as well (while retaining the explicit current-node rejection required by this legacy signature), and cover the unmatched-conditional case directly through `query_item`.

grovedb/src/element/query.rs
```diff
@@ -624,16 +458,371 @@ impl ElementQueryExtensions for Element {
...
+        sized_query.limit,
+        QueryBudget::min_caps(inherited_instance_limit, sized_query.query.limit),
+        original_offset,
+    );
+
+    if sized_query.query.left_to_right {
+        for item in sized_query.query.iter() {
+            cost_return_on_error!(
+                &mut cost,
+                query_item_internal(
```
QuantumExplorer, line 741: [P2] Preserve the subordinate method-version checks

Calling `query_item_internal` directly here bypasses `Element::query_item`'s `check_grovedb_v0_with_cost!`; similarly, `get_path_query_internal` now calls `get_query_apply_function_internal` directly and bypasses that public method's version check. These checks were enforced by the base composition. In a real-db reproduction, setting either `element.query_item = 1` or `element.get_query_apply_function = 1` made `query_raw` return the stored row successfully instead of `VersionError`. Please validate the subordinate slots before entering the internal helpers (or formally remove/retire those slots) so future/replay version tables cannot silently run today's semantics.

grovedb/src/query/mod.rs
```diff
@@ -1632,6 +1695,67 @@ impl HasSubquery<'_> {
...
+        .grovedb_versions
+        .path_query_methods
+        .per_instance_query_limits
+    {
+        0 => Err(Error::NotSupported(
+            "per-instance query limits (Query::limit) require a grove version that serves them"
+                .to_string(),
+        )),
+        1 => {
+            if grove_version.grovedb_versions.element.path_query_push == 0 {
```
QuantumExplorer, line 1736: [P2] Reject unknown `path_query_push` engines in the coherence preflight

This guard rejects only engine `0`, so every unknown nonzero selector is treated as accounting-capable even though the dispatcher supports exactly `[0, 1]`. I reproduced `query_raw` returning `Ok(empty)` for a limited missing-key query with capability `1` and `element.path_query_push = 2`; no row means the dispatcher is never reached. The same table becomes a `VersionError` only when data matches, making version validation data-dependent. Exact-match this slot here and return the typed unknown-version error for values other than `1`.

QuantumExplorer (commented):
Fresh rereview of the complete PR at c73322a8dc58d5ee045d07776336132640cdd89e. The latest commit fixes the previously reported merge, proof-precheck, empty-range, borrowed-decode, wrapper-preflight, and capability-value issues. Four additional reproducible findings remain.

grovedb-query/src/query.rs
QuantumExplorer: Fixed in #847: the serde representation is now versioned and hand-written — fixed always-present positional layout with a leading version; flat base-compatible JSON for v1/v2; the limit-bearing v3 nests under `body` so a pre-limit reader hard-fails instead of silently decoding an unlimited query; unknown flat keys refused; canonical version claims enforced. Base-writer/head-reader and head-writer/base-reader tests cover both format families.

grovedb/src/element/query.rs
QuantumExplorer: Fixed in #847: `Element::query_item` now runs the recursive `reject_unserved_instance_limits_in_query` preflight (keeping the current-node rejection its legacy signature requires), and the unmatched-conditional case is covered directly through `query_item` under GROVE_V3.

grovedb/src/element/query.rs
QuantumExplorer: Fixed in #847: `query_item_internal`, `get_query_apply_function_internal`, and `get_path_query_internal` validate their subordinate method-version slots themselves, so the engines' direct entry no longer bypasses them; doctored-slot tests pin all three returning `VersionError`.

grovedb/src/query/mod.rs
QuantumExplorer: Fixed in #847: both coherence gates (the shared recursive gate and the per-frame mirror) exact-match `path_query_push` — 0 is the incoherent-table refusal, 1 serves, anything else is a typed `UnknownVersionMismatch` — and the data-independent rejection (limited missing-key query) is a regression test.

PR #845, opened 2026-08-31
feat: serve per-instance query limits in V1 proofs and lift them through merges

QuantumExplorer:
Part 2 of 3, stacked on #844 (merge that first; this PR's base is its branch). Part 3 is the book chapter.

## V1 prover/verifier accounting

The bare `overall_limit: &mut Option<u16>` / `limit_left` counters become a shared **`V1LimitState { global, consumed_rows, consumed_total }`** threaded through both walks, plus a **frame-local instance budget** per layer: `min(inherited remaining, the layer query's own Query::limit)`, resolved through the new `SinglePathSubquery::instance_limit` (so synthesized path-component layers pass the chain through untouched, and the chain starts at the query root automatically).

- Each layer's merk walk runs under `min(global, instance)` — merk's per-layer "proof returns more data than limit" check is what enforces per-instance over-delivery on the verifier side (pinned by an adversarial test: a proof built for cap 4 must not verify under cap 2).
- Rows charge every budget; descents settle the enclosing frame from the `consumed_rows` delta; the empty-layer charge (`decrease_limit_on_empty_sub_query_result`) stays **global-only** — it bounds traversal work, instance caps bound rows. (No new `ProveOptions` field: V0 proofs bincode-embed `ProveOptions`, so a new field there would break the frozen wire format, and the V1 verifier always runs `ProveOptions::default()` anyway.)
- `consumed_total` also replaces the old detect-consumption-by-diffing-the-`Option` trick, which went blind when the global limit was `None` — a real case once instance caps can exist without a global one.
- Non-Merk layers (MMR / BulkAppend / Dense / CommitmentTree) and the count-offset dispatch keep their helper signatures; they're bounded through an effective-limit temp whose delta is absorbed back into the shared state.

**Behavior for every query without per-instance limits is unchanged** — the full workspace suite (5125 tests) passes untouched, which matters because the V1 proof format is live under GROVE_V3. The V0 prover/verifier stay frozen and reject limit-carrying queries at dispatch; absence-proof assembly and pre-V4 versions keep failing closed.

## Merge lifting (`path_query_methods.merge = 2`, GROVE_V4)

`PathQuery::merge` now merges limited path queries by **lifting**: an input's global `SizedQuery::limit` becomes its merged branch's `Query::limit`. The lift is semantically exact — the branch instance executes exactly once (its path is a concrete key chain), so "at most N rows from this input" and "at most N per instance" coincide. Authored per-instance limits ride along on their branches. Per the refuse-on-collision decision:

- a limited input landing **at the merged root** (its body would blend with other inputs' queries) is refused with a typed error;
- two limited branches **colliding on a key** are refused (`limits merge only as exclusive grafts`);
- limit-free inputs keep the v1 merge behavior identically (same code path).

`prove_query_many` therefore now serves limited path queries, and the verifier re-derives the identical merged query at the same grove version.

## Deliberately out of scope

Counting `add_parent_tree_on_subquery` rows against the limit (the M6 known limitation) is **not** included. Making the prover charge exactly where the verifier pushes parent rows requires auditing every descent arm (ordinary, cidx-family, the four non-Merk arms, and the aggregate descents, where the pairing is unclear) — an asymmetry there breaks honest proofs. It deserves its own focused change; the documented limitation stands.

## Tests

`per_instance_limit_tests.rs` grows to 21: proof round-trips differentially checked against the trusted read (top-k per parent, global∧instance composition, conditional-branch caps, ancestor caps, reverse direction), the wide-cap-vs-tight-cap adversarial rejection, absence-mode rejection, pre-V4 prover rejection, merge lift round-trips (instance-capped and global-limited inputs through prove/verify), and the collision/root-landing refusals. `merge_refuses_limits_and_offsets_at_every_merge_version` becomes `merge_refuses_offsets_at_every_version_and_lifts_limits_from_v2`, pinning the pre-v2 refusal and the v2 lifted structure.

## Summary by CodeRabbit

* **New Features**
  * Added support for per-instance query limits in GROVE_V4 trusted reads, V1 proofs, and query merging.
  * Query limits can provide a fresh result budget for each execution instance, such as “top K per parent.”
  * Limits compose with global budgets and are preserved when supported queries are merged.

* **Bug Fixes**
  * Improved proof generation and verification to correctly enforce global and per-instance limits.

* **Documentation**
  * Added comprehensive guidance covering query-limit behavior, version support, merging, and unsupported scenarios.


QuantumExplorer (commented):
Approved

QuantumExplorer (commented):
Reviewed the final merged head `ed50f67755c00c65ec8142a027a4b5e5105a4bbf`. Documentation note (P3): `grovedb-version/src/version/v4.rs:91-99` still says per-instance limits are served only on trusted reads and that proofs reject them. That contradicts this PR's V1 prover/verifier support and the updated generic capability documentation; please update the V4 capability summary as well.

grovedb/src/operations/proof/generate.rs
```diff
@@ -1832,7 +1856,7 @@ impl GroveDb {
         let limit = if path.len() < path_query.path.len() {
             None
         } else {
-            *overall_limit
+            limit_state.effective_layer_limit(frame_instance)
```
QuantumExplorer, line 1859: [P1] Avoid truncating the parent walk with a descendant-row cap. `Query::limit` counts rows returned from the subtrees, but this passes the cap to the parent Merk walk before any descent is evaluated. If the first matching child is empty and a later child contains data, `limit = 1` makes Merk emit only the empty child; the empty-layer logic correctly leaves the instance budget untouched, but the populated child has already been discarded. On this final head I reproduced `a_empty` followed by `b_full/k0`: public prove/verify returned 0 rows although trusted execution returns `b_full/k0`. The parent scan needs to continue past non-producing children (or use a continuation strategy) until the row cap is actually consumed.

grovedb/src/operations/proof/generate.rs
```diff
@@ -2377,18 +2401,28 @@ impl GroveDb {
                                 let mut lower_path = path.clone();
                                 lower_path.push(key.as_slice());
 
+                                let mut non_merk_effective =
```
QuantumExplorer, line 2404: [P1] Compose the MMR child's own instance limit here. `frame_instance` is the parent frame, and this specialized adapter bypasses the recursive call that normally min-composes `query_items_at_path(&lower_path)?.instance_limit`. With no tighter global/ancestor cap, `non_merk_effective` is `None`, so the helper proves every selected MMR leaf; the verifier mirrors the same omission and accepts rows beyond the child `Query::limit`. Derive a child frame from the lower query before generating and verifying this layer.

grovedb/src/operations/proof/generate.rs
```diff
@@ -2402,6 +2436,9 @@ impl GroveDb {
...
```
QuantumExplorer, line 2439: [P1] Apply the BulkAppend child query's instance cap, not just the parent frame. This adapter never creates the lower query frame, so a child `Query::limit = Some(1)` with no outer cap is ignored. I reproduced this through the public APIs with three BulkAppend values: `prove_query` and `verify_query_with_options` succeeded and returned all 3 rows. Please min-compose the lower query's `instance_limit` on both prover and verifier.

grovedb/src/operations/proof/generate.rs
```diff
@@ -2410,12 +2447,19 @@ impl GroveDb {
                                         [0u8; 32], // unused parameter
                                         total_count,
                                         chunk_power,
-                                        overall_limit,
+                                        &mut non_merk_effective,
```
QuantumExplorer, line 2450: [P1] Guard the empty lower range before delegated accounting. With an active global/ancestor/instance cap, an empty BulkAppend child query reaches `generate_bulk_append_layer_proof` with `Some` here; `query_items_to_range` returns `(total_count, 0)`, and line 3599 computes `0 - total_count`. A public `prove_query` reproduction panics in debug (`attempt to subtract with overflow`); release builds wrap and can charge 65,535 rows. CommitmentTree reaches the same helper. Return zero consumption when `start >= end.min(total_count)` (or reject the empty range) before subtracting.

grovedb/src/operations/proof/generate.rs
```diff
@@ -2432,19 +2476,29 @@ impl GroveDb {
                                 let mut lower_path = path.clone();
                                 lower_path.push(key.as_slice());
 
+                                let mut non_merk_effective =
```
QuantumExplorer, line 2479: [P1] Compose the Dense tree child's `Query::limit` into this budget. Like the other non-Merk adapters, this path bypasses recursive frame creation and passes only the parent/global cap; the verifier does the same. A Dense child with more matches than its own cap therefore proves and returns every selected position. Read the lower query's `instance_limit` and min-compose it on both sides.

grovedb/src/operations/proof/generate.rs
```diff
@@ -2458,19 +2512,29 @@ impl GroveDb {
...
```
QuantumExplorer, line 2515: [P1] Compose the CommitmentTree child's `Query::limit` before delegating. This uses only the parent/global frame, then delegates to the same BulkAppend machinery that I reproduced returning 3 rows under a child cap of 1; the verifier mirrors the omission. The lower query's instance cap must be included before both generation and verification.

grovedb/src/query/merge/v1.rs
```diff
@@ -0,0 +1,175 @@
...
+        if limits_in_play {
+            // Budgets never blend: a limit-carrying branch (lifted or
+            // authored) merges only as an exclusive graft. Any overlap
+            // with an existing conditional would need the two branches'
+            // bodies — and budgets — merged, which is refused by
+            // design.
+            let collides = merged_query
+                .conditional_subquery_branches
+                .as_ref()
+                .is_some_and(|branches| branches.keys().any(|item| item.contains(key.as_slice())));
```
QuantumExplorer, line 147: [P1] Treat overlap with the merged root/default query as a collision too. This predicate only inspects existing conditional branches. If an unlimited input already selects `key` and supplies a default subquery, a limited descendant below `key` passes this check; `add_conditional_subquery` then overrides the root/default behavior and silently drops one merge input. On the final head I reproduced root `Key(docs) -> Key(root-only)` plus limited `[docs] -> Key(limited)`: merge returned `Ok`, and the resulting `docs` branch contained only `limited`. Reject root item/default overlap or preserve both semantics when grafting.

grovedb/src/operations/proof/generate.rs
QuantumExplorer: Fixed (pushed to this branch after the squash-merge; carried in #847): a layer with subquery branches runs its merk walk under the global budget only — the instance chain budgets descendant rows and must not truncate child enumeration — while pure terminal layers keep min(global, instance); the verifier derives the identical per-layer value, and your a_empty/b_full repro is the regression test. This also surfaced and fixed a truncation-masked asymmetry: both sides now charge subquery-matched empty trees as empty children (global only).

grovedb/src/operations/proof/generate.rs
QuantumExplorer: Fixed in #847: the MMR arm min-composes `query_items_at_path(&lower_path)`'s `instance_limit` into the layer budget on both prover and verifier; covered by an MMR child-cap test (3 leaves, cap 1 → 1 row).

grovedb/src/operations/proof/generate.rs
QuantumExplorer: Fixed in #847: the BulkAppend arm min-composes the lower query's `instance_limit` on both sides; your 3-values/cap-1 repro is the regression test (1 row through public prove/verify).

grovedb/src/operations/proof/generate.rs
QuantumExplorer: Fixed in #847: the BulkAppend layer accounting saturates (`end.min(total_count).saturating_sub(start)`), so an empty child range charges zero instead of underflowing; regression test proves an out-of-range child query with an active cap and verifies 0 rows.

grovedb/src/operations/proof/generate.rs
QuantumExplorer: Fixed in #847: the Dense arm min-composes the lower query's `instance_limit` on both sides; covered by a Dense child-cap test (10 entries, cap 3 → 3 rows).

grovedb/src/operations/proof/generate.rs
QuantumExplorer: Fixed in #847: the CommitmentTree arm min-composes the lower query's `instance_limit` before delegating, and the delegated BulkAppend helper's fixes apply to it as well (same call path as the covered bulk test).

grovedb/src/query/merge/v1.rs
QuantumExplorer: Fixed in #847: overlap with the merged root's own selection is now a collision — assessed before the graft inserts its own item — with your root-only/limited repro as the regression test.

PR #847, opened 2026-08-31
fix: follow-ups from the per-instance limits reviews (#844/#845)

QuantumExplorer:
Follow-up PR for the review comments on #844 (merged) and #845 (merged before its final review-fix push landed — this PR carries those commits).

## P1 fixes reviewed on #845 (cherry-picked; the squash-merge raced past them)

- **Instance caps no longer truncate the parent walk.** A layer with subquery branches runs its merk walk under the **global** budget only: the instance cap counts descendant rows, and an empty first child consumes none of it, so truncating the child enumeration discarded later populated children (prove/verify returned 0 rows where the trusted read returned `b_full/k0`). Pure terminal layers keep `min(global, instance)`; the verifier derives the identical per-layer value.
- That exposed a truncation-masked prover/verifier asymmetry on subquery-matched **empty** trees (prover charged a row incl. instance; verifier charged nothing). Both sides now charge them as empty children — global budget only, mirroring the trusted read's empty-subquery charge.
- The four **non-Merk adapters** (MMR / BulkAppend / Dense / CommitmentTree) min-compose the lower query's own `Query::limit` on both prover and verifier — they bypass recursive frame creation, so a child cap of 1 verified 3 rows.
- The BulkAppend layer accounting **saturates on an empty child range** (positions at/past `total_count` clamped `end` below `start`: debug-panic underflow, ~65k rows charged in release).
- The merge lift treats **overlap with the merged root's own selection** as a collision — a grafted conditional overrides the root query's default/terminal semantics for that key and silently dropped the root-landing input's contribution.

## #844 post-merge follow-ups

- **Versioned serde representation** for `Query`, hand-written and format-aware: a fixed always-present positional layout with a leading `version` (non-self-describing formats round-trip; cross-generation payloads fail cleanly), flat base-compatible JSON for versions 1–2 (old readers keep serving everything they can serve), and the limit-bearing version 3 **nested under `body`** so a pre-limit reader hard-fails on missing fields instead of silently ignoring the unknown `limit` key and decoding an unlimited query. Unknown flat keys are refused; canonical version claims enforced. Cross-generation writer/reader tests cover both directions in both format families.
- `Element::query_item` runs the recursive whole-query preflight — an unmatched limited conditional previously made acceptance depend on database contents.
- The internal walk helpers validate their **subordinate method-version slots** (`element.query_item`, `element.get_query_apply_function`, `element.get_path_query`) which the wrapper-only checks stopped covering after the refactor; doctored-slot tests pin all three.
- Both coherence gates **exact-match the `path_query_push` selector**: an unknown engine value is a typed `UnknownVersionMismatch` up front, independent of whether any data matches.

Full workspace suite: 5156 passed / 0 failed.

## Summary by CodeRabbit

- **New Features**
  - Added version-aware query serialization for compatible JSON and binary representations across query versions.
  - Added validation for unsupported query versions and unknown fields.

- **Bug Fixes**
  - Improved query-limit enforcement across nested queries, proofs, empty results, and specialized data structures.
  - Corrected limit handling for sparse and descending queries.
  - Prevented silent data loss when limited queries overlap merged results.
  - Added clearer failures for unsupported or unknown query engine versions.

- **Tests**
  - Expanded coverage for serialization compatibility, query limits, proof behavior, merges, and version validation.

### DIFF
--- a/grovedb-query/src/query.rs
+++ b/grovedb-query/src/query.rs
@@ -936,21 +936,28 @@ mod tests {
 /// broke as fields were appended (`serde(default)` cannot turn EOF
 /// into an omitted field), and a self-describing reader from *before*
 /// a field silently ignored its unknown key — turning a bounded query
-/// into an unlimited one. This module mirrors the bincode codec's
-/// versioning instead:
+/// into an unlimited one, or a read-mode query into plain key
+/// selection. This module mirrors the bincode codec's versioning
+/// instead:
 ///
 /// - **Non-self-describing formats** (`!is_human_readable`, e.g.
-///   serde-bincode): a fixed layout with a leading `version` integer
-///   and every field always present, so the format round-trips and a
-///   reader from another generation fails at the layout instead of
-///   misreading.
+///   serde-bincode): a framed layout — a leading `MAGIC` sentinel,
+///   then a version integer, then every field always present. A
+///   plain `version: u8` alone would NOT fail closed: the released
+///   unframed layout begins with the `items` vector length, so a
+///   legacy reader would consume a small version byte as a length and
+///   reinterpret the remaining fields as query items. The magic
+///   decodes there as an absurd items length instead, which errors —
+///   and this reader requires the magic exactly, so released payloads
+///   error cleanly here too (positional layouts carry no
+///   self-description to migrate on).
 /// - **Self-describing formats** (`is_human_readable`, e.g. JSON):
-///   versions 1 and 2 keep the flat, pre-limit map layout (plus a
-///   `version` key old readers ignore), preserving interop for every
-///   query an old reader can serve. Version 3 — the limit-bearing
-///   form old readers *cannot* serve — nests its fields under a
-///   `body` key, so an old reader hard-fails on the missing flat
-///   fields instead of silently dropping the limit.
+///   version 1 — the only content the **released** derived layout can
+///   serve — keeps the flat map layout (plus a `version` key old
+///   readers ignore). Versions 2 and 3 (read-mode- and limit-bearing
+///   forms) nest their fields under a `body` key, so an old reader
+///   hard-fails on the missing flat fields instead of silently
+///   dropping the read mode or the limit.
 ///
 /// Decoding validates canonicality exactly like bincode: the version
 /// must be the lowest that can represent the contents.
@@ -988,9 +995,17 @@ mod query_serde {
         Ok(())
     }
 
-    /// Fixed positional layout — every field always present.
+    /// The positional frame sentinel. Chosen so a legacy unframed
+    /// reader, which decodes the leading bytes as its `items` vector
+    /// length, sees an absurd length and errors instead of
+    /// reinterpreting the payload.
+    const POSITIONAL_MAGIC: u64 = u64::from_be_bytes(*b"grvquery");
+
+    /// Framed positional layout — magic, version, then every field
+    /// always present.
     #[derive(Serialize)]
     struct PositionalRef<'a> {
+        magic: u64,
         version: u8,
         items: &'a Vec<QueryItem>,
         default_subquery_branch: &'a SubqueryBranch,
@@ -1004,6 +1019,7 @@ mod query_serde {
     #[derive(Deserialize)]
     #[serde(rename = "Query")]
     struct PositionalOwned {
+        magic: u64,
         version: u8,
         items: Vec<QueryItem>,
         default_subquery_branch: SubqueryBranch,
@@ -1014,7 +1030,8 @@ mod query_serde {
         limit: Option<u16>,
     }
 
-    /// The nested body of the human-readable version-3 form.
+    /// The nested body of the human-readable version-2 and version-3
+    /// forms.
     #[derive(Serialize, Deserialize)]
     struct BodyOwned {
         items: Vec<QueryItem>,
@@ -1023,7 +1040,7 @@ mod query_serde {
         left_to_right: bool,
         add_parent_tree_on_subquery: bool,
         read_mode: Option<Box<ReadMode>>,
-        limit: u16,
+        limit: Option<u16>,
     }
 
     #[derive(Serialize)]
@@ -1034,7 +1051,7 @@ mod query_serde {
         left_to_right: bool,
         add_parent_tree_on_subquery: bool,
         read_mode: &'a Option<Box<ReadMode>>,
-        limit: u16,
+        limit: &'a Option<u16>,
     }
 
     impl Serialize for Query {
@@ -1042,6 +1059,7 @@ mod query_serde {
             let version = wire_version(self.read_mode.is_some(), self.limit.is_some());
             if !serializer.is_human_readable() {
                 return PositionalRef {
+                    magic: POSITIONAL_MAGIC,
                     version,
                     items: &self.items,
                     default_subquery_branch: &self.default_subquery_branch,
@@ -1055,7 +1073,11 @@ mod query_serde {
             }
             use serde::ser::SerializeStruct;
             match version {
-                3 => {
+                2 | 3 => {
+                    // Both non-released forms nest: the released
+                    // derived reader ignores unknown flat keys, so
+                    // either feature would silently vanish in a flat
+                    // map.
                     let mut state = serializer.serialize_struct("Query", 2)?;
                     state.serialize_field("version", &version)?;
                     state.serialize_field(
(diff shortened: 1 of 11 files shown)


grovedb-query/src/query.rs
```diff
@@ -937,3 +928,316 @@ mod tests {
...
+        left_to_right: bool,
+        add_parent_tree_on_subquery: bool,
+        read_mode: &'a Option<Box<ReadMode>>,
+        limit: u16,
+    }
+
+    impl Serialize for Query {
+        fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+            let version = wire_version(self.read_mode.is_some(), self.limit.is_some());
+            if !serializer.is_human_readable() {
```
QuantumExplorer, line 1060: [P2] **Use a fail-closed positional version envelope**

Prepending `version: u8` inside an otherwise unframed positional struct does not make cross-generation decoding fail closed: a legacy reader can consume that byte as its old `items` length and reinterpret the following fields. I reproduced a fully consumed v3 payload that becomes three different legacy query items with no instance limit. Please use an outer tagged/framed representation with a non-colliding discriminator, and test each new layout against frozen released layouts with full-consumption checks.

grovedb-query/src/query.rs
```diff
@@ -937,3 +928,316 @@ mod tests {
...
+                            conditional_subquery_branches: &self.conditional_subquery_branches,
+                            left_to_right: self.left_to_right,
+                            add_parent_tree_on_subquery: self.add_parent_tree_on_subquery,
+                            read_mode: &self.read_mode,
+                            limit: self.limit.expect("version 3 iff limit is present"),
+                        },
+                    )?;
+                    state.end()
+                }
+                2 => {
```
QuantumExplorer: [P2] **Make human-readable v2 fail closed for released readers**

The flat v2 map is accepted by the released v5.0.1 derived-Serde `Query`: that reader ignores both unknown keys, `version` and `read_mode`, and constructs a plain query. This silently drops axis or sum-budget semantics. The new test fixture already contains `read_mode`, so it models only the immediately preceding generation rather than the actual released layout. Please give v2 the same incompatible nested/tagged envelope used for v3 and add an exact v5.0.1-layout regression.

grovedb/src/query/merge/v1.rs
```diff
@@ -138,13 +150,16 @@ pub(super) fn merge_v1(path_queries: Vec<&PathQuery>) -> Result<PathQuery, Error
...
-            // bodies — and budgets — merged, which is refused by
-            // design.
-            let collides = merged_query
-                .conditional_subquery_branches
-                .as_ref()
-                .is_some_and(|branches| branches.keys().any(|item| item.contains(key.as_slice())));
+            // — with an existing conditional, or with the merged root's
+            // own selection for this key — would need the two sides'
+            // bodies (and budgets) merged, which is refused by design.
+            let collides = overlaps_root_selection
```
QuantumExplorer: [P2] **Do not let unrelated limited branches change overlap acceptance**

The surrounding `limits_in_play` predicate inspects the entire partially accumulated `merged_query`, so a disjoint branch processed earlier can make a later limit-free overlap reject here. A focused test with the same three inputs showed `[root, overlap, disjoint-limited]` succeeds while `[root, disjoint-limited, overlap]` returns `NotSupported`. Please base this decision only on the current branch and the existing branch/root selection it actually collides with, and add a permutation regression.

grovedb/src/operations/proof/verify.rs
```diff
@@ -2021,8 +2040,20 @@ impl GroveDb {
...
+                                                frame_instance,
+                                                query
+                                                    .query_items_at_path(&path, grove_version)?
+                                                    .and_then(|lower_query| {
+                                                        lower_query.instance_limit
+                                                    }),
+                                            );
                                             let mut non_merk_effective =
-                                                limit_state.effective_layer_limit(frame_instance);
+                                                limit_state.effective_layer_limit(lower_instance);
```
QuantumExplorer: [P3] **Apply query direction before enforcing specialized-tree caps**

All four specialized adapters min-compose the new cap, but their result sequences are ascending and the loops stop after the first N without consulting `left_to_right`. I reproduced this with Dense positions 0..2: a descending limit-1 query returned position 0 instead of 2. MMR sorts indices ascending, Bulk iterates values ascending, and Commitment delegates to Bulk. Please order according to the query before decrementing the cap in both prover and verifier, and assert exact keys for descending MMR, Bulk, Dense, and Commitment tests.

grovedb/src/operations/proof/generate.rs
```diff
@@ -3594,9 +3676,17 @@ impl GroveDb {
...
+        // The range can be empty relative to the stored entries — e.g.
+        // a query whose positions all sit at or past `total_count`
+        // clamps `end` below `start` — in which case nothing was
+        // consumed; a plain subtraction here underflowed (panicking in
+        // debug, charging ~u16::MAX rows in release).
         if let Some(limit) = overall_limit.as_mut() {
-            let count = (end.min(total_count) - start).min(u16::MAX as u64) as u16;
+            let count = end
+                .min(total_count)
+                .saturating_sub(start)
```
QuantumExplorer: [P3] **Charge matching rows rather than the bounding span**

This count is the width of the proof range, not the number of rows selected by the query. The verifier later charges only values matching `sub_query.items`, so sparse keys desynchronize the shared budget. With stored/requested positions 0 and 100 under global limit 3, the prover exhausted its budget and omitted a later queried child; the verifier charged two rows and rejected the honest proof as missing that lower layer. Please derive the charge from actual semantic matches and share that accounting rule with verification; CommitmentTree inherits the same issue.

grovedb/src/operations/proof/verify.rs
```diff
@@ -2352,6 +2419,19 @@ impl GroveDb {
...
+                    } else if element.is_any_tree()
+                        && !element.is_non_empty_tree()
+                        && internal_query.has_subquery_or_matching_in_path_on_key(key)
+                    {
+                        // An EMPTY tree a subquery matches: an empty
+                        // child, not a result row. Mirror the prover's
+                        // charge exactly — global budget only (it
+                        // bounds walks across many empty children),
+                        // per-instance budgets count result rows.
+                        limit_state.charge_empty_layer();
```
QuantumExplorer, line 2442: [P3] **Emit empty parents when `add_parent_tree_on_subquery` requests them**

This branch only charges the empty descent and never pushes the matched parent row. In a focused proof containing one empty and one non-empty child with `add_parent_tree_on_subquery=true`, verification returned the non-empty parent plus its leaf but omitted the empty parent. That contradicts the documented parent-inclusion behavior and treats empty/non-empty matches inconsistently. Please emit the empty parent under the same ordering policy and add a mixed empty/non-empty regression.

QuantumExplorer (commented):
Six confirmed findings from an in-depth review of `84c5f03a4565345d62f1e4884a03e4693f7ee1f9`. Details and focused reproduction evidence are included inline.

grovedb-query/src/query.rs
QuantumExplorer: Fixed in af3723f5: the positional layout is framed — a leading magic sentinel (`grvquery` as u64) before the version, which the released unframed reader consumes as an absurd items-vector length and errors, closing the reinterpretation you reproduced; the framed reader requires the magic exactly. Full-consumption cross-generation tests (head↔released, all four option combinations) pin both directions.

grovedb-query/src/query.rs
QuantumExplorer: Fixed in af3723f5: version 2 nests under `body` exactly like version 3 (only version 1 — the sole content the released v5.0.1 derive can serve — stays flat), flat maps smuggling `read_mode` or `limit` are refused at head, and the test fixture is now the released five-field layout with regressions in both directions.

grovedb/src/query/merge/v1.rs
QuantumExplorer: Fixed in af3723f5: sub-level branches merge in a canonical partition — every limit-free branch first through the full merge machinery (still limit-free at that point), limited grafts last, colliding only with structures they actually touch — so acceptance is order-independent; your three-input permutation is the regression test (both orders merge and derive the identical query).

grovedb/src/operations/proof/verify.rs
QuantumExplorer: Fixed in af3723f5: the adapter verifiers order rows by the query direction before the cap applies (Bulk sorts+reverses `values_in_range`, MMR its verified leaves, Dense its entries; CT inherits Bulk's), so a descending cap keeps the last positions; exact-key descending limit-1 tests added for all four, including a CommitmentTree fixture.

grovedb/src/operations/proof/generate.rs
QuantumExplorer: Fixed in af3723f5: the BulkAppend prover now charges `expand_query_to_u64_positions(...).len()` — the same semantic-match set the verifier charges from — instead of the bounding span; your positions-{0,100}-under-global-3 scenario with a later sibling layer is the regression test (honest proof verifies, 3 rows).

grovedb/src/operations/proof/verify.rs
QuantumExplorer: Fixed in af3723f5: the empty-child verifier arm reports the matched parent when `should_add_parent_tree_at_path` holds, before its (uncharged, per the documented flag limitation) empty-child charge — empty and non-empty parents now behave alike, pinned by a mixed empty/non-empty regression.

PR #850, opened 2026-09-04
feat: merge a limited path query below a key another branch already owns

QuantumExplorer:

`PathQuery::merge` (v1, `GROVE_V4`+) grafts a limited input exclusively at the **first key past the common path** and refuses any collision there. That refuses compositions whose inputs merely share that key and diverge right below it:

- a limited page on `post` (an index range, `[…, post, hashtag]`) merged with an unlimited by-id fetch on `post` (`[…, post, 0]`): both graft at key `post`;
- once a third input in another contract lifts the common path up to the root, a limited page and a limited lookup under the same contract graft at the contract-id key.

Budgets never blend in these shapes; the merge just stopped one level too early. Platform's composite document queries (a page plus derived joins / lookups / counts under one proof) hit both cases immediately.

### The collision, before this PR

Two inputs under contract `C`. The by-id fetch has no limit and goes through the full merge machinery first, so it owns key `post` as a conditional branch. The page carries a limit, so it may only graft on an exclusive key. Its first key past the common path is `post` again, and the merge refused it there even though the two only share that one key.

```mermaid
flowchart TD
    subgraph inputs["inputs, common path = [C]"]
        A["A  page<br/>[C, post, hashtag]<br/>limit 20"]
        B["B  by-id fetch<br/>[C, post, 0]<br/>no limit"]
    end
    R(["merged root at [C]"])
    B -- "limit-free: merged first" --> R
    R -- "key post (conditional branch)" --> P["post"]
    P -- "0" --> B2["B's body: the ids"]
    A -. "limited: needs an exclusive key,<br/>but post is already owned" .-> P
    P -. "before: NotSupported" .-> X["refused"]
    style X fill:#fdd,stroke:#c33,color:#600
```

## What was done?

A limited branch colliding on a key that an earlier graft already owns now **descends** into it (`graft_below`): both branches are rebuilt as path queries rooted at `common_path + [key]` and merged with the very same rules, so they either land on exclusive keys of their own or are refused exactly as before (a body meeting the descended root, or a key selected by a range or by the merged root's own items, where nothing below can be told apart). Lifted limits already sit on the leaf bodies as per-instance caps, so the descent lifts nothing twice, and the existing partition (limit-free grafts first, limited grafts last) keeps acceptance independent of input order.

"Owns" follows execution: a key takes the first conditional branch whose selector contains it, in insertion order. So the descent happens only when the merged root selects the key exactly, the first conditional matching it is that exact key, and no range selects it. A conditional shadowed by an earlier range, or one defined for a key the root never selects, would never run, and grafting below it would silently drop the incoming rows, so those shapes refuse instead.

### The descent

The owning branch and the incoming branch become two path queries rooted one level down, at `[C, post]`, and go through `merge_v1` again. There they diverge onto `hashtag` and `0`, so each lands on an exclusive key of its own. The merged branch replaces the one that owned `post`.

```mermaid
flowchart TD
    R(["merged root at [C]"])
    R -- "key post" --> G["graft_below<br/>= merge_v1 rooted at [C, post]"]
    G -- "hashtag (exclusive)" --> A2["A's body<br/>per-instance cap 20<br/>(the lifted limit)"]
    G -- "0 (exclusive)" --> B2["B's body<br/>the ids, no cap"]
    style G fill:#e8f4ff,stroke:#369
```

Because the descent is the same merge, it recurses on its own: if the two still collide on the next key, that key's owner descends again, until the branches diverge or a refusal rule fires.

### Where each limited input goes now

```mermaid
flowchart TD
    S["limited input<br/>(global limit or branch caps)"] --> Q1{"path == common path?<br/>(lands at the merged root)"}
    Q1 -- yes --> Q2{"global limit, or a cap<br/>on the root body itself?"}
    Q2 -- yes --> X1["refused:<br/>budget would blend at the root"]
    Q2 -- "no (caps only on its branches)" --> Q3{"other bodies at the root too?"}
    Q3 -- yes --> X2["refused:<br/>a branch cap would govern<br/>rows the other bodies select"]
    Q3 -- no --> OK1["accepted: it IS the root body,<br/>caps ride along"]
    Q1 -- no --> K["first key K past the common path"]
    K --> Q4{"root selects K exactly,<br/>the first conditional matching K<br/>is that exact key,<br/>and no range selects K?"}
    Q4 -- yes --> D["graft_below:<br/>rebuild both at common + [K],<br/>merge_v1 again"]
    D --> Q5{"branches diverge below K?"}
    Q5 -- yes --> OK2["accepted: exclusive keys<br/>one level down"]
    Q5 -- no --> X3["refused:<br/>bodies would meet"]
    Q4 -- no --> Q6{"K selected by the root,<br/>inside a range, or matched<br/>by any conditional?"}
    Q6 -- yes --> X4["refused:<br/>nothing below can be told apart"]
    Q6 -- no --> OK3["accepted: exclusive graft at K<br/>(unchanged from before)"]
    style X1 fill:#fdd,stroke:#c33,color:#600
    style X2 fill:#fdd,stroke:#c33,color:#600
    style X3 fill:#fdd,stroke:#c33,color:#600
    style X4 fill:#fdd,stroke:#c33,color:#600
    style OK1 fill:#dfd,stroke:#393,color:#040
    style OK2 fill:#dfd,stroke:#393,color:#040
    style OK3 fill:#dfd,stroke:#393,color:#040
```

The root-landing rule is the one refinement outside `graft_below`: it used to refuse any root-lander carrying any limit. A root-lander whose caps sit on its own branches is now accepted when it is the sole body at the root. That is exactly how a branch merged one level down comes back through `graft_below`: the rebuilt path query for the owning branch lands at the descended root carrying the caps its leaves already had. Once the root body carries caps, a limit-free branch also takes the exclusive-graft path rather than the full merge machinery (whose instance-limit gate would otherwise refuse it), so an unlimited sibling beside a capped root body merges.

### The multi-contract case (a composite proof)

Three inputs: a limited page under contract `C1`, a limit-free per-post count under `C1`, and a limit-free profile lookup under contract `C2`. The common path is the tree root, so `C1` is the first key for two of them. The two limit-free inputs graft first and own `C1` and `C2`; the page then descends into `C1`, where `post` and `like` diverge.

```mermaid
flowchart TD
    R(["merged root"])
    R -- "C1" --> C1["graft_below at [C1]"]
    R -- "C2" --> C2["C2"]
    C1 -- "post" --> P["post"]
    P -- "hashtag" --> PH["page body<br/>cap 20"]
    C1 -- "like" --> L["like"]
    L -- "byPost" --> LC["count trees<br/>no cap"]
    C2 -- "profile" --> PR["profile"]
    PR -- "owner" --> PO["profile body<br/>no cap"]
    style C1 fill:#e8f4ff,stroke:#369
```

### Still refused

Same shapes, same answers as before, now reached one level down:

- two limited bodies that never diverge (both end at the same leaf: the descended merge finds two root-landers with caps);
- a limited body meeting a limit-free one at the same leaf (a cap would govern rows the other body selects);
- a key selected by a range rather than an exact key (rows under it cannot be attributed to one branch);
- a key whose first matching conditional is an earlier range (the exact-key branch is shadowed and never runs);
- a key the root defines a conditional for but never selects (a dormant branch; descending would drop the incoming rows, selecting it would add rows the root never asked for).

No wire or version change: the merge slot is unchanged, the verifier re-runs the same merge, and every previously accepted merge produces the identical query.

## How Has This Been Tested?

New tests in `per_instance_limit_tests`:

- `merge_grafts_a_limited_branch_below_a_key_a_limit_free_branch_owns`: proves, verifies and matches the trusted read; input-order independent
- `merge_grafts_two_limited_branches_that_diverge_below_a_shared_key`: same, with both branches limited
- `merge_still_refuses_limited_branches_that_never_diverge`: two limited bodies at one leaf, a limited body meeting a limit-free one, and a range-selected key all still refuse
- `merge_refuses_a_limited_graft_shadowed_by_a_range_conditional` and `merge_refuses_a_limited_graft_into_an_unselected_conditional`: the two ownership refusals above, both input orders
- `merge_grafts_below_an_exact_conditional_before_a_matching_range`: an exact conditional ahead of a matching range does own the key; proves and verifies
- `merge_grafts_an_unlimited_sibling_beside_a_root_branch_cap`: both directions, conditional and default subquery; proves and verifies
- `merge_nested_limited_grafts_is_independent_of_all_input_permutations`: a two-level shared prefix plus an unrelated root tree, every input permutation; identical merged query, identical read to the concatenated trusted reads, proves and verifies

Existing merge / per-instance-limit / merge-versioning suites unchanged and green; `cargo clippy -p grovedb --all-features -- -D warnings` clean; verify-only build OK.

## Summary by CodeRabbit

- **Improvements**
  - Query merging now supports limited branches that share an exact key but diverge beneath it.
  - Each limited branch preserves its own result limit during merging.
  - Merging remains rejected when branches do not diverge or when the shared key is selected by a range.

- **Tests**
  - Added coverage for compatible limited-branch merges and cases that must continue to be rejected.

### DIFF
--- a/grovedb/src/query/merge/mod.rs
+++ b/grovedb/src/query/merge/mod.rs
@@ -20,10 +20,12 @@
 //!   becomes its merged branch's per-instance cap (`Query::limit`) —
 //!   exact, because the branch instance executes exactly once — and
 //!   authored per-instance limits ride along on their branches.
-//!   Budgets never blend, so limits merge only as exclusive grafts: a
-//!   limited input landing at the merged root, or two limited branches
-//!   colliding on a key, are refused with typed errors. Offsets are
-//!   still refused.
+//!   Budgets never blend, so limits merge only as exclusive grafts.
+//!   Branches may descend through a selected key whose first matching
+//!   conditional is that exact key, until their paths diverge. A budget
+//!   on a body landing at the merged root, or overlapping selections
+//!   that cannot keep their budgets separate, are refused with typed
+//!   errors. Offsets are still refused.
 //!
 //!   (An intermediate carrying only the direction rules was once gated
 //!   here for `GROVE_V4`; since no grove version ever shipped it, it
--- a/grovedb/src/query/merge/v1.rs
+++ b/grovedb/src/query/merge/v1.rs
@@ -14,10 +14,18 @@
 //!
 //! Budgets never blend, so limits merge only as **exclusive grafts**:
 //!
-//! - a limited input whose whole path is the common path lands *at the
-//!   merged root*, where its query body would merge with the other
-//!   inputs' — refused;
-//! - two limited branches colliding on a key are refused;
+//! - an input whose whole path is the common path lands *at the merged
+//!   root*, where its query body merges with the other inputs': a
+//!   budget at that body (a global limit, or a cap on the body itself)
+//!   is refused, and caps on its own branches are accepted only when
+//!   it is the sole body at the root (so nothing else selects rows
+//!   under those branches);
+//! - a limited branch colliding on a key already owned by another
+//!   grafted branch **descends** into it and grafts where the two
+//!   actually diverge (the same merge, one level down — see
+//!   [`graft_below`]); it is refused only when the bodies would meet,
+//!   or when the key is selected by a range or by the merged root's own
+//!   items, where nothing below can be told apart;
 //! - limit-free inputs merge exactly as v0 does, modulo the direction
 //!   rules above, on the same code path.
 //!
@@ -61,16 +69,26 @@ pub(super) fn merge_v1(path_queries: Vec<&PathQuery>) -> Result<PathQuery, Error
                 "can not merge pathqueries with offsets".to_string(),
             ));
         }
-        let carries_limits = path_query.query.limit.is_some() || path_query.has_instance_limits();
         path_query
             .to_subquery_branch_with_offset_start_index(next_index)
             .and_then(|mut unsized_path_query| {
                 if unsized_path_query.subquery_path.is_none() {
                     // The input lands at the merged root, where its
                     // query body merges with the other root-level
-                    // inputs — budgets cannot blend, so limits are
-                    // refused here.
-                    if carries_limits {
+                    // inputs'. A budget AT that body — a global limit,
+                    // which has nowhere to lift, or a cap on the body
+                    // itself — would have to blend, so it is refused
+                    // here. Caps on the body's own branches are
+                    // exclusive to those branches and ride along, as
+                    // long as nothing else lands at the root beside it
+                    // (checked below, once every input is placed).
+                    let root_body =
+                        *unsized_path_query
+                            .subquery
+                            .ok_or(Error::CorruptedCodeExecution(
+                                "subquery must exist when subquery_path is none in merge",
+                            ))?;
+                    if path_query.query.limit.is_some() || root_body.limit.is_some() {
                         return Err(Error::NotSupported(
                             "can not merge a limited path query that lands at the merged root: \
                              its budget would have to blend with the other queries' result \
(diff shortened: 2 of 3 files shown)


QuantumExplorer (commented):
Reviewed

PR #919, opened 2026-09-06
fix(proofs): bind every tree and reference row of trunk/branch chunk proofs (#859)

QuantumExplorer:
Closes #859.

## Problem

Trunk and branch chunk proofs (`prove_trunk_chunk` / `prove_branch_chunk`) carried subtree rows as bare `KVValueHash` / `KVValueHashFeatureType` nodes, whose `value_hash` the merk verifier treats as opaque. The shared element extraction (`extract_elements_and_leaf_keys`) then **waived** the `H(value) == value_hash` check for any row that deserialized as a tree. Nothing ever recomputed the tree's `combine_hash(H(value), child_root)` composition, so the returned tree metadata — tree type, aggregate count or sum, root key — was unbound.

Reproduced at runtime on current `develop`, on both variants:

- a subtree row's bytes replaced by `CountTree(Some("phantom"), 1_000_000)` verifies against the genuine root hash and is returned as verified (trunk and branch);
- an **item** row can be disguised as a tree: `KV(key, item)` → `KVValueHash(key, forged_tree, H(item))` verifies and returns the forged tree;
- as a side effect of the same design, an honest chunk containing a **reference** row never verified at all (references are not waived).

The issue's audit-only claim is therefore real.

## Fix (GROVE_V4, `proof.chunk_proof_row_binding: 1`)

One new version slot gates prover and verifier together, dispatched through a new `operations/proof/chunk_proof_row_binding/{mod,v0,v1}.rs` module (no inline version branches):

- **Prover** (`bind_chunk_proof_rows`): every composite row of the chunk ops is rewritten to `KVValueHashFeatureTypeWithChildHash` via a per-row binder now **shared with the sum-budget window from #870** (`bind_composite_row`, extracted from `bind_sum_budget_window_rows` with no behaviour change): the child Merk root (`NULL_HASH` when empty), the non-Merk tree's own state root (through `bind_terminal_non_merk_tree`), or the referenced element's value hash. Hooked into `prove_trunk_chunk_non_serialized_v1` and `prove_branch_chunk_non_serialized`; a no-op below V4.
- **Verifier** (`check_chunk_proof_row`): replaces the inline waiver. v0 is the released logic moved verbatim; v1 requires the child-hash node for every tree or reference row, checks the two-input composition, and rejects trees/references in any other node form.
[...]

PR #921, opened 2026-09-06
fix(proofs): pin chunk row binding to one snapshot

QuantumExplorer:
Follow-up to #919.

Trunk and branch generation could return a proof that fails verification when an atomic commit lands between reading the parent chunk and binding its composite rows. The parent commitment came from the old state while the child root or referenced value came from the new state. Trunk ancestor proofs could also read a different state.

Start one snapshot read transaction directly in the V1 trunk and branch generation entry points, covering the entire operation. Snapshot selection needs no protocol gate: at a fixed committed state it changes neither proof bytes nor tracked operation costs, and ordinary query-proof generation already follows this pattern. The V0 trunk entry point is unchanged.

Regression tests force an atomic commit after chunk ops are collected and before row binding. They cover tree and reference rows in both trunk and branch proofs, check the pre-commit root, and require byte-for-byte equality with the pre-commit proof. Both tests failed before the fix and pass afterward.

Validation:
- 54 focused tests passed: row binding (19), trunk proofs (30), branch proofs (3), snapshot transactions (2).
- Verifier-only build check passed: `cargo check --offline --locked -p grovedb --no-default-features --features verify`.
- Formatting, diff checks, and commit hooks passed. Existing warnings remain in untouched files.


QuantumExplorer (commented):
Reviewed

PR #930, opened 2026-09-07
fix: validate indexed secondary roots against the stored element (#897)

### LINKED ISSUE #897: [audit][M019] Indexed metadata overwrite does not validate canonical secondary-root identity
Audit group: **M019**. Classification: **correctness**. Provisional severity: **medium**.

Direct indexed-tree metadata overwrite validates the primary against canonical state, but secondary validation opens the root using incoming metadata and compares the returned key with that same input. This does not establish that the incoming secondary root is the canonical stored root. The database can accept a secondary commitment inconsistent with the unchanged complete primary. Primary aggregate/type/schema guards remain effective; this is a state-mutation integrity problem rather than acceptance of an unchanged trusted proof root.

Expected behavior: Incoming secondary-root metadata must be compared with canonical stored authority before accepting an indexed-state overwrite.

Fix direction: Read canonical secondary root keys from the stored parent element and compare incoming keys and complete axis/root mappings before opening secondaries. Derive the accepted joint commitment from canonical state. Preserve historical version rules.

### Validation to complete

- Suggested: verify secondary root/axis identity against stored canonical metadata before any overwrite is accepted.
- Check complete primary-secondary relations and version-specific indexed variants after allowed updates.

Related tracking: [PR #657](https://github.com/dashpay/grovedb/pull/657) (merged), [PR #773](https://github.com/dashpay/grovedb/pull/773) (merged).

Scope: saved GroveDB worktree with revision context `2fa0f133877420a0d9c91ba7bc51b1775ab8c783`. This report does not establish that current `develop` or any deployed application is affected. Focused runtime validation remains outstanding.

<details>
<summary>Audit source and canonical finding identifiers</summary>

Source status: snapshot-backed (`git_worktree`); plain source locations are used because this is not a sealed commit-only scan.

Audited revision context: `2fa0f133877420a0d9c91ba7bc51b1775ab8c783`.

The findings were manually reconciled from a preserved scan bundle. The native scan ended before final completion; these are provisional source-review findings, not a completed native scan certification.

Canonical finding ID: `csf_f350f0dc8a4590b58cc683ec`

Primary fingerprint: `codex-security/v1:sha256:cda5ea4f9cbfc2a6b7e5b6b18472c439791de652118c85067764a8e74127934a`

Source locations:

- Location (entrypoint): `grovedb/src/operations/insert/mod.rs:20-68`
- Location (root_control): `grovedb/src/operations/insert/add_element_on_transaction/v1.rs:310-357`
- Location (root_control): `grovedb/src/operations/insert/add_element_on_transaction/v1.rs:440-480`
- Location (root_control): `grovedb/src/operations/insert/add_element_on_transaction/v1.rs:592-687`
- Location (root_control): `grovedb/src/operations/insert/add_element_on_transaction/v0.rs:331-359`
- Location (root_control): `grovedb/src/operations/insert/add_element_on_transaction/v0.rs:456-486`
- Location (root_control): `grovedb/src/operations/insert/add_element_on_transaction/v0.rs:648-682`
- Location (root_control): `grovedb/src/operations/indexed_tree.rs:317-353`
- Location (root_control): `merk/src/merk/open.rs:70-95`

</details>

QuantumExplorer:
Fixes #897 (audit M019).

## The vulnerability — confirmed real

Direct insertion of a non-empty indexed-tree element (`ProvableCountIndexedTree`, `ProvableSumIndexedTree`, `ProvableCountProvableSumIndexedTree`) validated the claimed secondary root key by opening the secondary Merk **with that key** and comparing the returned root key against the same input — circular. Every row of a secondary Merk is a node key in its storage, so `open_layered_with_root_key` succeeds for any of them and hands the claimed key back.

The primary was never affected: `open_transactional_merk_at_path` reads the root key from the **stored** parent element, so the primary comparison is against canonical state. But a caller supplying the canonical primary root key and aggregates plus a **non-root node key** of the secondary Merk as the secondary root passed every check and committed the hash of an interior/leaf node — authenticating a strict subtree of the index as the whole index. Rows outside that subtree vanish from the authenticated index (wrong top-k/rank results, false absence), while the untouched primary keeps every check green.

Reproduced on current `develop` for all three variants: the new regression tests, run before the fix, saw the forged insert return `Ok(())`.

## The fix

In the **v2 arm** of `add_element_on_transaction` (selected by `GROVE_V4`'s `add_element_on_transaction: 2`, so the version dispatch itself is the gate — same pattern as #922/#923):

- The claimed secondary root key must equal the **stored element's** value; the stored bytes are hash-committed by the parent Merk, so they are the canonical authority. Only after that equality is established are the secondaries opened (with what is now the canonical key) to derive the committed hashes.
- PCPSIT compares the complete `(tag, root_key)` axes mapping after the existing tags-only schema check.
- The no-stored-element / stored-non-tree case (unreachable — the primary open errors first) fails closed instead of open.
- The in-place conversion guards are preserved with identical messages; the stored element is fetched once for both checks, so tracked costs are unchanged.

v0/v1 keep the historical behaviour (`GROVE_V3` is live) and carry KNOWN-FLAW header notes pointing here. The batch path is unaffected: it refuses non-empty indexed overwrites outright.

## Tests

- `direct_insert_{pcit,psit}_existing_nonroot_secondary_node_key_rejected` and `direct_insert_pcpsit_existing_nonroot_axis_node_key_rejected`: forge with an **existing non-root row key** (`sort_key ‖ item_key`) — the shape the old code accepted (the pre-existing mismatch tests used nonexistent keys, which even the old code rejected because the open fails).
- `indexed_tree_direct_insert_guards_v1` now pins the frozen `GROVE_V3` arm explicitly; new `indexed_tree_direct_insert_guards_v2` pins the canonical-authority messages at latest.
- `verify_grovedb_psit_detects_coherently_rebound_relational_drift` staged its corrupt state through exactly this vector; it now rebinds through the `GROVE_V3` arm, simulating a pre-V4-written state, and still pins the relational verifier's detection.

Full `grovedb` lib suite green (3308 tests); clippy `--all-targets` clean on the touched crates.

### DIFF
--- a/grovedb-version/src/version/v4.rs
+++ b/grovedb-version/src/version/v4.rs
@@ -235,8 +235,15 @@
 //!   a tree element, which the batch resolver has always rejected; v1
 //!   accepted it and committed only `H(tree element bytes)`, a hash that does
 //!   not bind the subtree's contents, so the row could not be proved and
-//!   subtree changes never disturbed it. Gated because (i) moves a committed
-//!   root and (ii)/(iii) flip an accepted/rejected outcome.
+//!   subtree changes never disturbed it. (iv) A directly inserted non-empty
+//!   indexed-tree element (PCIT / PSIT / PCPSIT) has its claimed secondary
+//!   root keys validated against the STORED element's canonical values
+//!   (issue #897). v1 opened each secondary Merk with the incoming key and
+//!   compared the returned root key against that same input — circular, so
+//!   any existing node key of the secondary (every row is one) passed and
+//!   committed the hash of a non-root node, authenticating a strict subtree
+//!   of the index as the whole index. Gated because (i) moves a committed
+//!   root and (ii)/(iii)/(iv) flip an accepted/rejected outcome.
 //!
 //! - `storage_costs.add_basic_storage_removal_to_sectioned_storage_removal:
 //!   1` — combining a `BasicStorageRemoval` with a `SectionedStorageRemoval`
--- a/grovedb/src/operations/insert/add_element_on_transaction/mod.rs
+++ b/grovedb/src/operations/insert/add_element_on_transaction/mod.rs
@@ -28,7 +28,11 @@
 //!   later read failed on. The batch path already refuses it. v2 further
 //!   refuses a reference whose terminal is a tree, as the batch resolver
 //!   always has; v0/v1 accepted it and committed a hash that does not bind
-//!   the subtree's contents. Selected by `GROVE_V4`+.
+//!   the subtree's contents. v2 also validates a non-empty indexed-tree
+//!   element's claimed secondary root keys against the STORED element
+//!   (issue #897); v0/v1 opened each secondary with the incoming key and
+//!   compared the result against that same input, which any existing node
+//!   key of the secondary Merk passes. Selected by `GROVE_V4`+.
 //!
 //! The implementations are otherwise identical. See [v0] / [v1] / [v2].
 //!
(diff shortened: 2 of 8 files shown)


QuantumExplorer (commented):
Approved

PR #932, opened 2026-09-07
refactor(query): unify validation across reads and proofs

QuantumExplorer:

Query entry points independently rediscover query shapes and apply version/envelope gates, which can make trusted reads, proof generation, and verification disagree. Share the existing `PathQuery::classify()` decision while preserving historical contracts.

## What was done?

- Add internal, borrowed `ValidatedPathQuery` carrying the query, validated shape, aggregate kind/range, and applicable version/envelope checks.
- Use that decision in unified reads, proof generation, and verification; carry it through the V1 prover's recursive walks instead of rediscovering aggregate and pagination shapes.
- Preserve V0 error precedence, per-instance limit gates, snapshot handling, tree-dependent checks, and operation-cost ordering.
- Add regressions for aggregate leaves/carriers, conflicting shapes, descending merged/subset queries, and absence results. Pin 22 query/proof byte and cost fixtures across all four shipped versions.
- Document incompatible legacy result contracts in `docs/query-validation.md`, including ordinary offsets, scalar versus per-key aggregates, absence ordering, and subset pagination.

## How Has This Been Tested?

- `cargo test -p grovedb --lib --features verify --offline --locked`: 3,306 passed, 0 failed, 8 ignored on current `develop`.
- `cargo test -p grovedb --all-features --locked validated_path_query_tests`: 8 passed, including all 22 compatibility fixtures and regressions for unknown proof versions and legacy envelopes.
- `cargo check -p grovedb --lib --no-default-features --features verify --offline --locked`: passed; existing unused-import warnings remain.
- Pre-commit checks, including formatting, whitespace, conflict markers, and typos: passed.

## Breaking Changes

None. Public APIs, query serialization, proof bytes, historical version behavior, and operation costs are preserved. Legacy result-specific APIs retain their existing contracts.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added consistent query validation across reading, proof generation, and proof verification.
  * Improved support for aggregate queries, including count, sum, combined aggregates, and empty results.
  * Added documentation covering query validation, supported proof envelopes, and existing API compatibility.

* **Bug Fixes**
  * Improved consistency when handling conflicting, unsupported, or version-incompatible query shapes.

* **Tests**
  * Added coverage for cross-operation consistency, pagination compatibility, serialized outputs, and operation costs.


grovedb/src/query/validated.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    pub(crate) fn shape(&self) -> PathQueryShape<'q> {
+        self.shape
+    }
+
+    /// The aggregate terminal's depth is fixed by the leaf/carrier grammar.
+    /// The prover still opens and checks the actual tree at this depth.
+    #[cfg(feature = "minimal")]
+    pub(crate) fn aggregate_at_depth(
```
QuantumExplorer, line 237: **[altitude] The aggregate short-circuit is now depth arithmetic with no fail-closed backstop.**

The V1 prover fires the aggregate terminal on `depth == path.len() + 1 + subquery_path.len()`, a re-derivation of the carrier grammar that lives in `grovedb-query`'s `validate_carrier_*` (exactly one fan-out level, default branch only, no conditionals). Today the formula is provably equivalent to the old per-layer `query.items.iter().any(is_aggregate_*)` trigger — I verified this.

The risk is future drift: `shape.rs` explicitly contemplates loosening the grammar later ("loosening a grammar is additive"). If a new carrier form is admitted without updating this arithmetic, `aggregate_at_depth` returns `None` at the real terminal and the prover silently emits an ordinary element proof over the aggregate's inner range (`AggregateCountOnRange` delegates its bounds to the inner item), a wrong-proof-kind failure caught only at verification.

Suggestion: compute the terminal depth during classification and store it on `PathQueryShape::AggregateCarrier { terminal_depth, .. }` so the grammar and the depth can't diverge — or keep an item-based `debug_assert!`/hard error in `prove_subqueries_v1` for a layer whose effective items contain an aggregate the depth check missed.

grovedb/src/query/validated.rs
```diff
@@ -0,0 +1,255 @@
...
+                })
+            }
+            PathQueryShape::SumBudget { .. } if versions.sum_budget_in_v1_envelope != 1 => {
+                Some(if generating {
+                    "sum-budget windows in the V1 proof envelope are not emitted at this grove version"
+                } else {
+                    "sum-budget windows in the V1 proof envelope are not accepted at this grove version"
+                })
+            }
+            _ => None,
```
QuantumExplorer: **[altitude] Per-shape policy tables fail open for future shapes.**

`check_proof_capabilities` ends in a wildcard `_ => None`, and `for_proof`'s V0 message table falls through `_ => shape.aggregate_kind()`. A future `PathQueryShape` variant added to `classify()` compiles cleanly here and ships served/emitted at **every** grove version with no capability gate — exactly the class of omission the `unified_read_mode` / `axis_descent_in_v1_envelope` slots exist to prevent.

Suggestion: make these matches exhaustive over `PathQueryShape` (no wildcard), as `check_envelope` already is, so adding a shape is a compile error at each policy point.

grovedb/src/query/validated.rs
```diff
@@ -0,0 +1,255 @@
...
+        };
+        match message {
+            Some(message) => Err(Error::NotSupported(message.to_string())),
+            None => Ok(()),
+        }
+    }
+
+    /// Called after decoding; actual layer/tree eligibility is still the
+    /// verifier's responsibility. Keep the existing per-family error types.
+    pub(crate) fn check_envelope(&self, proof: &GroveDBProof) -> Result<(), Error> {
```
QuantumExplorer: **[simplification] `check_envelope`: three of five arms are dead, and `Ok` does not imply a V1 envelope.**

This function is only called from the SumBudget verify arm and `verify_axis_shape_walk`, so the `KeySelection`, `CountOffsetPaginated`, and `AggregateLeaf/Carrier` arms never run. The aggregate arm duplicates `aggregate_common::require_v1_envelope` and the count-offset arm duplicates `verify.rs`'s envelope gate string-for-string, with nothing enforcing parity as those live gates evolve.

More importantly, the `KeySelection` arm returns `Ok(())` for a **V0** proof, yet both call sites follow `check_envelope` with `let GroveDBProof::V1(..) = decoded else { unreachable!() }` — the "Ok ⇒ V1" invariant lives in the call sites' dispatch context, not this function's contract. Any future reuse from a context whose shape can be `KeySelection` turns a malformed attacker-supplied proof into a verifier panic.

Suggestion: replace with `fn require_v1_envelope(&self, decoded: GroveDBProof) -> Result<GroveDBProofV1, Error>` carrying only the axis/sum-budget messages — it deletes both `unreachable!()`s and the dead arms in one move.

QuantumExplorer (commented):
This is Claude. I ran a deep review of this PR (line-by-line, removed-behavior audit against `develop`, cross-file caller trace, plus reuse/simplification/efficiency passes).

**No correctness bugs found.** Error types, messages, precedence, V0-frozen behavior, proof bytes, and operation costs were all verified as preserved — structurally against the removed entry gates and empirically via the PR's own 22 cross-version fixtures. Verify-only feature builds compile, and all callers route through `for_proof`.

The six inline comments below are robustness/cleanup findings, ranked. The first two are the ones I'd address (here or in a fast follow-up): both concern seams that fail *open* when a future `PathQueryShape` is added, which cuts against the repo's fail-closed versioning discipline.

grovedb/src/query/validated.rs
```diff
@@ -0,0 +1,255 @@
...
+        query.reject_unserved_per_instance_limits(grove_version)?;
+        let prove_version = grove_version
+            .grovedb_versions
+            .operations
+            .proof
+            .prove_query_non_serialized;
+        let shape = query.classify_with_offset_gate(|| match prove_version {
+            // V0 rejected offsets before inspecting pagination syntax.
+            0 => {
+                query.reject_per_instance_limits("the V0 prover")?;
```
QuantumExplorer, line 100: **[simplification] `for_proof` spreads the prove-version decision across five checks, two of them unreachable.**

`prove_version == 0` is tested in three places (this closure, the message table, the trailing check) and `> 1` in two; `reject_per_instance_limits("the V0 prover")` is called twice. Both of those calls are unreachable for every shipped version table: at `prove_version == 0` (grove v1/v2), `per_instance_query_limits` is 0, so `reject_unserved_per_instance_limits` at the top already rejected any instance-limit-bearing query.

Every future edit must re-derive which of the five checks fires first to avoid breaking the pinned fixtures, and touching one copy silently desynchronizes offset-bearing from offset-free queries. A single `match prove_version { 0 => <V0 table>, 1 => {}, v => Err(unknown_prove_version(v)) }` after classification preserves observable behavior for every real version — keeping exactly one `> 1` rejection, since it's what makes `generate.rs`'s `unreachable!("validated proof version")` sound.

grovedb/src/query/validated.rs
```diff
@@ -0,0 +1,255 @@
...
+            }
+            1 => Ok(()),
+            received => Err(unknown_prove_version(received)),
+        })?;
+        let validated = Self {
+            query,
+            shape,
+            grove_version,
+        };
+        if prove_version == 0 {
```
QuantumExplorer, line 113: **[reuse] The aggregate family-name → "require V1 proof envelopes" mapping now lives in three places.**

This longhand table, `check_envelope`'s `AggregateKind` match below, and the `query_type_name` literals passed to `require_v1_envelope` in `aggregate_common.rs` all spell the same three family names and near-identical messages. Renaming a family or rewording the requirement must hit all three or generation and verification emit inconsistent errors.

Suggestion: a single `AggregateKind::proof_family_name() -> &'static str` (natural home next to `aggregate_kind()` in `shape.rs`) used by all sites.

grovedb/src/operations/proof/generate.rs
```diff
@@ -2252,7 +2019,12 @@ impl GroveDb {
         // top entry already ran, so a mismatched tree type is a
         // hard-error case (the caller asked for count-offset pagination
         // against something that isn't a count tree).
-        if path.len() == path_query.path.len() && path_query.has_non_zero_offset() {
+        if path.len() == path_query.path.len()
+            && matches!(
+                validated.shape(),
+                PathQueryShape::CountOffsetPaginated { .. }
```
QuantumExplorer: **[efficiency] The dispatch discards the shape's validated `inner`, so `prove_count_offset_layer_v1` re-validates and clones.**

This match deliberately uses `CountOffsetPaginated { .. }`, and the callee still runs `path_query.validate_count_offset_paginated().cloned()` (generate.rs:3383) — a redundant validation pass (three recursive `has_aggregate_*_anywhere` walks) plus a `QueryItem` clone per count-offset proof. The double validation predates this PR, but the PR applied exactly the right fix to the aggregate arm (borrow `inner` from the shape) and left this parallel case.

Suggestion: pass the shape's `inner` (or `validated`) into `prove_count_offset_layer_v1` and delete the revalidate-and-clone.

grovedb/src/query/validated.rs
QuantumExplorer: Agreed that this is a future grammar-drift risk. The current carrier validator fixes exactly one fan-out level followed by the optional subquery path, so the arithmetic is correct for the shapes accepted today.

I would address this with terminal metadata produced alongside internal classification, or a production error guard if an aggregate terminal reaches the ordinary proof path. Adding a required field to the public `PathQueryShape` would conflict with this PR's API-compatibility constraint, and a `debug_assert!` alone would not protect release builds. This is worth addressing before merging.

grovedb/src/query/validated.rs
QuantumExplorer: Agreed. These policy matches should enumerate every shape explicitly, including the read-side eligibility check, so a new variant requires a deliberate version-policy decision.

One qualification: adding a variant would already require changes to the exhaustive consumer dispatchers and `check_envelope`, so it would not compile across the entire crate unnoticed. The gap is that these particular policy sites could remain unchanged and silently allow the new shape. Exhaustive matches close that gap and fit the purpose of this refactor.

grovedb/src/query/validated.rs
QuantumExplorer: Agreed. The current callers restrict the shape to axis or sum-budget, so this does not establish a currently reachable verifier panic. However, returning `GroveDBProofV1` would express the invariant directly and remove both `unreachable!()` calls.

A helper scoped to these V1-only verification paths would also remove the unused selection, pagination, and aggregate arms. Their live checks remain in the dedicated verifiers, where their existing error types and ordering must be preserved.

grovedb/src/query/validated.rs
QuantumExplorer: I agree that the ordering deserves care, but a single version match after classification would change historical behavior, including for shipped versions.

For example, V0 proving with a nonzero offset over explicit keys currently returns `InvalidQuery("proved path queries can not have offsets")`. Running ordinary classification first instead returns a pagination-shape error. `legacy_pagination_contracts_remain_distinct` explicitly pins this distinction. The unknown-version precedence is also covered by `unknown_proof_versions_preserve_validation_precedence`.

The two V0 per-instance checks are on mutually exclusive execution paths: the offset callback returns an error immediately, so that request never reaches the trailing check. Their rejection branches are redundant for the shipped version tables, but the calls themselves are reachable. We can simplify surrounding control flow, but should retain the offset gate's position within classification and preserve the existing error precedence.

grovedb/src/query/validated.rs
QuantumExplorer: Agreed as a small cleanup. A crate-private `AggregateKind::proof_family_name()` could supply the existing names to generation and aggregate verification without expanding the public API. Narrowing the envelope helper as discussed above would remove one duplicate site as well.

The helper should preserve the current family names and messages verbatim; this PR is preserving the error contract.

grovedb/src/operations/proof/generate.rs
QuantumExplorer: Agreed. The only caller has already classified this as `CountOffsetPaginated`, so it can pass the borrowed `inner` directly to `prove_count_offset_layer_v1` and remove the redundant syntax validation and clone.

The target-tree eligibility checks and pagination execution should stay in the helper at their existing point in the walk. Removing the pure validation and clone should leave the tracked operation costs unchanged. This is directly within the shared-validation scope and worth addressing in this PR.

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude (original reviewer). Agreed on all points — a required field on the public `PathQueryShape` was the wrong form for the fix, and a `debug_assert!` alone leaves release builds unprotected. I'll implement the production guard: a hard internal error in `prove_subqueries_v1` when a layer's effective items carry an aggregate that `aggregate_at_depth` did not claim, restoring the old item-based fail-closed property without touching the public API.

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude. Agreed, and I verified the qualification: the consumer dispatchers (`run_path_query`, `verify_path_query`, `check_envelope`) all match shapes exhaustively, so a new variant already fails compilation there — the gap is exactly these policy tables. I'll make `check_proof_capabilities`, the `for_proof` V0 table, and `for_read`'s read-mode eligibility check exhaustive over `PathQueryShape`.

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude. Agreed — no reachable panic today; the fix is about making the invariant structural. I'll replace `check_envelope` with a `require_v1_envelope(&self, decoded: GroveDBProof) -> Result<GroveDBProofV1, Error>` scoped to the axis/sum-budget verification paths (their exact messages preserved), which removes both `unreachable!()` call sites and the dead selection/pagination/aggregate arms; the live gates in the dedicated verifiers stay untouched.

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude. Fair counterexample — it refutes the literal reading of my sketch, though not the intended one: the offset gate must stay inside classification exactly where it is (V0's offset-before-pagination-syntax precedence, pinned by `legacy_pagination_contracts_remain_distinct`, depends on it). What I'll consolidate is only the surrounding flow: keep the closure as-is, then `if prove_version == 0 { <message table> }`, `check_proof_capabilities`, and a single trailing `match prove_version { 0 => reject_per_instance_limits, 1 => {}, v => unknown }` — same observable precedence on every path, including capability-error-before-unknown-version for hypothetical v>1 tables, with the duplicated calls gone.

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude. Agreed — I'll add a crate-private `AggregateKind::proof_family_name() -> &'static str` with the names verbatim and use it from generation and the aggregate verification call sites; the envelope-helper cleanup from the thread above removes the third copy.

grovedb/src/operations/proof/generate.rs
QuantumExplorer: This is Claude. Agreed on all points — single caller confirmed, and the removal is pure (no tracked-cost change). I'll thread the shape's borrowed `inner` into `prove_count_offset_layer_v1` and delete the revalidate-and-clone, leaving the target-tree eligibility check and pagination walk where they are.

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude. Fixed in 3b8046d6: `prove_subqueries_v1` now hard-errors (`CorruptedCodeExecution`) if an aggregate item reaches a layer the depth check didn't claim — the item-based fail-closed property is back in release builds, with no public API change.

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude. Fixed in 3b8046d6: `for_read`, `for_proof`'s V0 table, and `check_proof_capabilities` now match exhaustively over `PathQueryShape` — a new shape variant is a compile error at each policy site.

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude. Fixed in 3b8046d6: `check_envelope` is now `require_v1_envelope(decoded) -> Result<GroveDBProofV1, Error>`; both `unreachable!()`s and the dead selection/pagination/aggregate arms are gone, and a wrong-shape call fails closed with an internal error.

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude. Fixed in 3b8046d6 along the agreed lines: the offset gate stays inside classification untouched; the trailing checks collapse into one `match prove_version` (V0 per-instance / V1 / unknown), removing the duplicated calls. Both pinned precedence tests and the 22 fixtures pass (3308 passed, 0 failed).

grovedb/src/query/validated.rs
QuantumExplorer: This is Claude. Fixed in 3b8046d6: crate-private `AggregateKind::proof_family_name()` now supplies the names verbatim to the `for_proof` V0 table and the three aggregate verify wrappers.

grovedb/src/operations/proof/generate.rs
QuantumExplorer: This is Claude. Fixed in 3b8046d6: the dispatch binds the shape's `inner` and passes it into `prove_count_offset_layer_v1`; the revalidate-and-clone is deleted. Cost fixtures unchanged.

PR #965, opened 2026-09-14
fix: answer a single-path axis read over an absent path with the empty result

QuantumExplorer:

A single-path axis read (`PathQuery::new_axis*`) whose path does not exist fails on every surface: `run_path_query` with a storage error (`InvalidParentLayerPath`), `prove_query` with an explicit refusal ("a single-path axis read must produce exactly one axis descent ... the path does not name an indexed tree carrying that axis"). Key selection (`test_proof_for_invalid_path`) and the branched axis shape already answer an absent path with an authenticated empty result, so this was the one provable shape that could not.

Surfaced in Dash Platform (dashpay/platform#4753): a ranked windowed top-K (`IN_TIME_RANGE newest` on a `range == step` grid) descends into the current window's bucket, which only exists once the first document lands in it, and until then every request came back as an internal error instead of an empty leaderboard. The Platform side worked around it by switching single-branch pinned reads to the branched envelope with one key; with this change that workaround goes away and the single-path shape answers correctly on its own, including for `OFFSET > 0` (which the branched envelope cannot attest).

## What was done?

Absence was always provable here. The generic V1 walk emits the layers that authenticate it: the parent of the missing segment proves the key absent through its single-key Merk proof, an empty ancestor is bound to `NULL_HASH` (`verify_empty_tree_binding`), every present non-empty tree on the path must carry its lower layer, and a present axis position must be an indexed tree with an axis layer (the #799 guards). So a verified walk with zero axis layers can only mean the path does not exist, or that a non-tree element blocks it (surfaced as a row).

- `verify_path_query`, `AxisRead` arm: zero axis layers now map to the traversal's empty answer through the new `absent_axis_path_into_verified`: no entries and `skipped: Some(0)` for a ranked page (the population is zero), no entries and `None` for a bounded walk, `0` for a value-range aggregate. A present non-tree row on the queried path is rejected. More than one layer stays an error.
- `prove_query` (`prove_query_non_serialized_v1`): with zero descents the prover settles the same question against the proof's own snapshot (`path_is_present`): an absent path yields the proof; a present path that does not lead to an indexed tree, or that runs through a non-tree element, is still refused.
- `run_path_query`, `AxisRead` arm: probes the path the way the branched arm already does and returns the same empty answer (`absent_axis_run`) for an absent path; a present non-tree element part-way down falls through to the primitive's kind error.
- `RankOfKey` has no empty answer and fails over an absent path on every surface with `PathKeyNotFound`, exactly as it does for a key absent from a present tree.
- `AxisKeys::empty_for_axis` added, mirroring `AxisEntries::empty_for_axis`.
- Docs (`unified-path-query.md`) and CHANGELOG updated.

No proof format or `GroveVersion` change: the envelope is unchanged, no prover at any version ever emitted a proof this verifier newly accepts (the prover refused), and the verify-side change is a strict relaxation that the walk's existing guards keep sound.

## How Has This Been Tested?

New tests in `axis_descent_proof_tests.rs`:
- `a_single_path_axis_read_over_an_absent_path_answers_empty_everywhere`: absent last segment, absent intermediate segment and an empty-tree ancestor round-trip empty through the reader (entries and keys-only), the prover and the verifier against the live root hash, for ranked (`skipped: Some(0)` with a requested offset of 1), bounded and both aggregate folds; rank-of-key is refused on both sides with `PathKeyNotFound`.
- `an_absent_path_proof_covers_only_its_own_gap`: the absence proof verifies another key inside its Merk gap as absent (which it truly is) and fails for a key past a hashed-out subtree and for the present tree next to it.
- `stripping_a_single_path_axis_layer_is_rejected_not_read_as_absence`: stripping the axis layer of a present indexed tree, populated or empty, from an honest single-path proof is still rejected.
- `a_single_path_axis_read_over_a_present_non_indexed_target_fails_generation` (replaces the test that pinned the old refusal for a missing path): a present plain tree and a path through an item are still refused by prover and reader.

Ran: `cargo test -p grovedb --lib` (3485 passed), `cargo clippy -p grovedb --all-targets -- -D warnings`, `cargo check -p grovedb --no-default-features --features verify`, `cargo fmt`.

## Breaking Changes

None. Reads and proofs over an absent path that previously errored now return the empty answer; everything that verified before still verifies to the same result.

## Checklist:

## Summary by CodeRabbit

* **Changed**
  * Single-path axis queries now return empty results when the requested path is missing or traverses an empty tree.
  * Ranked queries report no skipped entries, bounded traversals return no results, and value-range aggregates return zero.
  * `RankOfKey` continues to report a missing-path error.
  * Paths that exist but do not lead to an indexed tree continue to be rejected.
  * Proof generation and verification now consistently reflect these outcomes.

* **Documentation**
  * Updated query and proof documentation to describe missing-path behavior.


QuantumExplorer (commented):
Reviewed

PR #979, opened 2026-09-17
ci: re-pin PR Hygiene so it can verify access granted by the organisation

shumkov:
Re-pins the shared review workflow onto dashpay/stale_prs_are_bad#35.

Writes were enabled on `dashpay/platform` for the first time and six of twelve pull requests were immediately marked `configuration-error` — *"An assigned owner/reviewer lacks verified write access"*. The people named are administrators.

One listing of collaborators only names the people a token can see, and a repository-scoped Actions token does not enumerate those who reach a repository **through the organisation** rather than as collaborators of it. Four of the eight owners in platform's policy are in that position, including the one in `fallback.owners` — so essentially every pull request was affected, not six.

Anyone the listing omits is now asked about directly, once per run, and remembered. The listing still answers for everyone it names, so the request-volume saving it was introduced for is unaffected.

An answer that never arrived is no longer reported as *no* access — those are different things, and only one of them is about the person.

**This repository is affected too.** `lklimek` on dash-evo-tool and `QuantumExplorer` on rust-dashcore both reach their repository through the organisation, so the same thing would have happened on the day writes were turned on here. grovedb and tenderdash happen to name only direct collaborators, which is why tenderdash has been writing since 13 September without hitting it.

## Summary by CodeRabbit

- **Chores**
  - Updated the pull request review policy workflow to use a newer reusable workflow version.
  - Existing review event triggers and job conditions remain unchanged.


thepastaclaw:
🕓 Queued for automated review — 8th in line, estimated start in ~5 h (commit d80a8b11)
_Estimated review time once started: ~1.3 h (two-phase automated review; median of recent runs)._

- [ ] **Request priority review** — click to move this review to the front of the queue.

